Issue #211, opened 2019-09-20
enhance setup.py to facilitate installs for developers


GavinHuttley:
This will be a supplement to what's in the c3dev conda environment (which should rely on the following).
`setupy.py` can be modified to specify optional installs. For example, the following

```
$ pip install path/to/cogent3[dev] 
```
(may not have syntax correct) should install essential modules for developers. [See this project's setup.py](https://github.com/python-attrs/attrs/blob/master/setup.py) for a good example.

PR #218, opened 2019-09-20
ENH: facilitated installs for developers in setup.py, enhances #211

wjjmjh:
to execute the extras_require installs: 

`cd path/to/cogent3`

`pip install -e .[dev]`


wjjmjh:
removed redundancies in setup.py and did black on it.

Issue #221, opened 2019-09-23
app.align support building guide tree for protein


GavinHuttley:
If a user picks "protein" model for progressive_align, allow building the guide tree from paralinear

GavinHuttley:
insert `elif` clause at line 203 that catches case where `guide_tree is None` and `self._moltype.label` == "protein"`. In this case, use paralinear.

Issue #220, opened 2019-09-23
add cogent3 version field to json serialised data


GavinHuttley:
Prior to release, we need to enable backwards compatibility for  json data. I think the best way to do that is to version all json serialised objects. At least that way, the key identifier needed for such a capacity is in place.

The task:

- every `.to_rich_dict()` method should add a `"version": __version__` element (where `__version__` is the `cogent3` version string.
- every `cogent3.util.deserialise.deserialise_<name>` function should just pop those elements (`data.pop("version", None)`).

GavinHuttley:
Every `to_rich_dict()` method that uses `get_object_provenance` needs to be modified in the same logical group (i.e. when ` for_pickle is False`).

Do not add the version info to a structure that is for pickling.

PR #226, opened 2019-09-24
#220 #221

wjjmjh:
Thanks Gavin for the support! 

Please have a look and let me know if anything unsuitable.


src/cogent3/app/align.py
```diff
@@ -200,6 +200,8 @@ def __init__(
         if callable(guide_tree):
             self._make_tree = guide_tree
             guide_tree = None  # callback takes precedence
+        elif guide_tree is None and self._moltype.label == "protein":
+            self._distance = "paralinear"
```
GavinHuttley: I think this logic (and thus my original suggestion) is not right. The problem is this will have no effect on construction of `dist_calc` (currently line 205). So I now think the `elif` clause is incorrect

Instead, insert lines 203-204 before line 199 and change it to
```python
if guide_tree is None and self._moltype.label == "protein":
    distance = "paralinear"
```

change line 204 to `distance = "paralinear"`

src/cogent3/util/deserialise.py
```diff
@@ -276,7 +289,7 @@ def deserialise_object(data):
 
     if type(data) is str:
         data = json.loads(data)
-
+    data.pop("version", None)
```
GavinHuttley: delete this line as not needed. Should be addressed in all called functions.

PR #254, opened 2019-09-30
ENH: implemented to_moltype method in core.sequence.Sequence, fixes #247

### LINKED ISSUE #247: Sequence needs a `to_moltype()` method
`cogent3.core.sequence.Sequence` (line 644) needs a `to_moltype()` method. This is a pre-requisite for the #242.

This method needs to also copy annotations onto the new object. These objects are immutable. his means we need to create instances for the new object. This might be feasible as follows:

```python
new # is the new sequence object, different moltype
for ann in self.annotations:
    n = new.add_annotation(ann.__class__, ann.type, ann.name, ann.map)
```
One test is checking the new annotations parent is the new sequence. See test_features lines 28 and 37. The `get_slice()` method on the annotation should slice the correct part.

wjjmjh:
Thanks Gavin! 

I tried `new.add_annotation(ann.__class__, ann.type, ann.name, ann.map)` but it isn't working very well, so I implemented the pattern below and added corresponding tests. 

For some convenience, I also added some functions in cogent3/core/annotation._Annotatable class, please have a look and let me know if any unsuitable codes! 

Many thanks!


GavinHuttley:
There is a problem with this approach -- it adds methods onto `_Annotatable` that all sub-classes inherit. That's a problem because the method names are a bit confusing in those contexts. For instance, `Alignment.add_simple_variable()` does not have an obvious meaning.

I think I get what the issue was with my original approach (some of the extra arguments to the constructor). Here's an alternate approach.

## a new method `_Annotatable.copy_to_seq()`

Creates a new annotation with identical type, name, position etc.. on a new sequence.
`_Annotatable.copy_to_seq(new_seq)`. The method should check that the incoming sequence has exactly the same length as `self.parent`. To be safe, the method should be able to just use the contents of the `anno._serialisable` dict, (possibly need to do a `deepcopy`?), which contains all the construction arguments.

wjjmjh:
Many thanks Gavin! I will implement this new approach as soon as possible!

src/cogent3/core/annotation.py
```diff
@@ -172,11 +173,45 @@ def detach_annotations(self, annots):
...
+        seq : Sequence
+            represents the new sequence.
+
+        Returns
+        -------
+        a new annotation
+        """
+        assert len(seq) == len(self.parent)
+        serialisable = copy.deepcopy(self._serialisable)
+        if isinstance(self, AnnotatableFeature):
```
GavinHuttley: All these `ifinstance()` conditionals could possibly be replaced with the following:
```python
serialisable["parent"] = seq
new = self.__class__(**serialisable)
```

src/cogent3/core/sequence.py
```diff
@@ -713,20 +713,18 @@ def __init__(
...
     def to_moltype(self, moltype):
-        """returns copy of self with moltype seq"""
+        """returns copy of self with moltype seq
+
+        Parameters
+        ----------
+        moltype : str
+            molecular type
+        """
         make_seq = moltype.make_seq
```
GavinHuttley: insert at line 723
```python
moltype = get_moltype(moltype)
```

PR #273, opened 2019-10-11
Version exception for is_master_process(), fixes #272

### LINKED ISSUE #272: address difference in multiprocess between python 3.6 and 3.7
- skip test_is_master_process in 3.6
- raise RuntimeError exception in actual function if python version is 3.6, and a test that only runs if it’s 3.6!

u6052029:


tests/test_util/test_parallel.py
```diff
@@ -69,6 +71,14 @@ def test_is_master_process(self):
                 master_processes += 1
         self.assertEqual(master_processes, 0)
 
+    @skipIf(sys.version_info[1] >= 7, "exception test for Python 3.6 and below")
```
GavinHuttley: delete and below

PR #277, opened 2019-10-14
ENH: implemented to_moltype methods for both alignment and array_alignment, fixes #242

### LINKED ISSUE #242: SequenceCollection needs to `to_moltype()` method
take a MolType instance, or string. Checks current class and makes sure it returns an object of the same class.

wjjmjh:


tests/test_core/test_alignment.py
```diff
@@ -1185,6 +1186,32 @@ def test_to_moltype(self):
...
+        s1 = Sequence("TTTTTTAAAA", name="test_seq1")
+        s2 = Sequence("AAAATTTTTT", name="test_seq2")
+        s3 = Sequence("AATTTTTAAA", name="test_seq3")
+        s1.add_annotation(Feature, "exon", "fred", [(0, 6)])
+        s2.add_annotation(Feature, "exon", "fred", [(4, 10)])
+        s3.add_annotation(Feature, "exon", "fred", [(2, 7)])
+        data = {"seq1": s1, "seq2": s2, "seq3": s3}
+        seqs = self.Class(data=data)
+        rna = seqs.to_moltype(RNA)
+        if isinstance(self, AlignmentTests):
```
GavinHuttley: this check needs to be based on `self.Class` as discussed above.

tests/test_core/test_alignment.py
```diff
@@ -1185,6 +1186,32 @@ def test_to_moltype(self):
...
+            for seq in rna.seqs:
+                self.assertTrue(len(seq.data.annotations) > 0)
+                self.assertIsInstance(seq.data.annotations[0], _Annotatable)
+        elif isinstance(self, ArrayAlignmentTests):
+            for seq in rna.seqs:
+                self.assertTrue(len(seq.annotations) == 0)
+
```
GavinHuttley: you need to add an annotation onto the actual `Alignment` (not just its sequences). Here's an example on how to add a feature to an Alignment:
```python
from cogent3 import make_aligned_seqs
aln = make_aligned_seqs(dict(a="ACGT", b="A-GG"), array_align=False, moltype="dna")
f = aln.add_feature("demo", "one", [(0, 1), (2,4)])
aln.annotations
```

PR #307, opened 2019-10-25
ENH: replaced to_type usages with to_moltype in align app and sample …

wjjmjh:
…app, fixes 243

### DIFF
--- a/src/cogent3/app/align.py
+++ b/src/cogent3/app/align.py
@@ -129,13 +129,13 @@ class align_to_ref(ComposableSeq):
             # eliminating positions with deletions in the reference
             result = result.filtered(no_ref_gap)
             if aligned is None:
-                aligned = result.to_type(array_align=False)
+                aligned = result.to_moltype(self._moltype)
                 continue
 
-            aligned = aligned.add_from_ref_aln(result.to_type(array_align=False))
+            aligned = aligned.add_from_ref_aln(result.to_moltype(self._moltype))
 
         # default to ArrayAlign
-        new = aligned.to_type(array_align=True)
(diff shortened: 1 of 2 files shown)


src/cogent3/app/align.py
```diff
@@ -129,13 +129,13 @@ def align_to_named_seq(self, seqs):
             # eliminating positions with deletions in the reference
             result = result.filtered(no_ref_gap)
             if aligned is None:
-                aligned = result.to_type(array_align=False)
```
GavinHuttley: we don't need the conversion, just do the following
```python
aligned = result
continue
```

src/cogent3/app/align.py
```diff
@@ -129,13 +129,13 @@ def align_to_named_seq(self, seqs):
...
+                aligned = result.to_moltype(self._moltype)
                 continue
 
-            aligned = aligned.add_from_ref_aln(result.to_type(array_align=False))
```
GavinHuttley: as the `add_from_ref_aln` method already does the type conversion when required, change this to
```python
aligned = aligned.add_from_ref_aln(result)
```

src/cogent3/app/align.py
```diff
@@ -129,13 +129,13 @@ def align_to_named_seq(self, seqs):
...
+            aligned = aligned.add_from_ref_aln(result.to_moltype(self._moltype))
 
         # default to ArrayAlign
-        new = aligned.to_type(array_align=True)
```
GavinHuttley: all cases of `to_type` that do not include `moltype=` should be left as original.

PR #325, opened 2019-11-01
BLD: added cibuildwheel to pipeline

tla256:


azure-pipelines.yml
```diff
@@ -56,3 +58,38 @@ jobs:
...
         tox -e py36
         codecov
+
+- job: Linux_Build
+  pool: {vmImage: 'Ubuntu-16.04'}
+  steps: 
+    - task: UsePythonVersion@0
+    - bash: |
+        python -m pip install --upgrade pip
+        python -m pip install tox pytest click cython numpy pandas plotly scitrack tqdm tinydb cibuildwheel==0.11.1 pytest-azurepipelines
```
GavinHuttley: Can we change these massive long lines to `python -m pip install .[dev] cibuildwheel==0.11.1 pytest-azurepipelines`?

That should trigger install of the dev dependencies (and thus it's an implicit test we haven't forgotten something) while also adding the other just for CI dependencies.

azure-pipelines.yml
```diff
@@ -56,3 +58,38 @@ jobs:
...
+        cibuildwheel --output-dir wheelhouse .
+    - task: PublishBuildArtifacts@1
+      inputs: {pathtoPublish: 'wheelhouse'}
+- job: MacOS_Build
+  pool: {vmImage: 'macOS-10.13'}
+  steps: 
+    - task: UsePythonVersion@0
+    - bash: |
+        python -m pip install --upgrade pip
+        python -m pip install tox pytest click cython numpy pandas plotly scitrack tqdm tinydb cibuildwheel==0.11.1
```
GavinHuttley: as above

azure-pipelines.yml
```diff
@@ -56,3 +58,38 @@ jobs:
...
+- job: Windows_Build
+  pool: {vmImage: 'vs2017-win2016'}
+  steps:
+    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x86}}
+    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.6', architecture: x64}}
+    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x86}}
+    - {task: UsePythonVersion@0, inputs: {versionSpec: '3.7', architecture: x64}}
+    - bash: |
+        python -m pip install --upgrade pip
+        python -m pip install tox pytest click cython numpy pandas plotly scitrack tqdm tinydb cibuildwheel==0.11.1
```
GavinHuttley: as above

azure-pipelines.yml
tla256: I see that `pytest-azurepipelines` is already in `[dev]`, so I will omit that, but the change has been made, will push shortly.

Issue #196, opened 2019-09-08
io writers should check destination more thoroughly before deleting


GavinHuttley:
**[Original report](https://bitbucket.org/pycogent3/cogent3-dev/issue/196) by GavinH (Bitbucket: [557058:e40c23e1-e273-4527-a2f8-5de5876e870d](https://bitbucket.org/557058:e40c23e1-e273-4527-a2f8-5de5876e870d), ).**

----------------------------------------

If the destination not a tinydb, the writer should check if it contains files with a different suffix to it's own format. If so, it should ignore any if_exists value and require user to directly delete.

GavinHuttley:
fixed in 28e17b5a196bfbd30fddee61c1226081d51b69c3

PR #330, opened 2019-11-04
#196, io writers should check destination more thoroughly before deleting 

wjjmjh:
revised `_has_other_suffixes` function to adapt to the design pattern.


src/cogent3/app/data_store.py
```diff
@@ -570,15 +570,15 @@ def _has_other_suffixes(self, path, suffix):
...
+                return True
+        return False
 
     def _source_create_delete(self, if_exists, create):
         exists = os.path.exists(self.source)
         if exists and if_exists == RAISE:
             raise RuntimeError(f"'{self.source}' exists")
         elif exists and if_exists == OVERWRITE:
-            if self._has_other_suffixes(self.source, self.suffix):
+            if not self._has_other_suffixes(self.source, self.suffix):
```
GavinHuttley: Oops, I think I should have asked for you to make this call the same as the one you do below, i.e. raise `RuntimeError` with the same message as below. So that means this does become `if self._has_other...: raise...`

tests/test_app/test_data_store.py
```diff
@@ -281,6 +279,30 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
     ReadClass = ReadOnlyDirectoryDataStore
     WriteClass = WritableDirectoryDataStore
 
+    def test_write_class_source_create_delete(self):
+        # tests the case when the directory has other different suffixes to self.suffix
+        os.mkdir("delme_dir")
```
GavinHuttley: This approach to creating a working directory is fragile, for instance if it was not cleaned up by a previous test. Use the
`from tempfile import TemporaryDirectory` approach (e.g. see `test_app.test_io.test_findall_zip`). You can then create the `delme_dir` inside the temporary directory and the `tempfile` module will deal with cleanup.

tests/test_app/test_data_store.py
```diff
@@ -281,6 +279,30 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+        dstore = self.WriteClass(
+            "delme_dir", suffix=".json", if_exists=OVERWRITE, create=True
+        )
+        self.assertEqual(len(glob.glob(f"delme_dir/*.json")), 0)
+        with open(f"delme_dir/test_write_class_source_create_delete.dummySuffix", "w"):
+            pass
+        dstore = self.WriteClass(
+            "delme_dir", suffix=".json", if_exists=OVERWRITE, create=True
+        )
+        self.assertEqual(len(glob.glob(f"delme_dir/*.dummySuffix")), 1)
```
GavinHuttley: when there's another suffix, a `RuntimeError` should be raised (like line 346)

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
-            "delme_dir", suffix=".dummySuffix", if_exists=OVERWRITE, create=True
-        )
-        self.assertEqual(len(glob.glob(f"delme_dir/*.dummySuffix")), 0)
-        try:
-            os.removedirs("delme_dir")
-        except OSError:
-            pass
+        with TemporaryDirectory(dir=".") as dirname:
+            # tests the case when the directory has the file with the same suffix to self.suffix
+            os.mkdir(f"{dirname}{os.sep}delme_dir")
```
GavinHuttley: replace this with
```python
path = os.path.join(dirname, "delme_dir")
outdir = os.mkdir(path)
```

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+            with open(
+                f"{dirname}{os.sep}delme_dir/test_write_class_source_create_delete.json",
```
GavinHuttley: replace this line with
```python
os.path.join(path, "test_write_class_source_create_delete.json",
```

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+                "w",
+            ):
+                pass
+            dstore = self.WriteClass(
+                f"{dirname}{os.sep}delme_dir",
```
GavinHuttley: replace this `f"{dirname}{os.sep}delme_dir"` with `path`

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+                suffix=".json",
+                if_exists=OVERWRITE,
+                create=True,
+            )
+            self.assertEqual(len(glob.glob(f"{dirname}{os.sep}delme_dir/*.json")), 0)
+            # tests the case when the directory has the file with the different suffix to self.suffix
+            with open(
+                f"{dirname}{os.sep}delme_dir/test_write_class_source_create_delete.dummySuffix",
```
GavinHuttley: Following above, use `os.path.join(path, <flename>)` here

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+                "w",
+            ):
+                pass
+            with self.assertRaises(RuntimeError):
+                dstore = self.WriteClass(
+                    f"{dirname}{os.sep}delme_dir",
```
GavinHuttley: and again

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+                    suffix=".json",
+                    if_exists=OVERWRITE,
+                    create=True,
+                )
+            self.assertEqual(
+                len(glob.glob(f"{dirname}{os.sep}delme_dir/*.dummySuffix")), 1
```
GavinHuttley: and again

tests/test_app/test_data_store.py
```diff
@@ -280,28 +279,47 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+            )
+            # tests the case when the directory has the file with the same suffix to self.suffix
+            dstore = self.WriteClass(
+                f"{dirname}{os.sep}delme_dir",
+                suffix=".dummySuffix",
+                if_exists=OVERWRITE,
+                create=True,
+            )
+            self.assertEqual(
+                len(glob.glob(f"{dirname}{os.sep}delme_dir/*.dummySuffix")), 0
```
GavinHuttley: and again

tests/test_app/test_data_store.py
```diff
@@ -290,6 +322,14 @@ def setUp(self):
 
     def tearDown(self):
```
GavinHuttley: don't use this, use the temporary directory and create zip inside it.

tests/test_app/test_data_store.py
```diff
@@ -276,6 +278,36 @@ class DirectoryDataStoreTests(TestCase, DataStoreBaseTests):
...
+            path = os.path.join(dirname, "delme_dir")
+            os.mkdir(path)
+            with open(
+                os.path.join(path, "test_write_class_source_create_delete.json"), "w"
+            ):
+                pass
+            dstore = self.WriteClass(
+                path, suffix=".json", if_exists=OVERWRITE, create=True
+            )
+            self.assertEqual(len(glob.glob(os.path.join(path, "*.json"))), 0)
```
GavinHuttley: pretty sure this can be replaced with `self.assertEqual(len(dstore), 0)`, so you can remove the glob import and calls.

Issue #342, opened 2019-11-05
refactor of `BlastResult`


GavinHuttley:
Not used in `parse.blast`, but inherited from in `parse.blast_xml`.

- Migrate attributes, methods necessary for `BlastXMLResult` to work onto that class
- remove all methods with `NotImplementedError`, or no actual code (!)
- remove `parse.blast.BlastResult`
- make methods on `BlastXMLResult` pep8 compliant (`name_underscore`)
- add test method that exercises `best_hits_by_query`

[blast.txt](https://github.com/cogent3/cogent3/files/3807298/blast.txt)

PR #343, opened 2019-11-06
Issue #342

ChristopherBradley:


src/cogent3/parse/blast_xml.py
```diff
@@ -181,29 +177,68 @@ class BlastXMLResult(BlastResult):
...
+            SUBJECT_START,
+            SUBJECT_END,
+            E_VALUE,
+            BIT_SCORE,
+        ]
     )
 
+    # standard comparison for each field, e.g.
+    # want long matches, small e-values
+    _lt = lambda x, y: x < y
```
GavinHuttley: delete all these and do a top level import as
```python
from operator import lt as _lt, ...
```
[see standard library docs](https://docs.python.org/3/library/operator.html).

src/cogent3/parse/blast_xml.py
```diff
@@ -181,29 +177,68 @@ class BlastXMLResult(BlastResult):
...
+    _le = lambda x, y: x <= y
+    _gt = lambda x, y: x > y
+    _ge = lambda x, y: x >= y
+    _eq = lambda x, y: x == y
+
+    FieldComparisonOperators = {
```
GavinHuttley: pep8, but make semi-private
```python
_field_comp...
```

tests/test_parse/test_blast_xml.py
```diff
@@ -209,6 +209,25 @@ def test_parse_hsp_details(self):
             gap_hsp = self.result[query][0][1]
             self.assertEqual(gap_hsp["GAP_OPENINGS"], "33")
 
+    def test_best_hits_by_query(self):
+        q, best_hits = next(self.result.best_hits_by_query(n=1))
```
GavinHuttley: exercising best hits

tests/test_parse/test_blast_xml.py
```diff
@@ -209,6 +209,25 @@ def test_parse_hsp_details(self):
...
+        self.assertEqual(best_hit["QUERY_ALIGN"], "ELEPHANTTHISISAHITTIGER")
+        self.assertEqual(best_hit["MIDLINE_ALIGN"], "ORCA-WHALE")
+        self.assertEqual(best_hit["SUBJECT_ALIGN"], "SEALSTHIS---HIT--GER")
+        self.assertEqual(best_hit["SUBJECT_START"], "5")
+        self.assertEqual(best_hit["SUBJECT_END"], "19")
+        self.assertEqual(best_hit["PERCENT_IDENTITY"], "55")
+        self.assertEqual(best_hit["POSITIVE"], "555")
+        self.assertEqual(best_hit["GAP_OPENINGS"], 0)
+        self.assertEqual(best_hit["ALIGNMENT_LENGTH"], "14")
+
```
GavinHuttley: add another test method (`test_best_hits_limited`) that uses a different value of `n`, should never return more than is possible to return.

tests/test_parse/test_blast_xml.py
```diff
@@ -228,6 +229,16 @@ def test_best_hits_by_query(self):
...
 
+    def test_best_hits_limited(self):
+        """Exercising best hits with different values of n.
+        result should never contain identical hits
+        """
+        for q, best_hits in self.result.best_hits_by_query(n=5):
+            assert len(best_hits) == 3
+            assert best_hits[0].values() != best_hits[1].values()
+            assert best_hits[0].values() != best_hits[2].values()
+            assert best_hits[2].values() != best_hits[2].values()
```
GavinHuttley: Need to do this test differently.
```python
records = [h for _, h in self.result.best_hits_by_query(n=5)]
self.assertEqual(len(records), 3)
values = {tuple(h.values()) for h in records}
self.assertEqual(len(records), 3)
```

PR #375, opened 2019-11-12
deserialise_annotations handles nested annotations

ChristopherBradley:


tests/test_core/test_features.py
```diff
@@ -636,6 +636,14 @@ def test_nested_to_rich_dict(self):
             self.nested_feature.to_rich_dict(),
         )
 
+    def test_nested_deserialise_annotation(self):
+        """nested annotations can be deserialised"""
+        got = self.s.to_json()
+        new = deserialise_object(got)
+        new_exon1 = new.annotations[0]
+        new_nested_feature = new_exon1.annotations[0]
+        self.assertEqual(new_nested_feature.to_json(), self.nested_feature.to_json())
```
GavinHuttley: suggest using the object `to_rich_dict()` methods as that should be robust to changes in dict traversal order that might (I have no evidence this will be the case, but just in case...)

Issue #304, opened 2019-10-24
repr() of app.dist.fast_slow_dist not showing full setting


GavinHuttley:
just displays `fast_slow_dist(type='distance')`. It should show the `fast_calc` and `slow_calc` settings...

PR #380, opened 2019-11-12
#304 repr() of app.dist.fast_slow_dist showing full setting

wjjmjh:


src/cogent3/app/dist.py
```diff
@@ -39,6 +39,9 @@ def __init__(self, distance=None, moltype=None, fast_calc=None, slow_calc=None):
         self._moltype = moltype if moltype is None else get_moltype(moltype)
```
GavinHuttley: insert `self._formatted_params()` at line 39

src/cogent3/app/dist.py
```diff
@@ -39,6 +39,9 @@ def __init__(self, distance=None, moltype=None, fast_calc=None, slow_calc=None):
...
         self._sm = None
 
+        self._formatted.append("fast_cal='%s'" % str(fast_calc))
```
GavinHuttley: delete these two lines

PR #383, opened 2019-11-13
SequenceCollection.annotate_from_gff works in Alignment subclass, but not in the generic SequenceCollection class

ChristopherBradley:
Changing the .named_seqs attribute to .get_seq(name) had no effect on the gaps, as .get_seq directly calls .named_seqs. 

However the current implementation does remove gaps if using the Alignment class, but not the SequenceCollection class. I have made a test to demonstrate this.

### DIFF
--- a/tests/test_core/test_alignment.py
+++ b/tests/test_core/test_alignment.py
@@ -878,22 +878,26 @@ class SequenceCollectionBaseTests(object):
             ["seq5", "prog2", "snp", "2", "3", "1.0", "+", "1", '"yyy"'],
         ]
         gff = list(map("\t".join, gff))
-        aln.annotate_from_gff(gff)
-        aln_seq_1 = aln.named_seqs["seq1"]
-        if not hasattr(aln_seq_1, "annotations"):
-            aln_seq_1 = aln_seq_1.data
-        aln_seq_2 = aln.named_seqs["seq2"]
-        if not hasattr(aln_seq_2, "annotations"):
-            aln_seq_2 = aln_seq_2.data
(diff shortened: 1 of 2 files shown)


tests/test_core/test_alignment.py
```diff
@@ -871,12 +871,18 @@ def test_copy_annotations(self):
...
         """SequenceCollection.annotate_from_gff should read gff features"""
-        aln = self.Class({"seq1": "ACGU", "seq2": "CGUA", "seq3": "CCGU"})
+        aln = self.Class({"seq1": "ACGU", "seq2": "CGUA", "seq3": "C-GU"})
         gff = [
             ["seq1", "prog1", "snp", "1", "2", "1.0", "+", "1", '"abc"'],
+            ["seq3", "prog2", "del", "1", "3", "1.0", "+", "1", '"xyz"'],
             ["seq5", "prog2", "snp", "2", "3", "1.0", "+", "1", '"yyy"'],
         ]
         gff = list(map("\t".join, gff))
+        if isinstance(aln, ArrayAlignment):
```
GavinHuttley: change this clause to `if self.Class == ArrayAlignment:`

tests/test_core/test_alignment.py
```diff
@@ -888,6 +894,11 @@ def test_annotate_from_gff(self):
         self.assertEqual(aln_seq_1.annotations[0].name, "abc")
         self.assertEqual(len(aln_seq_2.annotations), 0)
 
+        if isinstance(aln, Alignment):
```
GavinHuttley: use `self.Class == Alignment` instead of `isinstance()`

PR #384, opened 2019-11-13
Refactored parse.gff

ChristopherBradley:
Could not get cogent3.util.misc.open_(f) to work with pathlib.Path.
Allowed input as StringIO to keep old tests, and I guess backwards compatibility.
Converted the example gff2 data `data_lines` to a file to be read in so it could be checked with existing results.

### DIFF
--- a/src/cogent3/parse/gff.py
+++ b/src/cogent3/parse/gff.py
@@ -2,15 +2,27 @@
 
 __author__ = "Peter Maxwell"
 __copyright__ = "Copyright 2007-2019, The Cogent Project"
-__credits__ = ["Peter Maxwell", "Matthew Wakefield", "Gavin Huttley"]
+__credits__ = ["Peter Maxwell", "Matthew Wakefield", "Gavin Huttley", "Christopher Bradley"]
 __license__ = "BSD-3"
 __version__ = "2019.10.24a"
 __maintainer__ = "Peter Maxwell"
 __email__ = "pm67nz@gmail.com"
 __status__ = "Production"
 
+from pathlib import Path
+from cogent3.util.misc import open_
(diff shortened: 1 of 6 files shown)


src/cogent3/parse/gff.py
```diff
@@ -2,15 +2,38 @@
...
 
-def GffParser(f):
+from cogent3.util.misc import open_
+
+
+def gff_parser(f):
+    if isinstance(f, str):
+        with open_(f) as infile:
+            yield from gff2_parser(infile)
+    elif isinstance(f, Path):
```
GavinHuttley: Remove the `elif isinstance(f, Path)` clause

src/cogent3/parse/gff.py
```diff
@@ -2,15 +2,38 @@
...
 
+from io import StringIO
+from pathlib import Path
 
-def GffParser(f):
+from cogent3.util.misc import open_
+
+
+def gff_parser(f):
+    if isinstance(f, str):
```
GavinHuttley: convert `pathlib.Path` instances to str here
```python
f = f if not isinstance(f, Path) else str(f)
```

src/cogent3/parse/gff.py
```diff
@@ -21,12 +21,11 @@
 
 
 def gff_parser(f):
-    if isinstance(f, str):
+    """delegates to the correct gff_parser based on the version"""
+    if isinstance(f, str) or isinstance(f, Path):
+        f = f if not isinstance(f, Path) else str(f)
```
GavinHuttley: if you put the `f = f if not...` line before the `if isinstance` clause, you can remove the check for Path instance clause

Issue #382, opened 2019-11-12
add log suffix to allowed suffixes in data store


GavinHuttley:
modify usage of `_has_other_suffixes` so that it allows either the file suffix or "log".
Do this after #196 is closed.

GavinHuttley:
fixed in 05b0484b90c9634f4ba8224c35092758dc76e72a

PR #395, opened 2019-11-15
#382 added log suffix to allowed suffixes in data store

wjjmjh:
also implemented corresponding tests


src/cogent3/app/data_store.py
```diff
@@ -569,7 +569,10 @@ def __init__(
     def _has_other_suffixes(self, path, suffix):
         p = Path(path)
```
GavinHuttley: add line
```python
allowed = {suffix, "log"}
```

src/cogent3/app/data_store.py
```diff
@@ -569,7 +569,10 @@ def __init__(
...
         for f in p.iterdir():
-            if get_format_suffixes(str(f))[0] != suffix:
+            if (
```
GavinHuttley: change this conditional to
```python
if get_format_suffixes(str(f))[0] not in allowed:
...
```

src/cogent3/app/data_store.py
```diff
@@ -581,7 +584,7 @@ def _source_create_delete(self, if_exists, create):
             if self._has_other_suffixes(self.source, self.suffix):
                 raise RuntimeError(
                     f"Unsafe to delete {self.source} as it contains ",
-                    f"files other than {self.suffix}."
+                    f"files other than {self.suffix} or log files."
```
GavinHuttley: change to
```python
f"files other than .{self.suffix} or .log files."
```

src/cogent3/app/data_store.py
```diff
@@ -650,7 +653,10 @@ def __init__(
 
     def _has_other_suffixes(self, path, suffix):
         for f in zipfile.ZipFile(path).namelist():
```
GavinHuttley: refactor this function to be the same as above (define `allowed` as a set, etc..)

src/cogent3/app/data_store.py
```diff
@@ -663,7 +669,7 @@ def _source_create_delete(self, if_exists, create):
             if self._has_other_suffixes(self.source, self.suffix):
                 raise RuntimeError(
                     f"Unsafe to delete {self.source} as it contains ",
-                    f"files other than {self.suffix}."
+                    f"files other than {self.suffix} or log files."
```
GavinHuttley: change error message to be same as above

tests/test_app/test_data_store.py
```diff
@@ -305,6 +305,19 @@ def test_write_class_source_create_delete(self):
...
+                os.path.join(path, "test_write_class_source_create_delete.log"), "w"
+            ):
+                pass
+            with self.assertRaises(RuntimeError):
+                dstore = self.WriteClass(
+                    path, suffix=".json", if_exists=OVERWRITE, create=True
+                )
+            with self.assertRaises(RuntimeError):
+                dstore = self.WriteClass(
+                    path, suffix=".log", if_exists=OVERWRITE, create=True
```
GavinHuttley: don't define the suffix as `.log`. The pattern is the data store can contain both the allowed suffix and members with name ending in `.log`, never only `.log` files.

Issue #199, opened 2019-09-10
available_apps no longer displaying input/output/data types


GavinHuttley:
**[Original report](https://bitbucket.org/pycogent3/cogent3-dev/issue/199) by GavinH (Bitbucket: [557058:e40c23e1-e273-4527-a2f8-5de5876e870d](https://bitbucket.org/557058:e40c23e1-e273-4527-a2f8-5de5876e870d), ).**

----------------------------------------

The shift to making the `_input_types`, _output_types`, `_data_types` instance attributes makes it more difficult to discern what these values are.

Those arguments are always passed to the parent class constructor, but it's not clear how we can get at them without having to instantiate every one. And if the latter approach is chosen, we'd need to define a separate function for instantiating them all (e.g. `test_init._get_all_composables`) since they often have positional arguments.

PR #405, opened 2019-11-18
available_apps are now displaying input/output/data types, #199

wjjmjh:
Hi Gavin! 

I found this ticket was very related to the tasks I did before, so I implemented some codes to fix it, hope you don't mind, feel free to leave comments there, I will fix it as soon as possible! Many thanks for your teaching! 

Basically, I implemented a very simple pattern to composable apps to make available_apps be able to display input/output/data types. 

By the way, I was also thinking to put this block of codes
`
super(compassable_app_here, self).__init__(
            input_types=self._input_types,
            output_types=self._output_types,
            data_types=self._data_types,
        )
`
into its parent class, but this approach worked not very well, to avoid breaking current structure. I finally adopted the simplest modification. Hope you can have a look! 

Thanks Gavin!

### DIFF
--- a/src/cogent3/app/__init__.py
+++ b/src/cogent3/app/__init__.py
@@ -22,24 +22,23 @@ def _get_app_attr(name, obj, mod, is_composable):
     _output_types = getattr(obj, "_output_types", [])
     _data_types = getattr(obj, "_data_types", [])
 
-    if type(_input_types) == str:
-        _input_types = [_input_types]
-    if type(_output_types) == str:
-        _output_types = [_output_types]
-    if type(_data_types) == str:
-        _data_types = [_data_types]
-
     if _input_types:
+        if type(_input_types) == str:
+            _input_types = [_input_types]
         in_type = [{None: ""}.get(e, e) for e in _input_types]
     else:
         in_type = []
 
     if _output_types:
+        if type(_output_types) == str:
+            _output_types = [_output_types]
         out_type = [{None: ""}.get(e, e) for e in _output_types]
     else:
         out_type = []
 
     if _data_types:
+        if type(_data_types) == str:
+            _data_types = [_data_types]
         data_type = [{None: ""}.get(e, e) for e in _data_types]
     else:
         data_type = []
--- a/src/cogent3/app/__init__.py
+++ b/src/cogent3/app/__init__.py
@@ -21,9 +21,8 @@ def _get_app_attr(name, obj, mod, is_composable):
 
     _types = {"_input_types": [], "_output_types": [], "_data_types": []}
 
-    for types in _types.keys():
-        tys = types
-        types = getattr(obj, types, None) or []
+    for tys in _types.keys():
+        types = getattr(obj, tys, None) or []
         types = [types] if type(types) == str else types
         _types[tys] = [{None: ""}.get(e, e) for e in types]
 
--- a/src/cogent3/app/__init__.py
+++ b/src/cogent3/app/__init__.py
@@ -25,20 +25,16 @@ def _get_app_attr(name, obj, mod, is_composable):
         tys = types
         types = getattr(obj, types, None) or []
         types = [types] if type(types) == str else types
-        _types[tys] = types
-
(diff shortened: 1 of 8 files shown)


src/cogent3/app/__init__.py
```diff
@@ -18,9 +18,31 @@
 
 def _get_app_attr(name, obj, mod, is_composable):
```
GavinHuttley: I don't think you should need to make changes to this function. See other comments.

src/cogent3/app/io.py
```diff
@@ -150,6 +150,10 @@ class load_aligned(_seq_loader, ComposableAligned):
 
     klass = ArrayAlignment
 
+    _input_types = None
```
GavinHuttley: these attributes are meant to be `frozenset` instances, not raw strings, so that we can apply set operations in evaluating app compatability. That's why the function in `__init__` is wrong

src/cogent3/app/io.py
```diff
@@ -362,6 +378,10 @@ def write(self, data, identifier=None):
 class write_seqs(_checkpointable):
     """Writes sequences to text files in standard format."""
 
+    _input_types = ("sequences", "aligned")
```
GavinHuttley: You must use the TYPES defined in composable! Quite clearly, the original code was also wrong, i.e. `input_types=SEQUENCE_TYPE`.

src/cogent3/app/io.py
```diff
@@ -422,8 +442,13 @@ class load_json(Composable):
 
     _type = "output"
 
+    _input_types = None
+    _output_types = "serialisable"
```
GavinHuttley: use the appropriate defined TYPE

src/cogent3/app/io.py
```diff
@@ -451,12 +476,15 @@ class write_json(_checkpointable):
 
     _type = "output"
 
+    _input_types = "serialisable"
```
GavinHuttley: and again, I don't know how so many of these were not changed to using the TYPEs

src/cogent3/app/__init__.py
wjjmjh: Thanks so much for your comments Gavin! 

Regarding FrozenSet issue, I noticed all the raw Strings would be converted into frozen sets in the inherited class `ComposableType`, do we still need make the attributes to be frozen sets in composable apps?

GavinHuttley: You're right, it is not necessary.

src/cogent3/app/__init__.py
```diff
@@ -61,5 +83,5 @@ def available_apps():
     table = Table(header, rows)
     # todo the inputs/outputs/data type columns no longer being populated as
```
GavinHuttley: as you're fixing this issue you should delete this comment line

src/cogent3/app/__init__.py
```diff
@@ -18,39 +18,23 @@
...
-    else:
-        out_type = []
-
-    if _data_types:
-        if type(_data_types) == str:
-            _data_types = [_data_types]
-        data_type = [{None: ""}.get(e, e) for e in _data_types]
-    else:
-        data_type = []
+    for types in _types.keys():
```
GavinHuttley: tidy the loop
```python
for tys in _types.keys():
        types = getattr(obj, tys, None) or []
```

src/cogent3/app/__init__.py
```diff
@@ -18,39 +18,23 @@
...
-    _output_types = getattr(obj, "_output_types", [])
-    _data_types = getattr(obj, "_data_types", [])
 
-    if _input_types:
-        if type(_input_types) == str:
-            _input_types = [_input_types]
-        in_type = [{None: ""}.get(e, e) for e in _input_types]
-    else:
-        in_type = []
+    _types = {"_input_types": [], "_output_types": [], "_data_types": []}
```
GavinHuttley: given these are now dictionary keys, no need for the leading underscore, i.e. `_types = {"input_types": [], "output_types": [], "data_types": []}

PR #409, opened 2019-11-25
ENH: enabled `get_annotations_from_any_seq` in class `Alignment` to handle multiple arguments, fixed #225

### LINKED ISSUE #225: Alignment. get_annotations_from_any_seq incorrect handling of multiple arguments
This method calls methods on sequences. The following fails:

```python
aln.get_annotations_from_any_seq(annotation_type="*", name="chr1_80692396_T>A")
```
unless the second argument is removed. The traceback is
```python
get_annotations_from_any_seq() got an unexpected keyword argument 'name'
```

wjjmjh:


src/cogent3/core/alignment.py
```diff
@@ -2073,8 +2073,12 @@ def remapped_to(self, map):
         result = Aligned(map[self.map.inverse()].inverse(), self.data)
         return result
 
-    def get_annotations_matching(self, alignment, annotation_type="*", *args):
-        for annot in self.data.get_annotations_matching(annotation_type, *args):
+    def get_annotations_matching(
```
GavinHuttley: We need to support the extended querying (e.g. `extend_query=True`) so add `**kwargs` to this method and the called method.

src/cogent3/core/alignment.py
```diff
@@ -4158,17 +4162,17 @@ def get_projected_annotations(self, seq_name, *args):
...
-    def get_annotations_from_seq(self, seq_name, annotation_type="*", *args):
+    def get_annotations_from_seq(self, seq_name, annotation_type="*", name=None, *args):
         aligned = self.named_seqs[seq_name]
         return aligned.get_annotations_matching(
-            self, annotation_type=annotation_type, *args
+            self, annotation_type=annotation_type, name=name, *args
         )
 
-    def get_annotations_from_any_seq(self, annotation_type="*", *args):
+    def get_annotations_from_any_seq(self, annotation_type="*", name=None, *args):
```
GavinHuttley: We need to support the extended querying (e.g. `extend_query=True`) so add `**kwargs` to this method and the called method.

PR #420, opened 2019-12-02
changed dotplot to display the alignment even if there are no gaps

ChristopherBradley:


src/cogent3/draw/dotplot.py
```diff
@@ -214,7 +210,7 @@ def __init__(
 
         self.seq1 = seq1
         self.seq2 = seq2
-        self._aligned_coords = get_align_coords(map1, map2, aligned=is_aligned)
```
GavinHuttley: I'm not sure if this logic is right. There should not be an alignment path if the sequences came from a `SequenceCollection`, for example. Try out the doc/visualisation dotplot notebooks to see they're not broken.

PR #430, opened 2019-12-04
Added Alignment repr policy

u6052029:


src/cogent3/core/alignment.py
```diff
@@ -1841,6 +1843,19 @@ def apply_pssm(
 
         return array(result)
 
+    def set_repr_policy(self, num_seqs=None, num_pos=None):
+        """specify policy for repr(self)
+
+        Parameters
+        ----------
+
+        - num_seqs: number of sequences to included in represented display
```
GavinHuttley: this docstring style is wrong. See `AlignmentI.counts_per_seq` for an example.

src/cogent3/core/alignment.py
```diff
@@ -1841,6 +1843,19 @@ def apply_pssm(
...
+        - num_pos: length of sequences to included in represented display
+        """
+        if not any([num_seqs, num_pos]):
+            return
+        self._repr_policy = dict(num_seqs=num_seqs, num_pos=num_pos)
```
GavinHuttley: assert both values are int's

PR #445, opened 2019-12-08
support custom variables in user_function

jamesmartini:


tests/test_app/test_composable.py
```diff
@@ -300,6 +300,17 @@ def foo(self, val, *args, **kwargs):
     def bar(self, val, *args, **kwargs):
         return val.distance_matrix(calc="hamming", show_progress=False)
 
+    def _demo(self,ctx, aln):
```
GavinHuttley: change `aln` to `expect`

tests/test_app/test_composable.py
```diff
@@ -300,6 +300,17 @@ def foo(self, val, *args, **kwargs):
...
+        assert ctx.frame_start
```
GavinHuttley: assert that `ctx.frame_start == expect`

tests/test_app/test_composable.py
```diff
@@ -300,6 +300,17 @@ def foo(self, val, *args, **kwargs):
...
+        return aln
+
+    def test_user_function_custom_variables(self):
+        demo = user_function(
+            self._demo, ("aligned", "serialisable"), ("aligned", "serialisable")
+        )
+        foo = demo(1)
+        foo.frame_start = 2
+
```
GavinHuttley: need to test by calling `foo()`. Suggest the following:
```python
frame_start = 2
foo.frame_start = frame_start
foo(frame_start)
```

PR #446, opened 2019-12-08
Develop to_regex method and associated test

jamesmartini:


src/cogent3/core/moltype.py
```diff
@@ -742,6 +742,17 @@ def to_json(self):
         data = self.to_rich_dict(for_pickle=False)
         return json.dumps(data)
 
+    def to_regex(self, seq):
```
GavinHuttley: here's a more efficient implementation
```python
degen_indices = self.get_degenerate_positions(sequence=seq, include_gap=False)
seq = list(seq)  # seq can now be modified
for index in degen_indices:
    expanded = self.ambiguities[seq[index]]
    seq[index] = f"[{''.join(expanded)}]"
return "".join(seq)
    
```

src/cogent3/core/moltype.py
```diff
@@ -742,6 +742,17 @@ def to_json(self):
...
+        """returns regular expression for input sequence"""
```
GavinHuttley: change docstring to "returns a regex pattern with ambiguities expanded to a character set"

Issue #408, opened 2019-11-25
possible performance improvement for likelihood calculation


GavinHuttley:
We currently use Cython for a number of key steps in the likelihood calculation. Given recent work showing `numba.jit` was able to get better performance than a hand-tuned Cython module, I'm wondering whether we can improve likelihood calculation also.

See line 24 of `likelihood_tree.py`, which imports the compiled cython variant (`_likelihood_tree.pyx`). We can easily turn off the cython (pyrex) external module by raising an import error. 

Can we decorate methods of `_PyLikelihoodTreeEdge` using `numba.jit`?

Use the following code to benchmark performance, using numbers of sequences equal to:
- 3 
- 35

```python
from cogent3 import load_aligned_seqs, load_tree
from cogent3.app.evo import model

aln = load_aligned_seqs("data/brca1.fasta", moltype="dna")
aln = aln.take_seqs(aln.names[:35])
print(len(aln.names))
tree = load_tree("data/murphy.tree")
tree = tree.get_sub_tree(aln.names)

codon = model("GNC", tree=tree, 
              show_progress=True,
              opt_args=dict(max_evaluations=1000, limit_action="ignore"))
result = codon(aln)
```

PR #462, opened 2019-12-20
#408 performance improvement migration

wjjmjh:
implemented Numba solutions for likelihood tree and migrated it into Cogent3..


azure-pipelines.yml
```diff
@@ -1,5 +1,5 @@
 trigger:
-  - master
+  - feature
```
GavinHuttley: does this mean the tests only get triggered by commits to the feature branch? Is it possible to completely omit the `trigger` section?

src/cogent3/evolve/likelihood_tree.py
```diff
@@ -5,7 +5,7 @@
 
 import numpy
 
-from cogent3.util.modules import ExpectedImportError, importVersionedModule
+from . import likelihood_tree_numba as likelihood_tree_numba
```
GavinHuttley: Change that to `import likelihood_tree_numba as likelihood_tree` and change the usages

src/cogent3/evolve/likelihood_tree.py
```diff
@@ -177,10 +169,11 @@ class _PyLikelihoodTreeEdge(_LikelihoodTreeEdge):
     LOG_BASE = numpy.log(BASE)
 
     def sum_input_likelihoodsR(self, result, *likelihoods):
-        result[:] = 1.0
-        for (i, index) in enumerate(self.indexes):
-            result *= numpy.take(likelihoods[i], index, 0)
-        return result
+        return likelihood_tree_numba.sum_input_likelihoods(
+            numpy.ascontiguousarray(self.indexes),
```
GavinHuttley: given `self.indexes` is an attribute, call `numpy.ascontiguousarray` on creation of that attribute. But you will have to leave the `result` conversion as is

azure-pipelines.yml
GavinHuttley: It does limit to feature, so this should be removed on merge with develop

src/cogent3/evolve/likelihood_tree.py
```diff
@@ -177,10 +171,9 @@ class _PyLikelihoodTreeEdge(_LikelihoodTreeEdge):
     LOG_BASE = numpy.log(BASE)
 
     def sum_input_likelihoodsR(self, result, *likelihoods):
-        result[:] = 1.0
-        for (i, index) in enumerate(self.indexes):
-            result *= numpy.take(likelihoods[i], index, 0)
-        return result
+        return likelihood_tree.sum_input_likelihoods(
```
GavinHuttley: See if it's possible to make the creation of `result` ensure contiguous memory. In addition, add the following
```python
if not result.flags['C_CONTIGUOUS']:
    result = numpy.ascontiguousarray(result)
```

GavinHuttley: add the same check for `self.index`, may not be necessary but overhead very small

src/cogent3/evolve/likelihood_tree.py
```diff
@@ -172,7 +172,11 @@ class LikelihoodTreeEdge(_LikelihoodTreeEdge):
 
     def sum_input_likelihoodsR(self, result, *likelihoods):
         return likelihood_tree.sum_input_likelihoods(
-            self.indexes, numpy.ascontiguousarray(result), likelihoods,
+            self.indexes
+            if self.indexes.flags["C_CONTIGUOUS"]
```
GavinHuttley: put this conditional logic before the method call, makes the code easier to read.

PR #480, opened 2020-01-10
Added deepcopy method to _SequenceCollectionBase

jamesmartini:
Couldn't figure out how to copy `._repr_policy`.

The test created in `SequenceCollectionBaseTests` mirrors the one in `AlignmentTests` (This looks like it checks the data structures are new, and tests the usage of `sliced`).


src/cogent3/core/alignment.py
```diff
@@ -554,6 +554,17 @@ def copy(self):
...
         return result
 
+    def deepcopy(self, sliced=True):
+        """Returns deep copy of self."""
+        for seq in self.seqs:
+            try:
+                new_seq = seq.deepcopy(sliced=sliced)
+            except:
+                new_seq = seq.copy()
+        result = self.__class__(self.map, new_seq, moltype=self.moltype, info=self.info.copy())
```
GavinHuttley: this is not right -- it's providing only a single sequence, and the tests must therefore be incorrect.

- Create a dict `new_seqs` which has keys `seq.name` and values `new_seq`.
- Add the argument `force_same_data=True`, to avoid redundant copying
- since this is a deepcopy method, you must use `copy.deepcopy(self.info)`

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,64 @@ def setUp(self):
         self.b = Alignment(["AAA", "AAA"])
         self.c = SequenceCollection(["AAA", "AAA"])
 
+    def test_deepcopy(self):
+        """correctly deep copy aligned objects in an alignment"""
```
GavinHuttley: These tests are incorrect. The property of deepcopy means the new object at all layers must be different. 

1 - you need to use `self.Class` to create the object. Since the test suite has different subclasses, all of them will be executed
2 - skip annotation checking since that code is tested elsewhere
3 - you need to check the `sliced` argument does not raise exceptions
4 - add a `AlignmentTests.test_sliced_deepcopy` method. In that case, you should create a bigger alignment and slice it. Then compare the results of `deepcopy(sliced=False)` with `deepcopy(sliced=True)` the length of underlying sequences should differ.

src/cogent3/core/alignment.py
```diff
@@ -554,6 +554,23 @@ def copy(self):
...
 
+    def deepcopy(self, sliced=True, force_same_data=True):
+        """Returns deep copy of self."""
+        new_seqs = dict()
+        for seq in self.seqs:
+            try:
+                new_seq = seq.deepcopy(sliced=sliced)
+            except:
+                new_seq = seq.copy()
+            new_seqs.update({seq.name: new_seq})
```
GavinHuttley: replace with

```python
new_seqs[seq.name] = new_seq
```

src/cogent3/core/alignment.py
```diff
@@ -554,6 +554,23 @@ def copy(self):
...
+            seq_names = list(new_seqs.keys())
+
+        if force_same_data:
+            self._force_same_data(new_seqs, seq_names)
+        result = self.__class__(new_seqs)
```
GavinHuttley: this is missing key arguments -- see the equivalent call in the `copy()` method above.

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,16 @@ def setUp(self):
...
+        path = "data/brca1_5.paml"
+        aln = load_aligned_seqs(path, array_align=False, moltype="dna")
```
GavinHuttley: create a small dictionary of sequence data (look at one of the other test methods) and just pass that directly to self.Class, along with specifying the moltype. A dict with 2 sequences is enough, I think.

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,16 @@ def setUp(self):
...
+        aln = self.Class(aln.named_seqs)
+        aln = aln[20:30]
+        new_seq = aln.deepcopy()
+        self.assertEqual(aln.to_rich_dict(), new_seq.to_rich_dict())
```
GavinHuttley: using `to_rich_dict()` is a more complete check than what I suggested -- nice!

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,16 @@ def setUp(self):
         self.b = Alignment(["AAA", "AAA"])
         self.c = SequenceCollection(["AAA", "AAA"])
 
+    def test_deepcopy(self):
+        """correctly deep copy aligned objects in an alignment"""
+        path = "data/brca1_5.paml"
+        aln = load_aligned_seqs(path, array_align=False, moltype="dna")
+        aln = self.Class(aln.named_seqs)
+        aln = aln[20:30]
+        new_seq = aln.deepcopy()
```
GavinHuttley: change `new_seq` to `copied`

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,16 @@ def setUp(self):
         self.b = Alignment(["AAA", "AAA"])
         self.c = SequenceCollection(["AAA", "AAA"])
 
+    def test_deepcopy(self):
+        """correctly deep copy aligned objects in an alignment"""
+        path = "data/brca1_5.paml"
+        aln = load_aligned_seqs(path, array_align=False, moltype="dna")
+        aln = self.Class(aln.named_seqs)
```
GavinHuttley: change `aln` to `seqs`.

Variable names have meaning and when `self.Class` is a `SequenceCollection`, it's not an alignment.

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,16 @@ def setUp(self):
...
+        aln = aln[20:30]
+        new_seq = aln.deepcopy()
+        self.assertEqual(aln.to_rich_dict(), new_seq.to_rich_dict())
+        self.assertNotEqual(id(new_seq), id(aln))
```
GavinHuttley: you need some additional clauses. Specifically,
```python
for name in seqs.names:
    self.assertNotEqual(id(copied.named_seqs[name]), copied.named_seqs[name])
```

tests/test_core/test_alignment.py
```diff
@@ -2467,6 +2477,18 @@ def test_slice_align_info(self):
 class AlignmentTests(AlignmentBaseTests, TestCase):
     Class = Alignment
 
+    def test_sliced_deepcopy(self):
+        """correctly deep copy aligned objects in an alignment"""
+        path = "data/brca1_5.paml"
+        aln = load_aligned_seqs(path, array_align=False, moltype="dna")
```
GavinHuttley: same issue as before, just create a small sample data set as a dict, e.g. `{"seq1": "ACGTACGT", "seq2": "ACGTACGT", "seq3": "ACGTACGT"}`

tests/test_core/test_alignment.py
```diff
@@ -2467,6 +2477,18 @@ def test_slice_align_info(self):
...
+        aln = self.Class(aln.named_seqs)
+        aln = aln[20:30]
```
GavinHuttley:
```python
orig = self.Class(aln.named_seqs)
aln = orig[20:30]
```

tests/test_core/test_alignment.py
```diff
@@ -2467,6 +2477,18 @@ def test_slice_align_info(self):
...
+        for name in ["Mouse", "Human", "HowlerMon", "NineBande", "DogFaced"]:
```
GavinHuttley: the loop here is incorrect. You need to create 2 copied objects, e.g. `sliced = aln.deepcopy(sliced=True)` and `notsliced = ...` . Then test that `len(notsliced.get_seq('seq1')) == len(orig.get_seq('seq1')` and `len(sliced.get_seq('seq1')) != len(orig.get_seq('seq1')`

tests/test_core/test_alignment.py
```diff
@@ -244,6 +244,16 @@ def setUp(self):
         self.b = Alignment(["AAA", "AAA"])
         self.c = SequenceCollection(["AAA", "AAA"])
 
+    def test_deepcopy(self):
```
GavinHuttley: In this test, you should also exercise using the `sliced=True` argument. All we care about for this test case is that it does not fail. The real test of when it works is in the other test method below

tests/test_core/test_alignment.py
```diff
@@ -2479,15 +2479,14 @@ class AlignmentTests(AlignmentBaseTests, TestCase):
...
-        for name in ["Mouse", "Human", "HowlerMon", "NineBande", "DogFaced"]:
-            new_seq = aln.named_seqs[name].deepcopy()
-            self.assertEqual(len(new_seq.data), 10)
-            new_seq = aln.named_seqs[name].deepcopy(sliced=False)
-            self.assertEqual(len(new_seq.data), len(aln.named_seqs[name].data))
+        data = {"seq1": "ACGACGACG", "seq2": "ACGACGACG"}
+        orig = self.Class(data)
+        aln = orig[2:5]
+        for name in orig.names:
+            copied = aln.named_seqs[name].deepcopy()
```
GavinHuttley: delete this line

tests/test_core/test_alignment.py
```diff
@@ -2479,15 +2479,14 @@ class AlignmentTests(AlignmentBaseTests, TestCase):
...
-        aln = self.Class(aln.named_seqs)
-        aln = aln[20:30]
-        for name in ["Mouse", "Human", "HowlerMon", "NineBande", "DogFaced"]:
-            new_seq = aln.named_seqs[name].deepcopy()
-            self.assertEqual(len(new_seq.data), 10)
-            new_seq = aln.named_seqs[name].deepcopy(sliced=False)
-            self.assertEqual(len(new_seq.data), len(aln.named_seqs[name].data))
+        data = {"seq1": "ACGACGACG", "seq2": "ACGACGACG"}
+        orig = self.Class(data)
+        aln = orig[2:5]
```
GavinHuttley: after this
```python
sliced = aln.deepcopy(sliced=True)
notsliced = aln.deepcopy(sliced=False)
```

tests/test_core/test_alignment.py
```diff
@@ -2479,15 +2479,14 @@ class AlignmentTests(AlignmentBaseTests, TestCase):
...
+        for name in orig.names:
+            copied = aln.named_seqs[name].deepcopy()
+            self.assertNotEqual(len(copied.data), len(aln.named_seqs[name].data))
```
GavinHuttley: change to `self.assertLessThan()`
change copied to `sliced.named_seqs[name].data`, `notsliced.named_seqs[name].data`.

tests/test_core/test_alignment.py
```diff
@@ -2479,15 +2479,14 @@ class AlignmentTests(AlignmentBaseTests, TestCase):
...
+            copied = aln.named_seqs[name].deepcopy(sliced=False)
+            self.assertEqual(len(copied.data), len(aln.named_seqs[name].data))
```
GavinHuttley: test that `len(sliced.named_seqs[name].data)` equals 3, not len of the other data. Likewise, test that `len(notsliced.named_seqs[name].data)` is the length of `orig`.

src/cogent3/core/alignment.py
```diff
@@ -554,6 +554,23 @@ def copy(self):
         result = self.__class__(self, moltype=self.moltype, info=self.info)
         return result
 
+    def deepcopy(self, sliced=True, force_same_data=True):
```
GavinHuttley: remove `force_same_data` argument

src/cogent3/core/alignment.py
```diff
@@ -554,6 +554,23 @@ def copy(self):
...
+        new_seqs = dict()
+        for seq in self.seqs:
+            try:
+                new_seq = seq.deepcopy(sliced=sliced)
+            except:
+                new_seq = seq.copy()
+            new_seqs[seq.name] = new_seq
+            seq_names = list(new_seqs.keys())
+
+        if force_same_data:
```
GavinHuttley: this is violating the immutability principle of this object -- self should never modify it's underlying data

src/cogent3/core/alignment.py
```diff
@@ -554,6 +554,23 @@ def copy(self):
...
+            self._force_same_data(new_seqs, seq_names)
+        result = self.__class__(new_seqs, moltype=self.moltype, info=deepcopy(self.info))
```
GavinHuttley: this call should set `force_same_data=True`

tests/test_core/test_alignment.py
```diff
@@ -2469,6 +2479,31 @@ def test_slice_align_info(self):
 class AlignmentTests(AlignmentBaseTests, TestCase):
     Class = Alignment
 
+    def test_sliced_deepcopy(self):
+        """correctly deep copy aligned objects in an alignment
+
```
GavinHuttley: test method docstrings should be a single short line because it get's displayed in the terminal on failures. Make the other remarks a comment. But that said, I don't think force_same data should be part of the interface, so those test modifications should be reversed.

Issue #411, opened 2019-11-26
resolve python 3.8 SyntaxWarning's


GavinHuttley:
Is there a tool for identifying all `SyntaxWarning` occurrences of these in the codebase and correcting them.

Python 3.8 raises some syntax warning's which should be addressed. For example
```
SyntaxWarning: "is not" with a literal. Did you mean "!="?
```
These warnings occur when running the test suite using python 3.8.

GavinHuttley:
We need to monitor [pyflakes](https://github.com/PyCQA/pyflakes/issues/483), they will make a release with a patch for this.

GavinHuttley:
fixed in 757e0a97b45a1df5df72ad83c82efc26896bd2dc

PR #493, opened 2020-01-16
Explored SyntaxWarnings associated to using py38

jamesmartini:
The python syntax warning
`SyntaxWarning: "is not" with a literal. Did you mean "!="?`
arises because 'is not' and 'is' are used to check whether two objects are exactly the same (their id and values are equal). The instances where these expressions are used in cogent have been replaced by '!=' and '==' to make the correct comparison.


GavinHuttley:
does not actually fix the issue completely, we will wait for a 3rd party tool for this (see #411)

PR #514, opened 2020-01-31
Implemented Hertz-Stormo alignment quality metric and testing

jamesmartini:


src/cogent3/core/alignment.py
```diff
@@ -2235,6 +2236,34 @@ class AlignmentI(object):
...
+        Computes the alignment quality for an alignment based on eq. (2) in noted reference.
+
+        Parameters
+        ----------
+        equifreq_mprobs : bool
+            If true, specifies equally frequent motif probabilities.
+
+        Notes
+        -----
+        G. Z. Hertz, G. D. Stormo - Published 1999, Bioinformatics, vol. 15 pg. 563-577.
```
GavinHuttley: add brief description in notes about the statistic, i.e. it's a log-likelihood ratio of the observed alignment column freqs versus the expected... be sure and mention it's log2

src/cogent3/core/alignment.py
```diff
@@ -2235,6 +2236,34 @@ class AlignmentI(object):
...
+        """
+        counts = self.counts_per_pos()
```
GavinHuttley: make sure the alignment has data. If counts total is 0, it doesn't. In which case return None

tests/test_core/test_alignment.py
```diff
@@ -1530,6 +1530,27 @@ class AlignmentBaseTests(SequenceCollectionBaseTests):
...
+
+        aln = make_aligned_seqs(["AAAC",
+                                 "ACGC",
+                                 "AGCC",
+                                 "A-TC"], moltype="dna")
+        got = aln.alignment_quality(equifreq_mprobs=False)
+        expect = 2 * log2(1 / 0.4) + log2(1 / (4 * 0.4)) + (1 / 2) * log2(1 / (8 / 15)) + (
+                1 / 4) * log2(1 / (4 / 15))
+        assert_allclose(got, expect)
+
```
GavinHuttley: 1. test when `alignment' is all gaps
2. when alignment has a single sequence
3. when alignment has two seqs, but one is all gaps

2 and 3 should evaluate when equifreq True or False

PR #586, opened 2020-03-26
Evaluated and tested base draw objects, fixes #554

### LINKED ISSUE #554: increase test coverage of cogent3.draw
look at base objects first and establish logical consistency across how they're being used

think about whether the design is right!

think like a user in a jupyter notebook!

u6052029:


src/cogent3/draw/drawable.py
```diff
@@ -167,7 +168,10 @@ def __init__(
         xtitle=None,
         ytitle=None,
     ):
-        self._traces = traces or []
+        if isinstance(traces, list):
+            self._traces = [UnionDict(trace) for trace in traces]
+        else:
+            self._traces = []
         title = title if title is None else dict(text=title)
```
GavinHuttley: I think we need to make this more general. We either have the default value, in which case it `is None`, otherwise it could be a series of dicts, so I suggest changing this to:
```python
if traces is None:
    self._traces = []
else:
    try: 
        self._traces = [UnionDict(trace) for trace in traces]
    except ValueError as msg:
        raise TypeError(f"expected a series of dicts, got {traces}")
```
Don't check what type, if a user inputs a dict this step should fail because iter on a dict should fail, for instance

Issue #17, opened 2019-03-20
Time heterogeneity - can we not have a shorter (and easier to spell) argument name.


GavinHuttley:
**[Original report](https://bitbucket.org/pycogent3/cogent3-dev/issue/17) by James Kondilios (Bitbucket: [5c0608ef6b98a11ccd3644e1](https://bitbucket.org/5c0608ef6b98a11ccd3644e1), GitHub: [JamesKondilios](https://github.com/JamesKondilios)).**

----------------------------------------

`time_heterogeneity=None` is an arg to the `evo.model` constructer and its just too bloody hard to spell, we're not arts researchers.

PR #630, opened 2020-05-03
updated repo to using latest Numba version

wjjmjh:
1, added dependency Numba and removed dependency sphinx_rtd_theme in setup.py;
2, updated tox file to adopt latest Numba version and construct a valid testing environment;
3, to make repo compatible with newest features of Numba,  pairwise_pogs_numba, and pairwise_seqs_numba updated, basically, it's about more restricted type checking;
4, also updated feature branch up-to-date.

Notes: this pull request possibly related to issue [#17](https://github.com/cogent3/c3dev/issues/17) of [c3dev](https://github.com/cogent3/c3dev)


setup.py
```diff
@@ -169,7 +169,7 @@ def CythonExtension(module_name, **kw):
     ],
     packages=find_packages(where="src"),
     package_dir={"": PACKAGE_DIR},
-    install_requires=["numba==0.48.0", "numpy", "scitrack", "tqdm", "tinydb"],
```
GavinHuttley, line 172: OK, now that I see the reason for the errors is a refactor in `numba`, we **do** need to define a minimum version here, I think `numba>0.48` will do it.

tox.ini
```diff
@@ -3,8 +3,8 @@ envlist = py36, py37, py37mpi, py38, doctest
 
 [testenv]
 passenv = *
-deps = numpy
-       numba==0.48.0
+deps = numba
```
GavinHuttley, line 6: this can be left as is

Issue #687, opened 2020-06-30
load_unaligned_seqs and load_aligned_seqs should support loading from cogent3 json


GavinHuttley:
The two functions to be modified are in `cogent3.__init__`.

Here's the current paradigm for how those functions work. Both functions call `cogent3.parse.sequence.FromFilenameParser`, which identifies the appropriate file format from the filename and then is a generator that yields (name, seq). This doesn't make sense for cogent3 formatted json files since those files are going to contain all the info to encode and deserialise an object.

I think add a new function, `cogent3.parse.sequence.load_from_json(filename, classes)`. The `load_from_json` function takes 2 arguments: `filename` is the path to the json file (opened using `cogent3.util.misc.open_`), `classes` is a series of the cogent3 types. The loaded json dict must have a type value that matches these inputs. If it does not, it should raise a TypeError and specify what the type in the file is and what's allowed.

Note that the function must also support the case where the json is structured as a cogent3 data_store record. In other words, `cogent3.app.data_store.load_record_from_json` correctly returns the identifier, data and a boolean indicating whether the object was "completed". If the latter is False, it should raise a TypeError.

Now back to the original goal of supporting loading from json. The `load_aligned_seqs` function needs to do the following:

1. Get the filename suffix using `cogent3.util.misc.get_format_suffixes` and if it's json does the following
2. calls `load_from_json` with the path and classes=(Alignment, ArrayAlignment)
3. returns the result of (2) (does not use the `make_aligned_seqs` function)

The solution for `load_unaligned_seqs` is the same as above except classes=(SequenceCollection,).

NOTE: the alignment and sequence types are already in the `cogent3.__init__` name space.

## Testing

- Add tests as methods on `tests.test_core.test_core_standalone.TestConstructorFunctions`
- Test the aligned and unaligned cases as separate methods
- create json files on the fly in a temp directory (see `test_data_store.py` line 95 for an example)
- checks case where the json was constructed by `cogent3.app.data_store.make_record_for_json`
- checks case where the json is for a `NotCompleted` object

PR #691, opened 2020-07-04
#687 enabled json write/load support on major objects.

wjjmjh:


src/cogent3/parse/sequence.py
```diff
@@ -90,6 +91,52 @@ def format_from_filename(filename, format=None):
...
+    ----------
+    filename: name of the json file
+    classes: A series of the Cogent3 types, for example: (Alignment, ArrayAlignment)
+
+    """
+    assert all(
+        (isinstance(klass, type) for klass in classes)
+    ), "classes should be a series of Cogent3 types, for example: (Alignment, ArrayAlignment)"
+
+    f = open_(filename)
```
GavinHuttley, line 107: open files as a context, e.g.

```python
with open_(filename) as f:
    content = json.loads(f.read())
```

src/cogent3/core/alignment.py
```diff
@@ -73,6 +73,7 @@
     extend_docstring_from,
     get_format_suffixes,
     get_object_provenance,
+    open_,
```
GavinHuttley, line 76: I think we should be using the `atomic_write` class instead for writing to files, so import that class instead.

src/cogent3/core/alignment.py
```diff
@@ -1245,15 +1246,21 @@ def write(self, filename=None, format=None, **kwargs):
         if filename is None:
             raise DataError("no filename specified")
 
+        suffix, cmp_suffix = get_format_suffixes(filename)
+        if format is None and suffix:
+            format = suffix
+
+        if format == "json":
+            f = open_(filename, "wt")
```
GavinHuttley, line 1254: using the context style and `atomic_write`, change this to:
```python
with atomic_write(filename mode="wt") as f:
    f.write(self.to_json())
```
then delete the `f.close()` line

src/cogent3/parse/sequence.py
```diff
@@ -90,6 +91,52 @@ def format_from_filename(filename, format=None):
...
+    content = json.loads(f.read())
+    f.close()
+    # loads from a record json file (generated by make_record_for_json)
+    if set(content.keys()) == {"identifier", "data", "completed"}:
```
GavinHuttley, line 111: we can simplify the logic here. First, just use the `load_record_from_json` function since if the nature of the records changes, that's where that logic will be specified.

```python
try:
    _, data, completed = load_record_from_json(content)
    if not completed:
        raise TypeError("json file is a record for type NotCompleted.")
except KeyError:
    data = content
```
This logic produces a `data ` variable that should be the same if it comes from a db record or from directly writing results of `to_json()`.

src/cogent3/parse/sequence.py
```diff
@@ -90,6 +91,52 @@ def format_from_filename(filename, format=None):
...
+    Parameters
+    ----------
+    filename: name of the json file
+    classes: A series of the Cogent3 types, for example: (Alignment, ArrayAlignment)
+
+    """
+    assert all(
+        (isinstance(klass, type) for klass in classes)
+    ), "classes should be a series of Cogent3 types, for example: (Alignment, ArrayAlignment)"
+
```
GavinHuttley, line 106: I think we use a string comparison for types, since the dict in the json file encodes this as a string. So let's create a set object:
```python
valid_types = {get_object_provenance(klass) for klass in classes}
```

src/cogent3/parse/sequence.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 """Classes for reading multiple sequence alignment files in different formats."""
 
+import json
```
GavinHuttley, line 4: add import of `deserialise_object`

src/cogent3/parse/sequence.py
```diff
@@ -90,6 +91,52 @@ def format_from_filename(filename, format=None):
...
+    f = open_(filename)
+    content = json.loads(f.read())
+    f.close()
+    # loads from a record json file (generated by make_record_for_json)
+    if set(content.keys()) == {"identifier", "data", "completed"}:
+        content["data"] = json.loads(content["data"])
+        if not content["completed"]:
+            raise TypeError("input record json file is uncompleted.")
+
+        for klass in classes:
```
GavinHuttley, line 116: the check for correct json contents is then just
```python
type_ = data.get("type", None)
if type_ is None:
    raise TypeError("json does not contain 'type' key")

if type_ not in valid_types:
    raise TypeError(f"Invalid data type: {type_} is not one of {valid_types}")
```

src/cogent3/parse/sequence.py
```diff
@@ -90,6 +91,52 @@ def format_from_filename(filename, format=None):
...
+            if content["data"]["type"] == get_object_provenance(klass):
```
GavinHuttley, line 117: Completion of the function is then
```python
return deserialise_object(data)
```

tests/test_core/test_alignment.py
```diff
@@ -1510,6 +1510,15 @@ def test_construction(self):
         seq2 = make_seq("AC", name="seq2")
         coll = SequenceCollection(data=[seq1, seq2])
 
+    def test_write_to_json(self):
+        # test writing to json file
+        aln = self.Class([("a", "AAAA"), ("b", "TTTT"), ("c", "CCCC")])
+        fn = mktemp(suffix=".json")
+        aln.write(fn)
+        result = load_unaligned_seqs(fn)
+        self.assertEqual(str(result), ">a\nAAAA\n>b\nTTTT\n>c\nCCCC\n")
```
GavinHuttley, line 1519: don't check str(result) since underlying structure based on dict and thus the record order is not guaranteed. Compare dict instead, i.e.

```python
self.assertEqual(result.to_dict(), dict(a="AAAA", ...)
```

tests/test_core/test_alignment.py
```diff
@@ -1510,6 +1510,15 @@ def test_construction(self):
         seq2 = make_seq("AC", name="seq2")
         coll = SequenceCollection(data=[seq1, seq2])
 
+    def test_write_to_json(self):
+        # test writing to json file
+        aln = self.Class([("a", "AAAA"), ("b", "TTTT"), ("c", "CCCC")])
+        fn = mktemp(suffix=".json")
```
GavinHuttley, line 1516: I think it's better to use a temporary directory as a context, since the directory and all it's contents cleaned up after exiting the context, e.g.

```python
with TemporaryDirectory('.') as dirname:
    fn = pathlib.Path(dirname) / "sample.json" # use pathlib because it handles windows/linux paths
    # other operations

# and here the directory and all it's contents have been deleted
```

tests/test_core/test_alignment.py
```diff
@@ -2565,6 +2574,15 @@ def test_seqlogo(self):
         aln = self.Class(data)
         logo = aln.seqlogo()
 
+    def test_write_to_json(self):
+        # test writing to json file
+        aln = self.Class([("a", "AAAA"), ("b", "TTTT"), ("c", "CCCC")])
+        fn = mktemp(suffix=".json")
```
GavinHuttley, line 2580: same issue of using Tmp directories

tests/test_core/test_alignment.py
```diff
@@ -1510,6 +1510,15 @@ def test_construction(self):
         seq2 = make_seq("AC", name="seq2")
         coll = SequenceCollection(data=[seq1, seq2])
 
+    def test_write_to_json(self):
+        # test writing to json file
+        aln = self.Class([("a", "AAAA"), ("b", "TTTT"), ("c", "CCCC")])
+        fn = mktemp(suffix=".json")
+        aln.write(fn)
+        result = load_unaligned_seqs(fn)
```
GavinHuttley, line 1518: Don't use `load_unaligned_seqs` for this. Directly use the `json` module, since you need to explicitly check that you wrote json content. I suggest the following:
```python
got = json.load(fn)
self.assertEqual(got, aln.to_rich_dict()) # that dict method is what's called by `to_json()`
```

tests/test_core/test_alignment.py
```diff
@@ -2565,6 +2574,15 @@ def test_seqlogo(self):
         aln = self.Class(data)
         logo = aln.seqlogo()
 
+    def test_write_to_json(self):
+        # test writing to json file
+        aln = self.Class([("a", "AAAA"), ("b", "TTTT"), ("c", "CCCC")])
+        fn = mktemp(suffix=".json")
+        aln.write(fn)
+        result = load_aligned_seqs(fn)
```
GavinHuttley, line 2582: same issue of checking the json content directly, don't use `load_aligned_seqs`

tests/test_core/test_core_standalone.py
```diff
@@ -114,6 +118,84 @@ def test_load_aligned_seqs(self):
...
+        """test loading an unaligned object from json file"""
+        with TemporaryDirectory(dir=".") as dirname:
+            json_path = os.path.join(dirname, "unaligned.json")
+            path = os.path.join(data_path, "brca1_5.paml")
+            unaligned = load_unaligned_seqs(path)
+            unaligned.write(json_path)
+
+            got = load_unaligned_seqs(json_path)
+            self.assertIsInstance(got, SequenceCollection)
+            self.assertTrue("Mouse" in got.to_dict())
```
GavinHuttley, line 131: change the `assertTrue` to
```python
self.assertEqual(got.to_dict(), unaligned.to_dict())
```

tests/test_core/test_core_standalone.py
```diff
@@ -114,6 +118,84 @@ def test_load_aligned_seqs(self):
...
+                out.write(json.dumps(uncompleted_record))
+            with open(completed_record_path, "w") as out:
+                out.write(json.dumps(completed_record))
+            # tests when provided record json file is uncompleted
+            with self.assertRaises(TypeError):
+                load_unaligned_seqs(uncompleted_record_path)
+            # tests when provided record json is completed
+            got = load_unaligned_seqs(completed_record_path)
+            self.assertIsInstance(got, SequenceCollection)
+            self.assertTrue("Human" in got.to_dict())
```
GavinHuttley, line 148: compare `to_dict()` result, as above

tests/test_core/test_core_standalone.py
```diff
@@ -114,6 +118,84 @@ def test_load_aligned_seqs(self):
...
+            alignment_json_path = os.path.join(dirname, "alignment.json")
+            alignment.write(alignment_json_path)
+            array_alignment = load_aligned_seqs(path, moltype="dna")
+            array_alignment_json_path = os.path.join(dirname, "array_alignment.json")
+            array_alignment.write(array_alignment_json_path)
+            # tests case Alignment
+            got = load_aligned_seqs(alignment_json_path)
+            self.assertIsInstance(got, Alignment)
+            self.assertEqual(got.moltype.label, "dna")
+            self.assertTrue("Human" in got.to_dict())
```
GavinHuttley, line 165: compare `to_dict()` result, as above

tests/test_core/test_core_standalone.py
```diff
@@ -114,6 +118,84 @@ def test_load_aligned_seqs(self):
...
+                out.write(json.dumps(uncompleted_record))
+            with open(completed_record_path, "w") as out:
+                out.write(json.dumps(completed_record))
+            # tests when provided record json file is uncompleted
+            with self.assertRaises(TypeError):
+                load_unaligned_seqs(uncompleted_record_path)
+            # tests when provided record json is completed
+            got = load_aligned_seqs(completed_record_path)
+            self.assertIsInstance(got, ArrayAlignment)
+            self.assertTrue("DogFaced" in got.to_dict())
```
GavinHuttley, line 189: compare `to_dict()` result, as above

src/cogent3/parse/sequence.py
wjjmjh: Hi Gavin! thanks again for your detailed instructions! 

This line of codes just checking if users input correct `classes` parameter as it should be a series of types.

GavinHuttley: correct, but my suggestion is to create this set object after you've checked the user has passed in types (so add this code below yours)

Issue #693, opened 2020-07-06
support for cogent3 reading/writing Tree json


GavinHuttley:
split out from #646. Complete after #692. modify `cogent3.__init__.load_tree` (and `cogent3.core.tree.TreeNode.write`) as per the others.

PR #696, opened 2020-07-11
#693 enabled json support on Tree object.

wjjmjh:


src/cogent3/core/tree.py
```diff
@@ -1448,6 +1452,12 @@ def write(self, filename, with_distances=True, format=None):
             the filename suffix. All attributes are saved in the xml format.
```
GavinHuttley, line 1452: add json to docstring, as per

```
default is newick, xml and json are alternates. Argument overrides
```

tests/test_app/test_tree.py
```diff
@@ -242,6 +250,22 @@ def test_uniformize_tree(self):
         u_c = make_uniform(c).get_newick()
         self.assertFalse(u_a == u_c)
 
+    def test_write_to_json(self):
```
GavinHuttley, line 253: move this test_method to `test_core/test_tree.py`

Issue #628, opened 2020-05-01
Alignment repr in jupyter not using repr settings


GavinHuttley:
The value in `_repr_policy` for `num_seqs` is not being used. To reproduce, in a jupyter notebook

```python
from cogent3 import load_aligned_seqs

aln = load_aligned_seqs("/Users/gavin/repos/Cogent3-hg/tests/data/brca1.fasta", moltype="dna")
aln.set_repr_policy(num_seqs=2)
aln
```
Produces display of entire alignment.

PR #698, opened 2020-07-25
#628 bug

wjjmjh:


tests/test_core/test_alignment.py
```diff
@@ -2601,6 +2601,18 @@ def test_slice_align_info(self):
         self.assertTrue(len(sub_align) == 3)
         self.assertEqual(sub_align.info["key"], "value")
 
+    def test_to_str(self):
```
GavinHuttley, line 2604: this test not correct. It should be testing `aln._repr_html_()`. Probably easiest to check by counting the number of `<tr>` elements.

src/cogent3/core/alignment.py
```diff
@@ -2774,15 +2774,15 @@ def _get_raw_pretty(self, name_order):
 
     def _repr_html_(self):
         html = self.to_html(
-            name_order=self.names[: self._repr_policy["num_seqs"]],
```
GavinHuttley, line 2777: should not be deleting `name_order`. This specifies the row order in the display

wjjmjh: Hi Gavin, `name_order` isn't deleted, I came up a fix that lets method `to_html`(a little bit below this code snippet) take `num_seqs` as an input parameter instead of `name_order` (sliced self.names), then method `to_html` will produce `name_order` inside it's logic. specifically, after 
`name_order = list(self.names)`
`name_order.remove(ref)`
`name_order.insert(0, ref)` 
Please let me know any issues!

wjjmjh: Apologies that I forgot updating the function docstring accordingly, let me know what do you think Gavin if you are happy with this fix, I am gonna push my updated docstring. cheers!

GavinHuttley: But `name_order` is part of the `to_html()` public API and an argument users will want to be able to set. So better for the `_repr_html_ ()` calling code provide the correct number of names.

Issue #699, opened 2020-07-26
Table.head() failure when number of rows < default


GavinHuttley:
```python
from cogent3 import make_table

table = make_table(data=dict(a=['a'], b=['b']))
table.head()
```

raises an index error. I fully expect that `table.tail()` will fail also.

wjjmjh:
Hi Gavin! this works well for me, no index error raised, however, I think this could potentially be a bug:
>>> table = make_table(data=dict(a=[i for i in range(10)], b=[i for i in range(10)]))
>>> table.head(nrows=5)

it turns out the output has got ten rows.

wjjmjh:
oops, sorry Gavin, I spotted the bug, only raised when using a notebook. 

I am on the way of fixing, cheers.

PR #703, opened 2020-08-04
#699 bug

wjjmjh:


tests/test_util/test_table.py
```diff
@@ -1280,6 +1280,12 @@ def test_head(self):
         self.assertEqual(head.data.shape[0], 3)
         self.assertEqual(len(head.output.splitlines()), 9)
         self.assertEqual(head.data.tolist(), self.t1_rows[:3])
+        # tests when number of rows < default
+        t = make_table(data=dict(a=["a"], b=["b"]))
+        t.head()
```
GavinHuttley, line 1285: these calls will be writing to stdout. Might be wise to redirect them to null by putting this in a content manager. See if pytest has a function to help with this.

tests/test_util/test_table.py
```diff
@@ -1294,6 +1300,12 @@ def test_tail(self):
         self.assertEqual(tail.data.shape[0], 3)
         self.assertEqual(len(tail.output.splitlines()), 9)
         self.assertEqual(tail.data.tolist(), self.t1_rows[-3:])
+        # tests when number of rows < default
+        t = make_table(data=dict(a=["a"], b=["b"]))
+        t.tail()
```
GavinHuttley, line 1305: same as above wrt to stdout

tests/test_util/test_table.py
```diff
@@ -1275,17 +1276,18 @@ def test_head(self):
         display = table.display
         head = TrapOutput()
```
GavinHuttley, line 1277: this is awesome, I think it means you can delete the entire TrapOutput class. Can you delete it and all uses and replace with this contextlib approach.

wjjmjh: Not a problem at all! thank you Gavin!

PR #705, opened 2020-08-06
ENH: coloured html representation for sequences, fixes #704

### LINKED ISSUE #704: add support for coloured representations of sequences
Add to `cogent3.core.sequence.Sequence` a `to_html()` and `_repr_html_()` methods. These should work like their counterparts on `AlignmentI`, i.e. using `self.moltype.get_css_style()` to get colour scheme.

The `to_html()` should match the signature of the alignment method (except things that don't make sense, like `name_order` etc..).

The `_repr_html_()` should be identical to what is produced by the `__repr__()` method, with the exception of colouring the text.

u6052029:
[ADDED] _repr_policy attribute to cogent3.core.sequence.Sequence
[ADDED] _repr_html method to cogent3.core.sequence.Sequence
[ADDED] to_html method to cogent3.core.sequence.Sequence


src/cogent3/core/sequence.py
```diff
@@ -1090,6 +1092,111 @@ def annotate_matches_to(self, pattern, annot_type, name, allow_multiple=False):
 
         return annot
 
+    def _repr_html_(self):
+        html = self.to_html(limit=self._repr_policy["num_pos"],)
```
GavinHuttley, line 1096: no quite right. Needs to truncate the sequence, as per the `__repr__` method. For example
```python
from cogent3 import make_seq

seq = make_seq("ACGGTGGGGGGGGG", moltype="dna")
seq
```
Displays `DnaSequence(ACGGTGG... 14)`. What I'm asking is the `ACGGTGG` be coloured. But see below.

src/cogent3/core/sequence.py
```diff
@@ -1090,6 +1092,111 @@ def annotate_matches_to(self, pattern, annot_type, name, allow_multiple=False):
...
+        return html
+
+    def to_html(
```
GavinHuttley, line 1099: This method looks like a near duplicate of the corresponding method on `AlignmentI`. This tells me we should move the alignment method code into a function inside `cogent3.format.alignment`. Call the function `rich_html`. 

Then both alignment and sequence `to_html()` methods call that function.

u6052029: Similarly, would we need a cogent3.format.sequence?

u6052029: Are you referring to the rich_html function in table.py?

PR #728, opened 2020-08-27
Added show_shape to Table _repr_policy, fixes #726

### LINKED ISSUE #726: allow turning off display of table shape in repr and _repr_html_
Add `show_shape` argument to `Table.set_repr_policy()`. Use this in the above named functions to include/exclude shape info in their output.

- add to the method docstring
- add a test of effect

u6052029:


src/cogent3/util/table.py
```diff
@@ -645,7 +645,10 @@ def __init__(
         self._column_templates = column_templates or {}
         # define the repr() display policy
         random = 0
-        self._repr_policy = dict(head=None, tail=None, random=random)
+        show_shape = True
```
GavinHuttley, line 648: defining as separate variable not required

src/cogent3/util/table.py
```diff
@@ -792,6 +796,8 @@ def _get_repr_(self):
             header = tuple(rows.keys())
         table = self.__class__(header=header, data=rows, **kwargs)
         table._column_templates.update(self._column_templates)
+        if not show_shape:
```
GavinHuttley, line 799: I think using show_shape needs to be in the __repr__ and _repr_html_ functions since entire blocks of text will need to be excluded. Just having an empty string I don't think does it -- unless you've check in the interactive interpreter / jupyter?

GavinHuttley (changes requested):
actual usage methods (i.e. __repr__ and _repr_html_) should use this information, so tests need to change also

GavinHuttley (commented):
Are all of those changes covered by tests?

u6052029:
Use of the show_shape policy appears in __repr__ and _to_html_, both are tested in the latest commit in this pr. I have removed the "include_shapes" parameter in _to_html_ as it seems to have the same function as show_shape. All the changes in 
40a5e37 involve replacing the "include_shapes" parameter in _to_html_ with the new policy.

Issue #715, opened 2020-08-20
Alignment._repr_html_ displays incorrect number of sequences


GavinHuttley:
The attached [mouse-release-92-v2.fasta.zip](https://github.com/cogent3/cogent3/files/5105812/mouse-release-92-v2.fasta.zip) alignment has 21952 sequences. Decompress it and load

```python
from cogent3 import load_aligned_seqs
aln = load_aligned_seqs("mouse-release-92-v2.fasta", moltype="dna")
repr(aln)
```
Returns `'21952 x 1000 dna alignment...`, but
```python
aln._repr_html_().splitlines()[-2]
```
returns `'<p><i>10 x 1000 (truncated to 10 x 60) dna alignment</i></p>'`

GavinHuttley:
@wjjmjh I've pushed a fix, but my patch needs testing

PR #730, opened 2020-08-31
TST: tests _repr_html_ displaying correct number of sequences, fixed …

wjjmjh:
…#715


GavinHuttley (commented):
Can you confirm your revised test fails on the version of the code before my patch?

Issue #716, opened 2020-08-20
add support to open_ for zipped files


GavinHuttley:
Currently, `cogent3.util.misc.open_` supports only `gzip` and `bzip2` compression formats.

Add support for standard `zip`.

PR #751, opened 2020-09-07
ENH #716

wjjmjh:
implemented func to open a single zip file.


src/cogent3/util/misc.py
```diff
@@ -130,10 +131,20 @@ def bytes_to_string(data):
     return data
 
 
+def zip_open(filename, mode="r", **kwargs):
```
GavinHuttley, line 134: rename function to `open_zip()`

src/cogent3/util/misc.py
```diff
@@ -130,10 +131,20 @@ def bytes_to_string(data):
...
+    """open a zip-compressed file"""
+    if mode == "rt":
+        mode = "r"
+    with ZipFile(filename) as zf:
```
GavinHuttley, line 138: I think we should assert, on opening a zip that it contains a single member. If not, raise a ValueError.

src/cogent3/util/misc.py
```diff
@@ -130,10 +131,20 @@ def bytes_to_string(data):
     return data
 
 
+def zip_open(filename, mode="r", **kwargs):
+    """open a zip-compressed file"""
+    if mode == "rt":
```
GavinHuttley, line 136: Why block mode "rt"?

tests/test_util/test_misc.py
```diff
@@ -536,6 +541,27 @@ def test_path_exists(self):
         # or string instance
         self.assertTrue(path_exists(__file__))
 
+    def test_open_zipped_file(self):
+        with TemporaryDirectory(dir=".") as dirname:
```
GavinHuttley, line 545: add tests for ensuring ValueError raised when multiple files in archive

src/cogent3/util/misc.py
```diff
@@ -130,10 +131,20 @@ def bytes_to_string(data):
     return data
 
 
+def zip_open(filename, mode="r", **kwargs):
+    """open a zip-compressed file"""
```
GavinHuttley, line 135: add to docstring
```
Note
----
Raises ValueError if archive has > 1 record
```

GavinHuttley (changes requested):
have a question and a couple of small changes

src/cogent3/util/misc.py
wjjmjh: Hi Gavin, I added this conversion simple because Python Zip only supports "r" and "w" modes,
```
        if mode not in {"r", "w"}:
            raise ValueError('open() requires mode "r" or "w"')
```
but we have got mode "rt" as default mode in function `open_`

GavinHuttley: OK, let's just trim off trailing "t".

Issue #774, opened 2020-09-18
likelihood function to_rich_dict() fails when there are multiple alignments


GavinHuttley:
The likelihood function can jointly fit a function to multiple alignments. But some methods break when that has been done.

```python
from cogent3 import load_aligned_seqs, make_tree, get_model

aln = load_aligned_seqs("~/repos/Cogent3/doc/data/long_testseqs.fasta")
half = len(aln) // 2
aln1 = aln[:half]
aln2 = aln[half:]

loci_names = ["1st-half", "2nd-half"]
loci = [aln1, aln2]
tree = make_tree(tip_names=aln.names)

mod = get_model("HKY85")
lf = mod.make_likelihood_function(tree, loci=loci_names, digits=2, space=3)
lf.set_param_rule("length", is_independent=False)
lf.set_param_rule("kappa")
lf.set_alignment(loci)
d = lf.to_rich_dict()
```

## How to fix

This will be tricky! You definitely need to talk with me about this.

PR #787, opened 2020-10-01
Bug #774

wjjmjh:


src/cogent3/evolve/likelihood_function.py
```diff
@@ -896,8 +896,22 @@ def to_rich_dict(self):
                 edge_attr[edge]["length"] = None
 
         model = self._model.to_rich_dict(for_pickle=False)
-        alignment = self.get_param_value("alignment").to_rich_dict()
-        mprobs = self.get_motif_probs().to_dict()
+
+        aln_defn = self.defn_for["alignment"]
+        if len(aln_defn.index) == 1:
```
GavinHuttley, line 901: we need a test that triggers this condition

src/cogent3/evolve/likelihood_function.py
```diff
@@ -896,8 +896,22 @@ def to_rich_dict(self):
...
+            alignment = self.get_param_value("alignment").to_rich_dict()
+            mprobs = self.get_motif_probs().to_dict()
+        else:
+            alignment = {a["locus"]: a["value"] for a in aln_defn.get_param_rules()}
+            mprobs = self.get_motif_probs()
+            if len(alignment) == 1:
```
GavinHuttley, line 907: de-indent this condition – it should not be nested since it should be dealt with by the condition at line 901

tests/test_evolve/test_likelihood_function.py
```diff
@@ -1352,6 +1352,23 @@ def test_to_rich_dict(self):
...
+        model = get_model("HKY85")
+        lf = model.make_likelihood_function(tree, loci=loci_names)
+        lf.set_alignment(loci)
+        for i, loci_name in enumerate(loci_names):
+            d = lf.to_rich_dict()
+            alignment = d["alignment"]
+            motif_probs = d["motif_probs"]
+            self.assertEqual(alignment[loci_name], loci[i].to_rich_dict())
+            self.assertEqual(motif_probs[loci_name], loci[i].get_motif_probs())
+
```
GavinHuttley, line 1371: add a test for when there's a single alignment.

In the new test, line 1363 become
```python
lf = model.make_likelihood_function(tree)
```
and line 1364 becomes
```python
lf.set_alignment(aln1)
```
There's no loop, and the result of `d["alignment"]` should equal `aln1.to_rich_dict()`

GavinHuttley (changes requested):
a few tweaks plus an additional test

Issue #788, opened 2020-10-01
discontinue cogent3.maths.stats.test.median


GavinHuttley:
Our implementation predates numpy's. So mark it as discontinued in version 2021.11.

See [c3dev wiki page](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) for instructions on how to deprecate.

GavinHuttley:
fixed in 1d5ff1ba2b3d42736f8f04de8507b5cd585b4fe9

PR #790, opened 2020-10-02
#788 Discontinued cogent3.maths.stats.test.median

Lmaster20:
No other uses of median() found.


GavinHuttley:
Thank you @Lmaster20 !

Issue #793, opened 2020-10-03
deserialise.deserialise_likelihood_function needs to handle mutliple alignments


GavinHuttley:
The fix to #774 returns a rich dict with multiple elements under the "alignment" key. The deserialise method now needs to handle correctly deserialising each of those and applying them to the deserialised likelihood function.

Requires new tests (obviously). Suggest modifications of those used to fix #774.

PR #795, opened 2020-10-08
Bug #793

wjjmjh:


src/cogent3/util/deserialise.py
```diff
@@ -262,17 +262,28 @@ def deserialise_likelihood_function(data):
...
     motif_probs = data.pop("motif_probs")
     param_rules = data.pop("param_rules")
     name = data.pop("name", None)
     lf = model.make_likelihood_function(tree, **constructor_args)
     lf.set_name(name)
-    lf.set_alignment(aln)
+    lf = model.make_likelihood_function(tree, **constructor_args)
+    if isinstance(constructor_args["loci"], list):
+        align = data["alignment"]
+        loci = [deserialise_seq_collections(align[k]) for k in align]
```
GavinHuttley, line 275: make this
```python
aln = [deserialise_seq_collections(align[k]) for k in align]
mprobs = [motif_probs[k] for k in motif_probs]
```

src/cogent3/util/deserialise.py
```diff
@@ -262,17 +262,28 @@ def deserialise_likelihood_function(data):
...
+        lf.set_alignment(loci)
+        mprobs = [motif_probs[k] for k in motif_probs]
+        with lf.updates_postponed():
```
GavinHuttley, line 278: move this `with` statement to after the if/else block

src/cogent3/util/deserialise.py
```diff
@@ -262,17 +262,28 @@ def deserialise_likelihood_function(data):
...
+            for motif_probs in mprobs:
+                lf.set_motif_probs(motif_probs)
+    else:
+        aln = deserialise_seq_collections(data.pop("alignment"))
+        lf.set_alignment(aln)
```
GavinHuttley, line 283: move this `lf.set_alignment(aln)` to after the if/else block, but before the `with` clause from above and add
```python
mprobs = [motif_probs]
```
inside the `else` block.

GavinHuttley (changes requested):
simplify the code in `deserialise_likelihood_function`

PR #817, opened 2020-11-25
Removing discontinued class ConvertFields

### LINKED ISSUE #811: remove code earmarked deprecated for 2020.11
several functions with `deprecated()` call now need to be removed in prep to for release

aberki1234:
Removes this class from table.py, and also removes an instance where it is used in psl.py. Fixes #811


src/cogent3/parse/table.py
```diff
@@ -21,47 +21,6 @@
...
-
-    def convert_by_line(self, line):
-        """converts each column in a line"""
-        return self.conversion(line)
-
-    def __call__(self, *args, **kwargs):
-        return self._func(*args, **kwargs)
-
-
 def SeparatorFormatParser(
```
GavinHuttley, line 24: there is a use of `deprecated()` for 2020.11 in this file that also needs to be removed.

GavinHuttley (changes requested):
just one additional removal, see my comments

also, make sure the next commit message includes "fixes #<issue number>"

src/cogent3/parse/table.py
GavinHuttley: My mistake, I didn't check the full file path ... doh!

Do a project-wide search for "deprecated(" (that will identify all uses). You'll find another one specifying 2020.11 in `cogent3/util/table.py`, that also needs to be removed.

PR #913, opened 2021-05-10
Maintenance

GavinHuttley:


setup.py
```diff
@@ -59,7 +60,7 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     platforms=["any"],
-    license=["BSD"],
+    license=__license__,
```
merwok, line 63: FWIW, license is a free-form field that’s not useful when you’re using a standard license that has a classifier.

Issue #941, opened 2021-09-09
JC69 model name is wrong


GavinHuttley:
indicates it's F81

```python
from cogent3 import get_model

sm = get_model("JC69")
sm.name
```
returns F81!

GavinHuttley:
Note, this affects `K80` too

jamesmartini:
Models JC69 and K80 are nested based on F81 and HKY85, respectively. Suggested fix made in pull request: change API so that the name argument is now in kwargs variable for the nested models, and make sure base models retain their name.

GavinHuttley:
fixed in d87ccbd54ddca5bc7bbd5feddd064e4d01c7f597

PR #947, opened 2021-09-21
API: Nested models now reflect their appropriate name. #941

jamesmartini:
Nested models now reflect their appropriate name, while base models retain their base name.

### DIFF
--- a/doc/doctest_rsts.py
+++ b/doc/doctest_rsts.py
@@ -13,6 +13,7 @@ from nbconvert.preprocessors import CellExecutionError, ExecutePreprocessor
 
 from cogent3.util.misc import atomic_write
 
+
 __author__ = "Gavin Huttley"
 __copyright__ = "Copyright 2007-2021, The Cogent Project"
 __credits__ = ["Gavin Huttley"]
--- a/doc/examples/demo-mpi-parallel.py
+++ b/doc/examples/demo-mpi-parallel.py
@@ -4,6 +4,7 @@ import time
 
 from cogent3.util import parallel
 
+
 # the following environment variable is created by PBS on job execution
 PBS_NCPUS = os.environ.get("PBS_NCPUS", None)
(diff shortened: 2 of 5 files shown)


GavinHuttley (changes requested):
Thanks heaps for this. Can you add a new test method to the `CannedModelsTest` class in`tests/test_evolve/test_models.py`, e.g.

```python
def test_model_names(self):
    """name attribute matches model name"""
    for model_name in models:
        model = get_model(model_name)
        self.assertEqual(model.name, model_name)    
```
There may be other models that fail this!

jamesmartini:
I had a problem with that test: The discrete time substitution model 'DT' substracts the motif_length from its name 'DT'. Two proposed solutions: 
1. Create a separate test for discrete models.
2. Or create an 'effective_name' argument in the kw variable for the DT model which tracks DT-motif_length and leaves 'name' as 'DT'.

GavinHuttley:
There's a third option, if `model_name` starts with "DT", only check that `model.name` starts with "DT"

jamesmartini:
Implementing an explicit check for "DT" in model.name reveals there is also another model AH96 which has name "AH96_mtREV24". The following proposed test checks that if model X has a name different to X then it checks that the name satisfies XZ for any string Z, otherwise checks equality:

```
    def test_model_names(self):
        """name attribute matches model name"""
        for model_name in models:
            model = get_model(model_name)
            if model.name is not model_name:
                self.assertTrue(model.name.startswith(model_name))
            else:
                self.assertEqual(model.name, model_name)
```

GavinHuttley:
Thanks heaps James! I played around a bit trying to simplify things and wound back pretty much where you were. Whoops!

jamesmartini:
> Thanks heaps James! I played around a bit trying to simplify things and wound back pretty much where you were. Whoops!

I really appreciate the detail and feedback. Cheers Gavin!

PR #1003, opened 2022-01-15
BUG: fixed name bug when added same name

bensilke-amd:


tests/test_core/test_sequence.py
```diff
@@ -2,6 +2,7 @@
 """Unit tests for Sequence class and its subclasses.
 """
 
+from cgi import test
```
GavinHuttley, line 5: remove this import, it's not required

src/cogent3/core/sequence.py
```diff
@@ -1024,6 +1024,10 @@ def __add__(self, other):
         else:
             other_seq = other
         new_seq = self.__class__(self._seq + other_seq)
```
GavinHuttley, line 1026: This is python 3.8+ syntax, so the tests will fail on python 3.7.
Also, `self._seq` is a string so will not have a `.name` attribute.

Replace the code with the following if expression and then create the instance
```python
name = self.name if self.name == other.name else None
new_seq = self.__class__(self._seq + other_seq, name=name)
```

GavinHuttley (changes requested):
Thanks heaps, but (of course) there's a couple of gotcha's.
First, the CI runs against Python versions 3.7, 3.8 and 3.9 (soon to add 3.10 to this matrix too). That means the most advanced language features we can use are those in python 3.7.
Second, your tests are nice and explicit. It occurs to me that, you should also check the string value of the sequence is correct from the concatenated sequences. For example,
```python
self.assertEqual(str(added_duplicates), even + even) # etc...
```

GavinHuttley (approved):

PR #1065, opened 2022-06-27
Add open_url function in util.io and its test in test_util.io

Nick-Foto:


tests/test_util/test_io.py
```diff
@@ -302,6 +304,150 @@ def test_open_zip_multi(self):
             with self.assertRaises(ValueError):
                 open_(zip_path)
 
+    def test_open_url(self):
```
GavinHuttley, line 307: There is a pattern in these tests that can be exploited to reduce the number of lines. Given the root for remote and local files and the different file modes you could simplify these tests, e.g.
```python
local_root = pathlib.Path("data")
remote_root = pathlib.Path("raw.githubusercontent.com/cogent3/cogent3/develop/tests/data")
for file_name in ("c_elegans_WS199_dna_shortened.fasta", "ff2_test.gff"):  # extend for any others
    for mode in ("r", "rb", "rt"):  # extend for any others
        with open_(local_root / file_name, mode=mode) as infile:
            local_data  = infile.read()
        with open_url(f"https://{remote_root / file_name}", mode=mode) as infile:
            remote_data = infile.read()
        # differences in line-endings between platforms mean the binary data may be different
        # You can handle this by splitting the lines so those different characters are removed, thus...
        self.assertEqual(remote_data.splitlines(), local_data.splitlines()) # should work
```

GavinHuttley (changes requested):
Thanks for the contribution! The tests are failing due to binary mode revealing platform-specific line endings. Easily addressed (see my suggested patch in the comment) along with the simplification of the tests.

tests/test_util/test_io.py
Nick-Foto: I pushed new changes. Please review and merge these changes.

Nick-Foto: About new problem in the CI, I installed Python3.7 in my laptop and checked. It works in my laptop with Python 3.7.13.  
After searching about it, I think it is a bug in Python 3.7 [socket module calls with long host names can fail with idna codec error.](https://bugs.python.org/issue32958)

GavinHuttley: I've reproduced the error and can confirm it is due to using `pathlib.Path()`. If you [look at the traceback](https://github.com/cogent3/cogent3/runs/7086100449?check_suite_focus=true#step:5:139) you'll see the url passed to Python's code is `'raw.githubusercontent.com\\cogent3\[etc...]`.

This is an artefact of `str(pathlib.Path())`. On windows this will convert `raw.githubusercontent.com/cogent3` to what is in the above. So I suggest change the `remote_root` (leave `local_root` unchanged) to a template string.

```python
remote_root = "https://raw.githubusercontent.com/cogent3/cogent3/develop/tests/data/{}" 
```
And use it as
```python
with open_url(remote_root.format(file_name)) as infile:
    remote_data = infile.read()
```

GavinHuttley:
You can ignore the issues identified by lgtm-com since that's a bug (which I have reported) on their end.

tests/test_util/test_io.py
Nick-Foto: I pushed new changes. Please review and merge these changes.

Issue #1082, opened 2022-07-06
Move checkpointing code from Composable onto _checkpointable class


GavinHuttley:
All attributes, methods related to checkpointing should be moved onto the writer classes (which inherit from `_checkpointable`.

Specifically, the methods `Composable.job_done()` and the attribute `Composable.checkpointable` should be shifted onto `_checkpointable`. The redundant attribute `self._checkpointable` should be removed.

GavinHuttley:
We have eliminated support for checkpointable apps.

PR #1090, opened 2022-07-09
DEV remove test for _type attribute

Nick-Foto:
@GavinHuttley 
I removed all cases of `_type` in tests. But this commit is **not** ready to merge.
Please look at this function. It cannot find `loader` when we don't have `_type` attribute. Please help me how to solve this.

```
def _set_checkpoint_loader(self):
        loader = {"sequences": load_unaligned}.get(self._out._type, load_aligned)
        loader = loader(format=self._format)
        self._load_checkpoint = loader
```

### DIFF
--- a/src/cogent3/app/io.py
+++ b/src/cogent3/app/io.py
@@ -524,8 +524,6 @@ class load_json(Composable):
     """Loads json serialised cogent3 objects from a json file.
     Returns whatever object type was stored."""
 
-    _type = "output"
-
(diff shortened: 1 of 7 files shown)


GavinHuttley:
Addressing `_set_checkpoint_loader` will be done by completing #1082.

Since this is a separate branch, I'm OK with tests being broken in a small way. But we do want to fix as many test failures resulting from an invidual patch as possible.

PR #1096, opened 2022-07-11
Composable refactor

Nick-Foto:
Change all app subclasses to inherit from the parent Composable class


GavinHuttley:
@Nick-Foto there are conflicts in `test_composable.py`

PR #1108, opened 2022-08-09
Composable refactor

Nick-Foto:


GavinHuttley:
There's a logic error in `composable._add()`. Delete the following lines
```python
    if self.app_type is not LOADER:
        raise TypeError("Left hand side of add operator must be of type loader")
```
It is not necessary for left hand most app to always be a loader.

**IFF** a loader is present in the group of apps to be composed, it must be left-most.

Apps can be composed without a loader.

Same logic applies to writers, i.e. **IFF** a writer is present in the group of apps to be composed, it must be right-most.

GavinHuttley (changes requested):
There's a logic error in `composable._add()` that caused the test failure. Delete the following lines
```python
    if self.app_type is not LOADER:
        raise TypeError("Left hand side of add operator must be of type loader")
```
**IFF** a loader is present in the group of apps to be composed, it must be left-most. Apps can be composed without a loader.

Same logic applies to writers, i.e. **IFF** a writer is present in the group of apps to be composed, it must be right-most.

I also noticed the error messages in this function can be improved. These types of error messages
```python
    if not self._return_types:
        raise TypeError(
            "Left hand side of add operator must have non-empty return types"
        )
```
need to be more informative so a dev/user can track down the problem. Change them to something like
```python
    if not self._return_types:
        raise TypeError(
            f"return type not defined for {self.__class__.__name__!r}"
        )
```

PR #1109, opened 2022-08-10
Composable refactor

Nick-Foto:


GavinHuttley (changes requested):
Revert ALL changes to `composable.Composable.apply_to()`

src/cogent3/app/align.py
```diff
@@ -331,10 +333,6 @@ class align_to_ref(Composable):
     ArrayAlignment.
     """
 
-    _input_types = SEQUENCE_TYPE
```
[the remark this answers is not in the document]
Nick-Foto: I was checking only `_input_types`. I found ALIGNED_TYPE corresponds to AlignedSeqsType, but I used the same type for input and output of the main function.

src/cogent3/app/evo.py
```diff
@@ -237,13 +221,12 @@ def _fit_aln(
 
         return lf
 
-    def main(self, aln, initialise=None, construct=True, **opt_args):
+    T = Union[SerialisableType, AlignedSeqsType]
+
+    def main(self, aln: T, initialise=None, construct=True, **opt_args) -> T:
```
GavinHuttley, line 226: I've been thinking about this a bit more. For every cogent3 decorated app EXCEPT writers, the input argument should not include `SerialisableType`. The reason being, they really do require a more specific type for execution. Whether an object can be serialised only matters for writers.

In this specific instance, change to
```python
def main(self, aln: AlignedSeqsType, initialise=None, construct=True, **opt_args) -> Union[SerialisableType, AlignedSeqsType]:
```

GavinHuttley (changes requested):
An important change. Please see specific example and apply to all apps you've decorated.

src/cogent3/app/evo.py
Nick-Foto: Do I need to remove `SerialisableType` from input of all main functions? and add a condition in decorator to check it?

src/cogent3/app/io.py
```diff
@@ -512,19 +456,15 @@ def main(self, data, identifier=None):
...
-    _output_types = SERIALISABLE_TYPE
-
     def __init__(self):
-        super(load_json, self).__init__(
-            input_types=self._input_types, output_types=self._output_types
-        )
+        pass
 
-    def main(self, path):
+    def main(self, path: SerialisableType) -> SerialisableType:
```
GavinHuttley, line 467: should be `path: IdentifierType`

src/cogent3/app/io.py
```diff
@@ -429,17 +388,13 @@ def main(self, data, identifier=None):
...
 
         output = data.to_string(format=self._format)
-        stored = self.data_store.write(identifier, output)
-        return stored
+        return self.data_store.write(identifier, output)
 
 
-class write_seqs(_checkpointable):
+@composable
+class write_seqs:
```
GavinHuttley, line 395: I'm not sure we should be converting these yet

src/cogent3/app/tree.py
```diff
@@ -127,15 +116,11 @@ def __init__(self, drop_invalid=False):
...
-            output_types=self._output_types,
-            data_types=self._data_types,
-        )
-        self._formatted_params()
         self._drop_invalid = drop_invalid
 
-    def main(self, dists):
+    T = Union[SerialisableType, AlignedSeqsType]
+
+    def main(self, dists: T) -> Union[SerialisableType, TreeType]:
```
GavinHuttley, line 123: change to `dists: PairwiseDistanceType`

GavinHuttley (changes requested):
Revert changes to io writer apps. Their init constructor call uses `super` and is passing much more than just the input, output and data types variables. Revert the xfails also.

We will need to deal with writers as a separate step.

src/cogent3/app/tree.py
Nick-Foto: Fist I had written `PairwiseDistanceType` for this. But for a test error, I changed it to `Union[SerialisableType, AlignedSeqsType]`.

GavinHuttley: That seems problematic. In that case, better to leave that app un decorated so we can tackle it separately.

GavinHuttley: I'd rather see such errors than just mark as `xfail`.

Nick-Foto:
> 

I looked at all classes. None of them has `supe().__init__` call passing more variables. Please help me to find writer classes.

PR #1112, opened 2022-08-13
ENH: add app registry to composable decorator

GavinHuttley:
[NEW] module level dictionary that records apps during decoration
    by their object provenance. This means a 3rd party developer
    could develop an app with the same class name, it would be
    distinguished by it's full provenance.

### DIFF
--- a/src/cogent3/app/composable.py
+++ b/src/cogent3/app/composable.py
@@ -951,6 +951,9 @@ class _connected:
         self.storage[id(instance)] = value
 
 
+__app_registry = {}
+
+
 def composable(klass=None, *, app_type: AppType = GENERIC):
     app_type = AppType(app_type)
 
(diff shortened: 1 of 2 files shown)


src/cogent3/app/composable.py
```diff
@@ -1019,6 +1022,8 @@ def wrapped(klass):
         if hasattr(klass, "__slots__"):
             klass.__slots__ += tuple(slot_attrs)
 
+        # register this app
+        __app_registry[get_object_provenance(klass)] = klass.__name__
```
Nick-Foto, line 1026: I would write this as:
__app_registry[klass.__name__] = get_object_provenance(klass)

Nick-Foto: I understand this logic, but why a user may want to have two classes with the same name.

GavinHuttley: I think the issue is more ensuring registration of the built-in apps is not overridden by another developer. It also means it's easier to check whether an object is composable by looking at full provenance.

PR #1115, opened 2022-08-15
Composable refactor

GavinHuttley:
address app inheritance and fix return type hints in evo

### DIFF
--- a/src/cogent3/app/evo.py
+++ b/src/cogent3/app/evo.py
@@ -20,6 +20,7 @@ from .typing import (
     AlignedSeqsType,
     BootstrapResultType,
     HypothesisResultType,
+    ModelCollectionResultType,
     ModelResultType,
     SerialisableType,
     TabularResultType,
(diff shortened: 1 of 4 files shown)


src/cogent3/app/evo.py
```diff
@@ -292,10 +294,8 @@ def __call__(self, other, *args, **kwargs):
         return other
 
 
-@define_app
-class model_collection:
-    """Fits a collection of models. Returns a
-    model_collection_result."""
+class _ModelCollectionBase:
```
Nick-Foto, line 297: Do we need to give this base class to users of composable decorator?

GavinHuttley: No, it has only those 2 uses.

Nick-Foto:
@GavinHuttley 
In _apply_to a lot of xfail tests fail for `ui` is None. Do you want to add a checking for this?

PR #1119, opened 2022-08-16
BUG: fix __repr__ of composed functions

GavinHuttley:


tests/test_app/test_tree.py
```diff
@@ -85,9 +84,6 @@ def test_composable_apps(self):
         )
         self.assertIsInstance(proc, tree_app.quick_tree)
         self.assertIsInstance(proc.input, dist.fast_slow_dist)
```
Nick-Foto, line 86: I would prefer using a different name for instance of `fast_slow_dist` class. It is the same as class name.

GavinHuttley: Changed to `calc_dist`

PR #1120, opened 2022-08-16
TST: updated test to reflect fundamental change to str representation of apps

GavinHuttley:
[CHANGED] The interception of constructor arguments using __new__ means we no
    longer have a per class customised representation. This required some tweaking of
    test to accomodate changed behaviour of the evo.hypothesis app.


Nick-Foto:
@GavinHuttley 
I looked at changes and couldn't find anything to comment about it.

PR #1121, opened 2022-08-17
Composable refactor

GavinHuttley:
Nearly finalised porting writers to new decorator approach

### DIFF
--- a/src/cogent3/app/io.py
+++ b/src/cogent3/app/io.py
@@ -20,7 +20,13 @@ from cogent3.parse.sequence import PARSERS
 from cogent3.util.deserialise import deserialise_object
 from cogent3.util.table import Table
 
-from .composable import LOADER, WRITER, NotCompleted, _checkpointable, define_app
+from .composable import (
+    LOADER,
+    WRITER,
+    NotCompleted,
+    _checkpointable,
+    define_app,
+)
 from .data_store import (
     IGNORE,
     OVERWRITE,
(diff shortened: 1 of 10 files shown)


src/cogent3/app/io.py
```diff
@@ -351,7 +358,7 @@ def main(self, path: IdentifierType) -> TabularType:
 
 
 @define_app(app_type=WRITER)
-class write_tabular:
+class write_tabular(_checkpointable):
```
Nick-Foto, line 361: Here we mixed decorator approach and inheritance. Do we expect other users to use `_checkpointable` by inheritance or will we change it in the future?

tests/test_app/test_dist.py
```diff
@@ -213,7 +216,6 @@ def test_composes_with_write_tabular(self):
             dist_calc = dist_app.fast_slow_dist(distance="hamming", moltype="protein")
             _ = dist_calc + writer
```
Nick-Foto, line 217: Do we always need have an assignment for result of composable apps? or 
Can we write this expression without assignment like this: `app1+app2` without assigning result to a variable? I imagine `app1+app2` like calling two functions: `function1(function2())`

src/cogent3/app/io.py
GavinHuttley: change in future

tests/test_app/test_dist.py
GavinHuttley: I prefer to do the assignment but it's possible to do what you've said

PR #1124, opened 2022-08-18
Composable refactor

GavinHuttley:
Fixes to `available_apps()`

### DIFF
--- a/src/cogent3/app/__init__.py
+++ b/src/cogent3/app/__init__.py
@@ -50,6 +50,6 @@ def available_apps():
     for name in __all__:
         importlib.import_module(f"cogent3.app.{name}")
 
-    rows = [_get_app_attr(app, __app_registry.get(app)) for app in __app_registry]
+    rows = [_get_app_attr(app, is_comp) for app, is_comp in __app_registry.items()]
     header = ["module", "name", "composable", "doc", "input type", "output type"]
     return Table(header=header, data=rows)
--- a/src/cogent3/app/__init__.py
+++ b/src/cogent3/app/__init__.py
@@ -33,24 +33,23 @@ def _get_app_attr(name, is_composable):
         name,
         is_composable,
         obj.__doc__,
-        _types["_data_types"],
-        _types["_return_types"],
+        ", ".join(sorted(_types["_data_types"])),
+        ", ".join(sorted(_types["_return_types"])),
     ]
 
 
 def available_apps():
     """returns Table listing the available apps"""
+    import cogent3.app as apps
+
     from cogent3.util.table import Table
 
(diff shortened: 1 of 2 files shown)


src/cogent3/app/__init__.py
```diff
@@ -33,24 +33,23 @@ def _get_app_attr(name, is_composable):
...
 
-    mod = importlib.import_module(f"{__name__}.composable")
-    rows.append(
-        _get_app_attr(
-            f"{mod.__name__}.user_function", issubclass(user_function, Composable)
-        )
-    )
-    header = ["module", "name", "composable", "doc", "outputs", "data type"]
-    return Table(header, rows)
+    rows = [_get_app_attr(app, __app_registry.get(app)) for app in __app_registry]
```
Nick-Foto, line 53: In `__app_registry`, keys are `get_object_provenance(klass)`. But here we use `app` as key. Are these equal?

src/cogent3/app/__init__.py
```diff
@@ -33,24 +33,23 @@ def _get_app_attr(name, is_composable):
         name,
         is_composable,
         obj.__doc__,
-        _types["_data_types"],
-        _types["_return_types"],
+        ", ".join(sorted(_types["_data_types"])),
+        ", ".join(sorted(_types["_return_types"])),
```
Nick-Foto, line 37: `sorted` changes order, but we may need order later.

Nick-Foto: We may need order to find type of the first argument of `main` function.

src/cogent3/app/__init__.py
GavinHuttley: I'm iterating over the keys in the dict, so yes.

src/cogent3/app/__init__.py
GavinHuttley: Not in the table. Note, the original app attributes are `frozenset()` (thus immutable).

These lists are created only to produce this table.

src/cogent3/app/__init__.py
GavinHuttley: Actually, there's no need to do the `__app_registry.get(app)`. The value must exist! Fixing that now.

src/cogent3/app/__init__.py
Nick-Foto: OK, Now I understand it is only for this table.

src/cogent3/app/__init__.py
Nick-Foto: We have used `get` function for dict a lot of times, so we forget to use normal usage of key indexing.

PR #1125, opened 2022-08-22
BUG: fix bug in _new

GavinHuttley:
[FIXED] was not correctly handling variable positional arguments.
    Use the inspect Signature.bind_partial method to solve this correctly!

### DIFF
--- a/tests/test_app/test_composable.py
+++ b/tests/test_app/test_composable.py
@@ -531,5 +531,41 @@ def test_composed_func_pickleable():
     assert unpickled.input is not None
 
 
+def test_composable_new1():
+    """correctly associate argument vals with their names when have variable
+    positional args"""
+
+    @define_app
+    class pos_var_pos:
+        def __init__(self, a, b, *args):
+            self.a = a
+            self.b = b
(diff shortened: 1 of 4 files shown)


tests/test_app/test_composable.py
```diff
@@ -531,5 +531,41 @@ def test_composed_func_pickleable():
...
+            self.a = a
+            self.b = b
+            self.args = args
+
+        def main(self, val: int) -> int:
+            return val
+
+    instance = pos_var_pos(2, 3, 4, 5, 6)
+    assert instance._init_vals == {"a": 2, "b": 3, "args": (4, 5, 6)}
+
```
Nick-Foto, line 550: I think we should add other test with __slots__.

GavinHuttley: Good point. I tried and it doesn't work.

How we're handling slots is incorrect and needs more thought. Beyond this patch.

Nick-Foto: We can forget slots for now.  We don't need to cover everything in the first version of new composable decorator.

PR #1133, opened 2022-08-25
BUG: fix progressive alignment zero edge lengths

xingjianleng:


src/cogent3/align/progressive.py
```diff
@@ -90,6 +90,10 @@ def TreeAlign(
         dists = dcalc.get_pairwise_distances().to_dict()
         tree = NJ.nj(dists)
 
+    for edge in tree.postorder():
+        if edge.name != "root":
```
Nick-Foto, line 94: You may need converting to lower case before comparing strings.

GavinHuttley: Not for this case, the "root" name is created by the instance, not the user.

Nick-Foto: OK. Thank you.

PR #1134, opened 2022-08-25
ENH: show TypeError when inheriting from a decorated class or repeate…

Nick-Foto:
…d decorating


Nick-Foto:
@GavinHuttley 
Please review and merge these changes. If there is a merge conflict after merging your recent PR, I will fix it.
I don't know why I cannot select reviewer in the right hand side of the window.

PR #1135, opened 2022-08-25
ENH: extend define_app supports decorating functions

GavinHuttley:
[NEW] converts function to a class with the same name as the
    function, then does standard decoration process. Intended to
    replace the user_function class / appify decorator pair.

### DIFF
--- a/src/cogent3/app/composable.py
+++ b/src/cogent3/app/composable.py
@@ -979,7 +979,7 @@ def _class_from_func(func):
     module = func.__module__  # to be assigned to the generated class
     sig = inspect.signature(func)
     class_name = func.__name__
-    _main = _set_hints(_main, *_get_raw_hints(func, 2))
+    _main = _set_hints(_main, *_get_raw_hints(func, 1))
 
     _class_dict = {"__init__": _init, "main": _main, "_user_func": staticmethod(func)}
 
--- a/src/cogent3/app/composable.py
+++ b/src/cogent3/app/composable.py
@@ -1,10 +1,8 @@
-import contextlib
 import inspect
 import json
 import os
 import pathlib
 import re
-import sys
 import textwrap
 import time
 import traceback
--- a/src/cogent3/app/composable.py
+++ b/src/cogent3/app/composable.py
@@ -1,11 +1,14 @@
+import contextlib
 import inspect
 import json
 import os
 import pathlib
 import re
+import sys
 import textwrap
 import time
 import traceback
+import types
 
 from copy import deepcopy
 from enum import Enum
@@ -784,6 +787,29 @@ WRITER = AppType.WRITER
 GENERIC = AppType.GENERIC
 
 
+def _get_raw_hints(main_func, min_params):
+    _no_value = inspect.Parameter.empty
+    params = inspect.signature(main_func)
+    if len(params.parameters) < min_params:
+        raise ValueError(
+            f"{main_func.__name__!r} must have at least {min_params} input parameters"
+        )
+    # annotation for first parameter other than self, params.parameters is an orderedDict
+    first_param_type = [p.annotation for p in params.parameters.values()][1]
+    return_type = params.return_annotation
+    if return_type is _no_value:
+        raise TypeError("must specify type hint for return type")
+    if first_param_type is _no_value:
+        raise TypeError("must specify type hint for first parameter")
+
+    if first_param_type is None:
+        raise TypeError("NoneType invalid type for first parameter")
+    if return_type is None:
+        raise TypeError("NoneType invalid type for return value")
+
+    return first_param_type, return_type
+
+
 def _get_main_hints(klass) -> Tuple[set, set]:
     """return type hints for main method
     Returns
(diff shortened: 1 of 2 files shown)


tests/test_app/test_composable.py
```diff
@@ -634,5 +633,34 @@ def main(self, val: int) -> int:
...
+def func2app(arg1: int, exponent: int) -> float:
+    return arg1 ** exponent
+
+
+def test_decorate_app_function():
+    """works on functions now"""
+    import inspect
+
+    sqd = func2app(exponent=2)
+    assert sqd(3) == 9
```
Nick-Foto, line 647: What is role of the `main` function here?

src/cogent3/app/composable.py
```diff
@@ -784,6 +787,29 @@ class AppType(Enum):
...
 
+def _get_raw_hints(main_func, min_params):
+    _no_value = inspect.Parameter.empty
+    params = inspect.signature(main_func)
+    if len(params.parameters) < min_params:
+        raise ValueError(
+            f"{main_func.__name__!r} must have at least {min_params} input parameters"
+        )
+    # annotation for first parameter other than self, params.parameters is an orderedDict
+    first_param_type = [p.annotation for p in params.parameters.values()][1]
```
Nick-Foto, line 798: If we can have `min_params > 1`, then I think this index should be: 
`[1:min_params]`
and return frozenset of hints

src/cogent3/app/composable.py
```diff
@@ -937,6 +955,46 @@ def _validate_data_type(self, data):
...
+    # these methods MUST be in function scope so that separate instances are
+    # created for each decorated function
+    def _init(self, *args, **kwargs):
+        self._args = args
+        self._kwargs = kwargs
+
+    def _main(self, arg, **kwargs):
+        kw_args = deepcopy(self._kwargs)
+        kw_args = {**kw_args, **kwargs}
+        args = (arg,) + deepcopy(self._args)
```
Nick-Foto, line 977: If we have more than one argument, we need `(*args,)`

tests/test_app/test_composable.py
```diff
@@ -634,5 +633,34 @@ def main(self, val: int) -> int:
...
+    import pickle
+
+    sqd = func2app(exponent=2)
+    u = pickle.loads(pickle.dumps(sqd))
+    assert u(4) == 16
+
+    p = get_object_provenance(func2app)
+    __app_registry.pop(p, None)
+
+
```
Nick-Foto, line 664: After you merged this, you can make a new issue to add more tests. I will do it.

Nick-Foto:
> [NEW] converts function to a class with the same name as the function, then does standard decoration process. Intended to replace the user_function class / appify decorator pair.

After adding a new class with the same name as function, how can we be sure that calling that name calls the class not the function?
Do we need a mechanism to remove that function name from namespace and replace it with a new name to prevent mismatch in names?
1- convert myfunc   -> myfunc_newname
2- newclass named myfunc -> it calls myfunc_newname

Nick-Foto:
> This pull request **introduces 6 alerts** when merging [2b482c7](https://github.com/cogent3/cogent3/commit/2b482c791fae1b095dfbddbb4ba1223d82984fd5) into [08f3600](https://github.com/cogent3/cogent3/commit/08f3600e8f57eb71d6c918b3aec1d6f8efd7901b) - [view on LGTM.com](https://lgtm.com/projects/g/cogent3/cogent3/rev/pr-b8e400bfb138cf5c2d220be988bd83a843ca7f89)
> 
> **new alerts:**
> 
> * 2 for Wrong name for an argument in a call
> * 2 for Unused import
> * 2 for Wrong number of arguments in a call

These alerts from LGTM are very interesting. Why those functions are calling `wrapped` function in the `define_app` decorator?

tests/test_app/test_composable.py
GavinHuttley: `main()` method (invoked within `__call__()`) is assembling the arguments for the call to the bound static method (the original function being decorated)

src/cogent3/app/composable.py
GavinHuttley: I disagree. This first provided type hint is used to dictate compatibility for composition because it's the primary input.

For the decorated class case, there must be 2 parameters: `main(self, data)`. But for the decorated function I'm now thinking we should just allow it to be one.

src/cogent3/app/composable.py
GavinHuttley: I am not sure about this, mainly because at present I have not made a decision about what the call api should support. This needs to be revisited.

GavinHuttley:
> After adding a new class with the same name as function, how can we be sure that calling that name calls the class not the function? Do we need a mechanism to remove that function name from namespace and replace it with a new name to prevent mismatch in names? 1- convert myfunc -> myfunc_newname 2- newclass named myfunc -> it calls myfunc_newname

The short answer is because this decorator is returning a different object. The original function is bound to this new object, but is never returned itself.

> These alerts from LGTM are very interesting. Why those functions are calling wrapped function in the define_app decorator?

These are false positives being reported by LGTM. Their algorithm has not correctly resolved that those objects are the result of a called to `wrapped` and not `wrapped` itself. They will need to be reported.

PR #1136, opened 2022-08-26
TST: Change composable tests to test apps with new decorator

Nick-Foto:


src/cogent3/app/composable.py
```diff
@@ -1011,8 +1011,8 @@ def define_app(klass=None, *, app_type: AppType = GENERIC, composable: bool = Tr
 
     app_type = AppType(app_type)
 
-    if inspect.isfunction(klass):
-        klass = _class_from_func(klass)
+#    if inspect.isfunction(klass):
```
GavinHuttley, line 1014: this should not be commented out

tests/test_app/test_composable.py
```diff
@@ -385,23 +385,22 @@ def slicer(val, index=2):
...
     from cogent3 import make_aligned_seqs
 
-    u_function = define_app(foo)
-    u_instance = u_function()
+    u_function = define_app(foo)()
 
     aln = make_aligned_seqs(data=[("a", "GCAAGCGTTTAT"), ("b", "GCTTTTGTCAAT")])
-    got = u_instance(aln)
+    got = u_function(aln)
+    print("got = ",got)
```
GavinHuttley, line 403: delete print statements

src/cogent3/app/composable.py
```diff
@@ -1054,6 +1065,50 @@ def wrapped(klass):
     return wrapped(klass) if klass else wrapped
 
 
+def remove_app(klass):
```
GavinHuttley, line 1068: delete this function

src/cogent3/app/composable.py
Nick-Foto: I had commented it. I don't know why you moved this part of code into `wrapped` function.

GavinHuttley: meant to be duplicated, not moved

GavinHuttley:
```python
def decorate_app(klass=None, /, ...):
    if klass.is function(convert to a class)
    def wrap(klass):
         if klass.is function(convert to a class)
```
the reason for both is the first one is the case where
```python
@decorate_app
def foo(): ...
```
the second one is
```python
@decorate_app(composable=False) # for example
def foo(): ...
```

tests/test_app/test_composable.py
Nick-Foto: done

src/cogent3/app/composable.py
Nick-Foto: done

src/cogent3/app/composable.py
Nick-Foto: I think both above cases run the first `if klass.isfunction` code.
But the first one is called when defining and decorating the class and the second one is called when creating an instance of class.

tests/test_app/test_composable.py
```diff
@@ -39,294 +42,383 @@
...
+            return val
+
+    class app_dummyclass_2:
+        def __init__(self, b):
+            self.b = b
+
+        def main(self, val: int) -> int:
+            return val
+
+    define_app(app_dummyclass_1)
```
GavinHuttley, line 62: why not actually decorate the classes?

tests/test_app/test_composable.py
```diff
@@ -39,294 +42,383 @@
...
+            return val
+
+    class app_dummyclass_3:
+        def __init__(self, c):
+            self.c = c
+
+        def main(self, val: int) -> int:
+            return val
+
+    define_app(app_dummyclass_1)
```
GavinHuttley, line 98: as above, why not actually add the decorator to the class definition?

tests/test_app/test_composable.py
```diff
@@ -39,294 +42,383 @@
...
+    """this should raise a ValueError"""
+
+    class app_dummyclass_1:
+        def __init__(self, a):
+            self.a = a
+
+        def main(self, val: int) -> int:
+            return val
+
+    define_app(app_dummyclass_1)
```
GavinHuttley, line 124: as above

tests/test_app/test_composable.py
```diff
@@ -39,294 +42,383 @@
...
+            return val
+
+    class app_dummyclass_3:
+        def __init__(self, c):
+            self.c = c
+
+        def main(self, val: int) -> int:
+            return val
+
+    define_app(app_dummyclass_1)
```
GavinHuttley, line 156: as above

tests/test_app/test_composable.py
Nick-Foto: Sorry, it remains from previous format that I had defined classes outside test functions. Now I will fix them.

Nick-Foto: I see you merged it. Next time I will change it.

PR #1137, opened 2022-08-29
Composable refactor

GavinHuttley:
Tweaks to testing and identifying some tests currently incorrect


tests/test_app/test_composable.py
```diff
@@ -74,31 +74,30 @@ def main(self, val: int) -> int:
 def test_composables_once():
     """composables can only be used in a single composition"""
 
+    @define_app
```
Nick-Foto, line 77: Thank you. I forgot to fix these.

PR #1141, opened 2022-08-29
Tidy composition checks

GavinHuttley:


src/cogent3/app/composable.py
```diff
@@ -853,18 +853,18 @@ def _disconnect(self):
...
     # Check order
-    elif isinstance(self, user_function) or isinstance(other, user_function):
-        pass
-    elif self.app_type is WRITER:
+    if self.app_type is WRITER:
         raise TypeError("Left hand side of add operator must not be of type writer")
     elif other.app_type is LOADER:
         raise TypeError("Right hand side of add operator must not be of type loader")
+
     if self._return_types & {"SerialisableType", "IdentifierType"}:
```
Nick-Foto, line 868: Can we write all of these as `elif` to be faster?

src/cogent3/app/composable.py
```diff
@@ -959,6 +959,9 @@ def _validate_data_type(self, data):
     }:
         return True
 
+    if isinstance(data, (tuple, list)):
```
Nick-Foto, line 962: And this one as `elif`.

src/cogent3/app/composable.py
Nick-Foto: I know they are logically unrelated, but all of them raise an error, so chaining them is possible.

PR #1144, opened 2022-08-30
ENH: support decorating a function with just positonal arguments

GavinHuttley:


src/cogent3/app/composable.py
```diff
@@ -983,10 +983,10 @@ def _init(self, *args, **kwargs):
         self._args = args
         self._kwargs = kwargs
 
-    def _main(self, arg, **kwargs):
+    def _main(self, arg, *args, **kwargs):
         kw_args = deepcopy(self._kwargs)
         kw_args = {**kw_args, **kwargs}
-        args = (arg,) + deepcopy(self._args)
+        args = (arg,) + args + deepcopy(self._args)
```
Nick-Foto, line 989: What if we change this to:
```
def _main(self, *args, **kwargs):
        kw_args = deepcopy(self._kwargs)
        kw_args = {**kw_args, **kwargs}
        new_args = args + deepcopy(self._args)
        bound = self._func_sig.bind(*new_args, **kw_args)
        ```

Nick-Foto: I tested this. My code doesn't work.

GavinHuttley:
We need to add a type hint to the first argument.

PR #1145, opened 2022-08-30
DEV: Added non_composable to enum, removed composable argument from d…

Nick-Foto:
…efine_app, excluded _add method for non_composable apps

### DIFF
--- a/src/cogent3/app/composable.py
+++ b/src/cogent3/app/composable.py
@@ -1134,8 +1134,9 @@ def define_app(klass=None, *, app_type: AppType = GENERIC):
             "__repr__",
             "__str__",
             "__new__",
+            "_validate_data_type",
         ]
-        excludes = ["__add__", "disconnect", "input", "apply_to"]
+        excludes = ["__add__", "disconnect", "apply_to"]
         if app_type is not NON_COMPOSABLE and getattr(klass, "input", None):
             raise TypeError(
                 f"remove 'input' attribute in {klass.__name__!r}, this functionality provided by define_app"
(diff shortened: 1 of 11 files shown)


Nick-Foto:
@GavinHuttley 
Please review and merge these changes.
Please remind making a tag before merging this branch.

Nick-Foto:
@GavinHuttley 
I changed algorithm for checking existing functions in class. Items in `method_list` and `__mapping` were different, so some of functions were not checked. Because `__str__` doesn't exist in `__mapping`, the previous code didn't check it.  
I added complete tests for this part of code, also tests for `_add` and `min_params` in `_get_raw_hints`.

src/cogent3/app/composable.py
```diff
@@ -848,15 +850,17 @@ def _disconnect(self):
 
 
 def _add(self, other):
-    if other.input is not None:
+    if hasattr(other, "input") and other.input is not None:
```
GavinHuttley, line 853: Remove this `hasattr` check. Instead add a new conditional before this
```python
if getattr(other, `app_type`, None) not in {WRITER, LOADER, GENERIC}:
    raise TypeError(f"{other!r} is not composable"
```

src/cogent3/app/composable.py
```diff
@@ -848,15 +850,17 @@ def _disconnect(self):
...
         raise ValueError(
             f"{other.__class__.__name__} already part of composed function, use disconnect() to free them up"
         )
     if other is self:
         raise ValueError("cannot add an app to itself")
 
     # Check order
-    if self.app_type is WRITER:
+    if self.app_type is NON_COMPOSABLE or other.app_type is NON_COMPOSABLE:
```
GavinHuttley, line 861: delete this conditional,  addressed above.

GavinHuttley (changes requested):
I suggest defining a new variable within the `define_app` function to simplify the expressions
```python
composable = app_type is not NON_COMPOSABLE
```
then you can just revert all those statements to using the composable variable making them easier to read

src/cogent3/app/composable.py
Nick-Foto: Why for the first condition you used getattr, but for the next, directly .app_type?

PR #1150, opened 2022-08-31
Fixed bug in repr of composable apps

GavinHuttley:


src/cogent3/app/composable.py
```diff
@@ -904,12 +904,22 @@ def _repr(self):
...
+    if not args and not kwargs:
+        obj._init_vals = {}
+        return obj
+
+    if hasattr(klass, "_func_sig"):
+        # we have a decorated function, the first parameter in the signature
+        # is not given to constructor, so we create a new signature excluding that one
+        params = klass._func_sig.parameters
+        init_sig = inspect.Signature(parameters=list(params.values())[1:])
+        bargs = init_sig.bind_partial(*args, **kwargs)
```
Nick-Foto, line 916: If we have a user function, does this work? because user function does not have `self`?

tests/test_app/test_composable.py
```diff
@@ -800,6 +800,41 @@ def power(val: int, pow: int = 1) -> int:
...
+    for name, func in fns.items():
+        app = define_app(func)
+        if len(name.split("_")) == 1:
+            instance = app()
+            expect = f"{name}()"
+        elif name.endswith("kw"):
+            instance = app(**args["kw"])
+            expect = f"{name}(pow={args['kw']['pow']})"
+        else:
+            instance = app(args["pos"])
```
Nick-Foto, line 830: I cannot see a parameter named "pos".

src/cogent3/app/composable.py
GavinHuttley: the bound function cannot access self, but the methods on the containing class can

Nick-Foto: So, does the starting index need to be zero, for decorated user function?

PR #1151, opened 2022-09-02
ENH: support string literal type hints

GavinHuttley:
[NEW] allows foo(name: "some_type")

[CHANGED] add constraints for SeqType

### DIFF
--- a/tests/test_app/test_typing.py
+++ b/tests/test_app/test_typing.py
@@ -86,6 +86,13 @@ def test_hints_from_strings():
     assert got == [SerialisableType, AlignedSeqsType]
 
 
+def test_hints_resolved_from_str():
+    got = get_constraint_names("DnaSequence")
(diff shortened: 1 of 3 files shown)


tests/test_app/test_typing.py
```diff
@@ -86,6 +86,13 @@ def test_hints_from_strings():
     assert got == [SerialisableType, AlignedSeqsType]
 
 
+def test_hints_resolved_from_str():
+    got = get_constraint_names("DnaSequence")
+    assert got == {"DnaSequence"}
+    got = get_constraint_names(Union[SerialisableType, "DnaSequence"])
+    assert got == {"SerialisableType", "DnaSequence"}
+
```
Nick-Foto, line 94: Can we add all types here as parametrized test to see if all of them correctly resolved?
I can add this after you merge this PR.

PR #1155, opened 2022-09-13
update and tidy

GavinHuttley:


src/cogent3/core/alphabet.py
```diff
@@ -56,38 +57,21 @@ class AlphabetError(Exception):
...
     """
     if num_elements <= 256:
-        return uint8
+        dtype = uint8
     elif num_elements <= 2 ** 16:
-        return uint16
-    return uint32
+        dtype = uint16
+    elif num_elements <= 2 ** 32:
+        dtype = uint32
```
Nick-Foto, line 68: I think it should be
<= 255
<= 2**16 -1
<= 2**32 -1

or 

< 256
< 2**16
< 2**32

src/cogent3/core/alphabet.py
```diff
@@ -56,38 +57,21 @@ class AlphabetError(Exception):
...
+    elif num_elements <= 2 ** 64:
+        dtype = uint64
```
Nick-Foto, line 70: and the same for his line

tests/test_core/test_alphabet.py
```diff
@@ -73,6 +74,12 @@ def test_get_array_type(self):
         self.assertEqual(get_array_type(10000), uint16)
         self.assertEqual(get_array_type(65536), uint16)
         self.assertEqual(get_array_type(65537), uint32)
+        self.assertEqual(get_array_type(1 + 2 ** 32), uint64)
+
+    def test_get_array_type(self):
+        """get_array_type should return unsigned type that fits elements."""
+        with self.assertRaises(NotImplementedError):
+            self.assertEqual(get_array_type(1 + 2 ** 64), uint64)
```
Nick-Foto, line 83: And that int boundaries in these lines

src/cogent3/core/alphabet.py
GavinHuttley: Thanks for the review! You're absolutely correct!

Nick-Foto: @GavinHuttley 
I wrote 2 ** 16 , but markup changed it to **2**16.

PR #1160, opened 2022-09-29
ENH: apps now handle numpy arrays as returned values

GavinHuttley:


src/cogent3/app/composable.py
```diff
@@ -946,7 +949,8 @@ def _call(self, val, *args, **kwargs):
     except Exception:
         result = NotCompleted("ERROR", self, traceback.format_exc(), source=val)
 
-    if not result and not isinstance(result, NotCompleted):
+    # an empty numpy array may be valid
```
Nick-Foto, line 952: We have this comment, but there is no check for empty numpy array.

tests/test_app/test_composable.py
```diff
@@ -923,5 +923,20 @@ def main(self, val: int) -> int:
...
+
+    @define_app
+    def power(val: ndarray, pow: int) -> int:
+        return val ** pow
+
+    sqd = power(pow=2)
+    d = array([3, 3])
+    assert (sqd(d) == array([9, 9])).all()
+
+    __app_registry.pop(get_object_provenance(power), None)
```
Nick-Foto, line 938: I think we need tests for non-numpy arrays and other cases of error in _call function.

src/cogent3/app/composable.py
GavinHuttley: The statement
```python
    # an empty numpy array may be valid
    if not is_np_array and not result and not isinstance(result, NotCompleted):
       ...
```
This is addressed by excluding numpy arrays from checking if `not result`

tests/test_app/test_composable.py
GavinHuttley: Good point. I do wonder how general this should be. For example, if a function returns `0`, this will raise an exception. But should it? This makes make me think our definition of a null result is too broad.

Perhaps we should limit our error definition to the case where `result is None` and let everything else through?

Nick-Foto: Yes, as you said, when return value is an empty list or zero value, it should be ok.
What if val (input value) was an empty list?

PR #1162, opened 2022-10-05
DEV: Add new functions in DataStoreDirectory

Nick-Foto:


src/cogent3/app/data_store_new.py
```diff
@@ -272,50 +287,150 @@ def completed(self) -> list[DataMember]:
...
+                sources,
+            ]
+            rows.append(row)
+
+        reprlib.aRepr.maxstring = maxtring  # restoring original val
+        return Table(header=header, data=rows, title="not completed records")
 
     def __repr__(self):
         notcompleted_path = Path(self.source / _NOT_COMPLETED_TABLE)
         if not notcompleted_path.exists():
```
GavinHuttley, line 430: this condition should be deleted. There should also be a test with a directory that does not have a `not_completed` subdir but returns a correct `repr`

tests/test_app/test_data_store_new.py
```diff
@@ -192,6 +198,31 @@ def test_multi_write(fasta_dir, w_dstore):
...
+    nc3 = NotCompleted("FAIL", "dummy3", "dummy_message3", source="dummy_source3")
+    (nc_dir / _NOT_COMPLETED_TABLE / "nc1.json").write_text(nc1.to_json())
+    (nc_dir / _NOT_COMPLETED_TABLE / "nc2.json").write_text(nc2.to_json())
+    (nc_dir / _NOT_COMPLETED_TABLE / "nc3.json").write_text(nc3.to_json())
+    len_not_completed = len(nc_dstore.not_completed)
+    assert len_not_completed == 3
+
+
 '''
 def test_write_wout_suffix(w_dstore):
```
GavinHuttley, line 227: delete the literal string block

src/cogent3/app/data_store_new.py
```diff
@@ -93,6 +97,8 @@ def write(self, unique_id: str, data: str | bytes) -> None:
             raise IOError("datastore is readonly")
         if self._if_member_exists is RAISE and unique_id in self:
             raise IOError("member exists")
+        if self._if_member_exists is READONLY and unique_id in self:
```
GavinHuttley, line 100: you could merge these two conditional
```python
if self._if_member_exists in (READONLY, RAISE) and unique_id in self:
    raise IOError(f"DataStore member_exists is {self._if_member_exists!r} and {unique_id!r} exists")
```

GavinHuttley (changes requested):
small number of changes

src/cogent3/app/data_store_new.py
Nick-Foto: Done

tests/test_app/test_data_store_new.py
Nick-Foto: Done

src/cogent3/app/data_store_new.py
Nick-Foto: Done

tests/test_app/test_data_store_new.py
```diff
@@ -208,108 +208,30 @@ def test_append(w_dstore):
...
 
-'''
 
-# todo: check combinations of args to constructor
-# todo: new tests for describe and summary
+def test_no_not_completed_subdir(nc_dir, nc_dstore):
+    nc_dstore.drop_not_completed()
+    Path(nc_dstore.source / _NOT_COMPLETED_TABLE).rmdir()
+    assert not Path(nc_dstore.source / _NOT_COMPLETED_TABLE).exists()
+    expect = f"6x member DataStoreDirectory(source='{nc_dir}', members=[brca1.fasta, long_testseqs.fasta, formattest.fasta...)"
```
GavinHuttley, line 227: this test is fragile in the sense that if we change (for example) the class name the test will break. I suggest changing to
```python
expect = "6x member"
assert repr(nc_dstore).startswith(expect)
```

src/cogent3/app/data_store_new.py
```diff
@@ -426,14 +426,11 @@ def summary_not_completed(self) -> TabularType:
         return Table(header=header, data=rows, title="not completed records")
 
     def __repr__(self):
-        notcompleted_path = Path(self.source / _NOT_COMPLETED_TABLE)
-        if not notcompleted_path.exists():
-            return "This isn't a results directory"
         num_completed = len(self)
         num_not_completed = len(self.not_completed)
         name = self.__class__.__name__
         sample = f"{str(list(self[:3]))[:-1]}..." if num_completed > 3 else list(self)
```
GavinHuttley, line 432: not sure why slicing twice ([:3] then [:-1]), also change to being a `repr()` rather than string
```python
sample = f"{list(self[:2])}..." if num_completed > 3 else list(self)
```

src/cogent3/app/data_store_new.py
```diff
@@ -239,9 +249,14 @@ def md5(self, unique_id: str, force=True) -> str:
         self._md5 = md5_setting
         return self._checksums.get(unique_id, None)
 
+    def drop_not_completed(self):
+        nc_dir = self.source / _NOT_COMPLETED_TABLE
+        for file in list(nc_dir.glob("*.json")):
+            file.unlink()
+
```
GavinHuttley, line 256: also delete corresponding `md5.txt` files, plus add test for this

src/cogent3/app/data_store_new.py
```diff
@@ -272,53 +287,150 @@ def completed(self) -> list[DataMember]:
...
+        unique_id = unique_id.replace(self.suffix, "txt")
+        unique_id = unique_id if cmp is None else unique_id.replace(f".{cmp}", "")
+        with open_(self.source / _NOT_COMPLETED_TABLE / unique_id, mode="w") as out:
+            out.write(md5)
 
     def write_log(self, unique_id: str, data: str) -> None:
         super().write(unique_id, data)
+        # check suffix compatible with this datastore
+        suf, cmp = get_format_suffixes(unique_id)
+        self.suffix = self.suffix or suf
```
GavinHuttley, line 334: should NOT be using `self.suffix` for logs. Should use `.log`

src/cogent3/app/data_store_new.py
```diff
@@ -272,53 +287,150 @@ def completed(self) -> list[DataMember]:
...
+        unique_id = unique_id if cmp is None else unique_id.replace(f".{cmp}", "")
+        with open_(self.source / _MD5_TABLE / unique_id, mode="w") as out:
+            out.write(md5)
 
     def write_not_completed(self, unique_id: str, data: str) -> None:
         super().write(unique_id, data)
+        # check suffix compatible with this datastore
+        suf, cmp = get_format_suffixes(unique_id)
+        self.suffix = self.suffix or suf
+        if suf != self.suffix:
```
GavinHuttley, line 320: should not be using `self.suffix` for not completed, should be using ".json", but see suggested refactor

src/cogent3/app/data_store_new.py
```diff
@@ -272,53 +287,150 @@ def completed(self) -> list[DataMember]:
...
+                for m in nc_dir.glob("*.json")
             ]
-            if ic_dir.exists()
+            if nc_dir.exists()
             else []
         )
 
     def write(self, unique_id: str, data: str) -> None:
         super().write(unique_id, data)
+        # check suffix compatible with this datastore
```
GavinHuttley, line 302: there is a lot of duplication between these write methods. I suggest we create a `_write()` method that generalises this logic, taking the expected suffix for each output type etc.. For example
```python
def _write(self, subdir: str, identifier: str, suffix: str, data: str, mdf5: bool) -> None:
    sfx, cmp = get_format_suffixes(unique_id)
    if sfx != suffix:
        ...
    with open_(self.source / subdir) as out:
        ...
    if md5:
        ...
```
Then get all the write methods to use it
```python
def write(self, unique_id: str, data: str) -> None:
        self._write("", unique_id, self.suffix, data, True)

def write_log(self, unique_id: str, data: str) -> None:
        self._write(_LOG_TABLE, unique_id, "log", data, False)

```

src/cogent3/app/data_store_new.py
```diff
@@ -272,53 +287,150 @@ def completed(self) -> list[DataMember]:
...
     def describe(self) -> TabularType:
-        ...
+        num_notcompleted = len(self.not_completed)
+        num_completed = len(self.completed)
+        num_logs = len(self.logs)
+        return Table(
+            header=["record type", "number"],
+            data=[
+                ["completed", num_completed],
+                ["incomplete", num_notcompleted],
```
GavinHuttley, line 353: For consistency, change
```python
["not_completed", num_notcompleted]
```

GavinHuttley (changes requested):
Have been a bit more thorough this time. In think we can simplify with a refactor.

GavinHuttley (changes requested):
Minor changes required to complete this

tests/test_app/test_data_store_new.py
```diff
@@ -30,8 +30,8 @@ def retain_nc_dstore(dstore):
     ]
     for i, item in enumerate(nc):
         dstore.write_not_completed(f"nc{i+1}", item.to_json())
-    assert 3 == len(dstore.not_completed)
-    assert 3 == len(list((ncdir / _MD5_TABLE).glob("*.txt")))
+    assert len(dstore.not_completed) == 3
```
GavinHuttley, line 33: Delete `retain_nc_dstore()` and all calls to it.

Replace it by setting `pytest.fixture(scope="function")` ([see pytest docs](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html)). With that setting, pytest will tear down and build the fixture for each function call.

GavinHuttley (changes requested):
use pytest to handle object setup and teardown

plus fix the members attribute so it's the concatenation of `completed` and `not_completed`. You'll need to modify tests that have relied on `len(dstore)` too.

PR #1165, opened 2022-10-09
MAINT: tidy code and tests

GavinHuttley:


tests/test_app/test_data_store_new.py
```diff
@@ -65,27 +69,12 @@ def nc_dir(tmp_dir):
...
-        Path(md5_dir).rmdir()
-    if Path(nc_dir).exists():
-        Path(nc_dir).rmdir()
     nc_dir.mkdir(parents=True, exist_ok=True)
     logs_dir.mkdir(exist_ok=True)
     not_dir.mkdir(exist_ok=True)
     (logs_dir / "scitrack.log").write_text((DATA_DIR / "scitrack.log").read_text())
-    return nc_dir
+    yield nc_dir
+    shutil.rmtree(nc_dir)
```
Nick-Foto, line 78: I have some small changes. I think it is better to put it in a commit instead of explaining all here. If you merge this, I will push it.

PR #1167, opened 2022-10-10
DEV: Change _write function and how to add members-Add new test for _…

Nick-Foto:
…limit in DataStoreDirectory


Nick-Foto:
@GavinHuttley 
Writing test for `self._limit` guided me to find some problems in the code.

Nick-Foto:
@GavinHuttley 
I think we may need a function to reload or reset `DataStoreDirectory`. When we have members, `self._completed` is created with list of members, then if user wants to reread files and reload all members, they can use this function to clear `self._completed` list and remake all members.
And what about `self._not_completed`?

PR #1169, opened 2022-10-11
MAINT: tidy code

GavinHuttley:


src/cogent3/app/data_store_new.py
```diff
@@ -186,66 +188,57 @@ def __init__(
...
 
     def read(self, unique_id: str) -> str:
         """reads data corresponding to identifier"""
-        # before open should make absolute path
-        with open_(self.source / unique_id) as input:
-            data = input.read()
-            if self._md5 and isinstance(data, str):
-                self._checksums[unique_id] = get_text_hexdigest(data)
+        with open_(self.source / unique_id) as infile:
+            data = infile.read()
```
Nick-Foto, line 221: I think you have removed `checksum` and `get_text_hexdigest`. Without `get_text_hexdigest`, how we have md5?

GavinHuttley: The `md5` is calculated and written on the write operation. This is in the `_write()` method under the `md5` clause. ([See here](https://github.com/cogent3/cogent3/pull/1169/files#diff-ae10de901aabb1e7506cd860c9fd7b0d9b464b7ccd6285e25df18d00d611325dL324).)

PR #1171, opened 2022-10-14
Data store refactor

Nick-Foto:


src/cogent3/app/data_store_new.py
```diff
@@ -431,23 +436,23 @@ def __repr__(self):
 
 @singledispatch
 def get_data_source(data) -> str:
-    ...
-
+    raise NotImplementedError("Cannot resolve a unique identifier from data")
```
GavinHuttley, line 439: change message to
```python
f"Cannot resolve a unique identifier from {type(data)}")
```

GavinHuttley (changes requested):
Very hard to review `io_new.py` given whole module is new and it seems to be dominantly duplicated. Can you reduce it to ONLY the new writer, removing all duplicates from `io.py`

Nick-Foto:
> Very hard to review `io_new.py` given whole module is new and it seems to be dominantly duplicated. Can you reduce it to ONLY the new writer, removing all duplicates from `io.py`

@GavinHuttley 
I had deleted all of those old sections, but I don't know why it is not included in the commit. Sorry, maybe I didn't add it because I assumed it as part of `as_completed` section, by mistake.
Now, I see I didn't add final state of test_io_new.py. Now, I pushed all these changes.

Nick-Foto:
@GavinHuttley 
I added new file `composable_new.py`, because new changes we have made in `composable.py` break old tests.

PR #1172, opened 2022-10-17
TST: fixed one of the xfail marked apply_to tests

GavinHuttley:


src/cogent3/app/composable.py
```diff
@@ -1380,16 +1380,19 @@ def _apply_to(
...
     for result in self.as_completed(inputs, parallel=parallel, par_kw=par_kw):
-        member = self(data=result.obj)  # which means writers must return DataMember
-        logger.log_message(member, label="output")
-        logger.log_message(member.md5, label="output md5sum")
+        member = self.main(
+            data=result.obj, identifier=get_data_source(result.source)
+        )  # writers must return DataMember
+        md5 = self.data_store.md5(member.unique_id)
+        logger.log_message(str(member), label="output")
+        logger.log_message(md5, label="output md5sum")
```
Nick-Foto, line 1388: I think we can add a method to `DataMember` as `write_to_logger` to put these lines in that method and call:
`member.write_to_logger(logger).`

Nick-Foto: If you agree, I can add this in the next PR.

GavinHuttley: I think the key next step is to incrementally resolve the xfails. When all the tests are and passing then we can look at over all design

PR #1174, opened 2022-10-18
Data store refactor

Nick-Foto:


src/cogent3/app/composable.py
```diff
@@ -1391,9 +1378,9 @@ def _apply_to(
         member = self.main(
             data=result.obj, identifier=get_data_source(result.source)
         )  # writers must return DataMember
-        md5 = self.data_store.md5(member.unique_id)
-        logger.log_message(str(member), label="output")
-        logger.log_message(md5, label="output md5sum")
+    md5 = member.md5
```
Nick-Foto, line 1381: I added this, because in the old `DataStoreMember`, we didn't have `unique_id`.

GavinHuttley (changes requested):
important issue in _as_completed, otherwise very nice!

src/cogent3/app/composable.py
```diff
@@ -1391,9 +1378,9 @@ def _apply_to(
         member = self.main(
             data=result.obj, identifier=get_data_source(result.source)
         )  # writers must return DataMember
-        md5 = self.data_store.md5(member.unique_id)
```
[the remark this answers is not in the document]
Nick-Foto: @GavinHuttley 
I fixed this.
I also added a function `write_to_logger` to move this code into it.

src/cogent3/app/data_store_new.py
```diff
@@ -173,6 +173,14 @@ def unique_id(self):
     def source(self):
         return self.unique_id
 
+    @property
+    def md5(self):
+        return self.data_store.md5(self.unique_id)
+
+    def write_to_logger(self, logger):
```
GavinHuttley, line 180: Doesn't belong on the DataMember. This is wrapping a logging method, revert this and resurrect the original code inside the loop.

src/cogent3/app/data_store.py
```diff
@@ -125,6 +125,11 @@ def close(self):
     def md5(self):
         return self.parent.md5(self, force=True)
 
+    def write_to_logger(self, logger):
```
GavinHuttley, line 128: 3 lines of code has now become 6. Delete this method. See below.

GavinHuttley (changes requested):
I think the DataMember should only deal with reading, writing, unique id and md5 of it's data. The logging code belongs inside the loop as it is not used anywhere else and is just 3 lines.

PR #1178, opened 2022-10-25
API: apply_to changed get_source argument name to id_from_source

GavinHuttley:


src/cogent3/app/composable.py
```diff
@@ -1311,7 +1311,7 @@ def is_composable(obj):
 def _apply_to(
     self,
     dstore,
-    get_data_source: callable = get_data_source,
+    id_from_source: callable = get_data_source,
```
Nick-Foto, line 1314: @GavinHuttley 
Sorry, I forgot to do this change in the previous PR.
Is it good to rename `get_data_source` function, too.

Nick-Foto: Because, as you renamed the argument, we don't get source. We get id from source in that function.

GavinHuttley: I agree, but I think that will require more time than I have right now.

PR #1200, opened 2022-11-08
ENH: improved robustness of composable apps

GavinHuttley:


tests/test_app/test_typing.py
```diff
@@ -1,4 +1,4 @@
-from typing import List, Tuple, Union
+from typing import List, Set, Tuple, Union
```
Nick-Foto, line 1: Why we don't use   list, tuple, set  here?

GavinHuttley: we do in the tests [see here](https://github.com/cogent3/cogent3/pull/1200/files#diff-f28422feb4ea07b15d9eb51efd9c0efdcd0f73ee20856cf1117e3f407403bfaaR101)

Nick-Foto: OK. Thank you.

GavinHuttley: now expanded those into 2 separate tests since the type subscripting doesn't work for python 3.8, using pytests skip decorator

Issue #1195, opened 2022-11-03
A db API for handling annotations


GavinHuttley:
Consider an example. Using the `RichGenbankParser` to load a bacterial genome takes forever because the annotations are always converted into `_Annotatable` instances. There are too many objects created on the fly, better to only create these objects on demand.

As a step towards addressing this, I propose we implement an `AnnotationDbABC` (abstract base class) which defines two key methods.
- `find_records()` which takes a `seq_name`, `bio_type`, `identifier`, `start`, `end`, and `strand` arguments and returns a dictionary with the essential components for building a `_Annotatable`.
- `describe()` which takes a `seq_name`, `bio_type`, `identifier` arguments as booleans for whether to return the distinct values for these. This method should return a dict of `{<arg_name>: set(<results>), ...}`

The two major use-cases to solve concern annotations in GFF3 or GenBank formats. So the issue will use the existing `cogent3` parsers to load.

In summary:

- add new file `cogent3/core/annotation_db.py`
    - `annotation_db` contains the abstract base class and 2 subclasses, one for GenBank and one for GFF3. The latter two should use the built in sqlite db engine.
- add new file `tests/test_core/test_annotation_db.py`
    - tests of the above

kiratalreja3:
Hey @GavinHuttley , I would like to take this up!

GavinHuttley:
fixed in e330639

PR #1201, opened 2022-11-10
db API for handling annotations added

kiratalreja3:
address the issue https://github.com/cogent3/cogent3/issues/1195

added the `annotation_db.py` and `test_annotationdb.py` files


GavinHuttley (changes requested):
tests fail due to hard-coded paths. Please look at other test files for examples on how to handle this. In short, add move files into tests/data and commit. Then refactor tests to derive relative paths.

GavinHuttley:
Please note how other files in cogent3 define module attributes which identify the `__contributors__,` `__version__` etc.. You will also need to add those lines to both the code and test files. Given my contributions to the implementation, my name needs to be in both

src/cogent3/core/annotation_db.py
```diff
@@ -14,6 +14,14 @@
 from cogent3.parse.genbank import MinimalGenbankParser
 from cogent3.parse.gff import gff_parser
 
+__author__ = "Kirat Alreja, Gavin Huttley"
+__copyright__ = "Copyright 2007-2022, The Cogent Project"
+__credits__ = ["Rob Knight", "Peter Maxwell", "Matthew Wakefield", "Gavin Huttley"]
```
GavinHuttley, line 19: change credits to just our names

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
+from cogent3 import load_unaligned_seqs, open_
+from cogent3.core.annotation_db import GenbankAnnotationDb, GffAnnotationDb
+from cogent3.parse.genbank import MinimalGenbankParser
+from cogent3.parse.gff import gff_parser
+import os
+
+base_path = os.path.dirname(os.path.dirname(__file__))
```
GavinHuttley, line 7: better to use pathlib for these. Also, since they're effectively constants, they should be allcaps.

```python
import pathlib # add at top of file

DATA_DIR = pathlib.Path(__file__).parent.parent
GFF_PATH = DATA_DIR / "prok_NoLocusTags.gff"
GB_PATH = DATA_DIR / "annotated_seq.gb"
```

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    number of rows for different combinations of bio_type/identifier"""
+
+    import pytest
+
+    db = GenbankAnnotationDb(genbank_path)
+    with open_(genbank_path) as infile:
+        data = list(MinimalGenbankParser(infile.readlines()))
+    record = data[0]
+
+    """only the bio type is provided, along with start and end"""
```
GavinHuttley, line 362: don't include literal strings in tests, just make it a comment since that doesn't take memory. Fix all such occurrences.

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+        expected += 1
+
+    assert got == expected
+
+
+def test_db_query_genbank():
+    """Test that the SQL query returns the correct
+    number of rows for different combinations of bio_type/identifier"""
+
+    import pytest
```
GavinHuttley, line 355: do pytest import at top of file, not inside function -- remove all such statements

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    expected = len(list(parsed_gff))
+    assert got == expected
+
+
+def test_db_query():
+
+    """Test that the SQL query returns the correct
+    number of rows for different combinations of bio_type/identifier"""
+
+    seqs = load_unaligned_seqs(fasta_path, moltype="dna")
```
GavinHuttley, line 174: you don't need `seq`, or `fasta_path` or the associated file. You only need a length! Remove all references to both and just define the sequence length at the top of the test function.

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    assert got == expected
+
+
+def test_find_records():
+
+    """Test that the coordinates the grouped correctly, and features
+    formed properly"""
+
+    db = GffAnnotationDb(gff_path)
+    seqs = load_unaligned_seqs(fasta_path, moltype="dna")
```
GavinHuttley, line 216: same issue with `seq` and `fasta_path` as above

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+from cogent3.core.annotation_db import GenbankAnnotationDb, GffAnnotationDb
+from cogent3.parse.genbank import MinimalGenbankParser
+from cogent3.parse.gff import gff_parser
+import os
+
+base_path = os.path.dirname(os.path.dirname(__file__))
+data_path = os.path.join(base_path, "data")
+gff_path = os.path.join(data_path, "prok_NoLocusTags.gff")
+genbank_path = os.path.join(data_path, "NC_000913.3.gb")
+fasta_path = os.path.join(data_path, "short.fa")
```
GavinHuttley, line 11: This file is not required so neither is the attribute (see comments below). Remove both.

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
+from cogent3 import load_unaligned_seqs, open_
+from cogent3.core.annotation_db import GenbankAnnotationDb, GffAnnotationDb
+from cogent3.parse.genbank import MinimalGenbankParser
+from cogent3.parse.gff import gff_parser
+import os
+
+base_path = os.path.dirname(os.path.dirname(__file__))
+data_path = os.path.join(base_path, "data")
+gff_path = os.path.join(data_path, "prok_NoLocusTags.gff")
+genbank_path = os.path.join(data_path, "NC_000913.3.gb")
```
GavinHuttley, line 10: The existing `tests/data/annotated_seq.gb` can be used for these tests. This means you don't need to add a new data file, which helps keep the repository more compact.

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+fasta_path = os.path.join(data_path, "short.fa")
+
+
+__author__ = "Kirat Alreja, Gavin Huttley"
+__copyright__ = "Copyright 2007-2022, The Cogent Project"
+__credits__ = ["Rob Knight", "Peter Maxwell", "Matthew Wakefield", "Gavin Huttley"]
```
GavinHuttley, line 16: change `__credits__` so it just includes our names

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    expected = list((list(parsed_gff)[0]).keys())[:-1]
+    assert got == expected
+
+
+def test_make_sql_query():
+    """Test that the SQLlite queries are correctly formed"""
+
+    db = GffAnnotationDb(gff_path)
+
+    """only the bio_type is provided"""
```
GavinHuttley, line 44: make a comment (does not use memory), not a literal string (uses memory)

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    got = db._make_sql_query(
+        seq_name=None,
+        bio_type=None,
+        identifier="RandomIdentifier",
+        start=None,
+        end=None,
+    )
+    assert got == expected
+
+    """both identifier and bio_type provided"""
```
GavinHuttley, line 62: remove all literal strings within functions, you can leave the docstrings under the function definition.

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    assert got == expected
+
+    """unique ID, single row returned"""
+    got = len(db.find_records(start=0, end=len(seq), identifier="id020000"))
+    expected = 1
+    assert got == expected
+
+
+def test_describe():
+
```
GavinHuttley, line 241: docstrings should be immediately below function signature line.

tests/test_core/test_annotationdb.py
```diff
@@ -0,0 +1,633 @@
...
+    """check exception when only end is provided"""
+
+    with pytest.raises(ValueError):
+        db._make_sql_query(
+            seq_name=None, bio_type=None, identifier=None, start=None, end=1000
+        )
+
+
+if __name__ == "__main__":
```
GavinHuttley, line 623: remove the `if __name__` block -- not required. `pytest` will find these tests without.

GavinHuttley (changes requested):
First, thanks for making the contribution.

There are several changes that need to be made before this can be accepted. Please read all the comments on the PR very carefully and address them at all occurrences in your code (if in they aren't explicitly commented).

Most important, it is undesirable to add large data files to a repository. This causes a lot of problems down the track. You should strip those files from your repository history and convert your tests to using the existing ones. It is ok to add smaller files (such as the gff sample). But your commit message should state where you got them from. In the case of `prok_...gff` we got the file from https://ftp.ncbi.nlm.nih.gov/asn1-converters/by_program/table2asn/DOCUMENTATION/.

In this case, there is already a genbank formatted record that can be used instead. (See the comments for details).

GavinHuttley:
Closing this as the API needs to be redone and the author has stopped responding

PR #1202, opened 2022-11-16
Restrict type hint complexity on composed apps

GavinHuttley:


src/cogent3/app/typing.py
```diff
@@ -121,3 +121,28 @@ def hints_from_strings(*strings: Iterable[str]) -> list:
...
+    if not level_type:
+        return depth + 1, hint
+
+    levels = []
+    depths = []
+    for arg in get_args(hint):
+        d, t = type_tree(arg, depth=depth)
+        levels.append(t)
+        depths.append(d)
+    depth = max(depths) + 1
```
Nick-Foto, line 138: Can we use depths[-1] to find max depth? because I think the last is the max.

GavinHuttley: For the following case `list[tuple[set[int]],set[int]]` the first element of `depths` is the maximum. So we have to use `max(depths)`

tests/test_app/test_composable.py
```diff
@@ -1000,5 +1001,16 @@ def take_int2(val: int) -> int:
...
+def test_complex_type(first, ret):
+    # disallow >2-deep nesting of types
+    with pytest.raises(TypeError):
+
+        @define_app
+        class x:
+            def main(self, data: first) -> ret:
+                return data
+
+
```
Nick-Foto, line 1014: I would add test for depth 1, depth 2 and depth 4.

Issue #1196, opened 2022-11-05
DEV: explore different code coverage tools


GavinHuttley:
getting too many CI failures due to codecov.

We need a tool that supports GitHub actions.

One suggestion is https://coveralls.io

Another is outlined by [Hynek](https://hynek.me/articles/ditch-codecov-python/)

Nick-Foto:
@GavinHuttley  
Did you try this option:
fail_ci_if_error: false

Nick-Foto:
@GavinHuttley   
Did you see this page:
https://github.com/codecov/codecov-action/issues/598

GavinHuttley:
> @GavinHuttley  
> 
> Did you try this option:
> 
> fail_ci_if_error: false

That doesn't solve issue since I rely on the codecov report as part of PR evaluation.

GavinHuttley:
> @GavinHuttley   
> 
> Did you see this page:
> 
> https://github.com/codecov/codecov-action/issues/598

There are **too many** such issues. None of the ones I've seen explain why our uploads work sometimes, but not all the time.

My conclusion is we need to replace codecov. The replacement is not obvious to me.

GavinHuttley:
I have coveralls working for SciTrack. See [the workflow file there](https://github.com/HuttleyLab/scitrack/blob/develop/.github/workflows/testing_develop.yml). You will also need to modify the `noxfile.py` so that it matches the cov output format and path setting as per the [scitrack version](https://github.com/HuttleyLab/scitrack/blob/develop/noxfile.py).

PR #1207, opened 2022-11-27
Changes to handling of unique_id on write operations

GavinHuttley:


src/cogent3/app/sqlite_data_store.py
```diff
@@ -271,6 +272,9 @@ def drop_not_completed(self) -> None:
         self._not_completed = []
 
     def write(self, *, unique_id: str, data: str) -> DataMember:
+        if unique_id.startswith(_RESULT_TABLE):
+            unique_id = Path(unique_id).name
+
```
Nick-Foto, line 277: Interestingly, today I was thinking about this.
I think wo should not need to add table_name in calling all write functions.
We both was thinking on this topic at the same time.
If we can get rid of suffix, it will be great.

tests/test_app/test_sqlite_data_store.py
```diff
@@ -260,26 +260,30 @@ def test_read_log(sql_dstore, log_data):
...
+    m = sql_dstore.write(unique_id=identifier, data=expect)
+    got = m.read()
     assert got == expect
     sql_dstore._mode = OVERWRITE
-    identifier = "result_table/brca1.fasta"
-    sql_dstore.write(unique_id=identifier, data=expect)
-    got = sql_dstore.read(identifier)
+    identifier = "brca1.fasta"
+    m = sql_dstore.write(unique_id=identifier, data=expect)
+    got = m.read()
```
Nick-Foto, line 286: Yesterday I had problems here, because `write` function had problem. `write` function returns a new `DataMember` after updating and adds it to the datastore, so we will have two similar members, but it should return None when it updates a DataMember.
I have done two fixes for this, in my code:
1 -  I fixed `write` function to return None after updating . So it won't add new member to datastore.
2 - I have added
`assert len_dstore == len(full_dstore)`
here to check updating not adding new member.

Nick-Foto: I have these changes in my code. You don't need to do anything.

GavinHuttley: I don't think it should ever return `None`. If a member already exists in the collection, then it should return that instance

src/cogent3/app/sqlite_data_store.py
GavinHuttley: Not having the table name as part of the `unique_id` is appealing. All it should require is that our read method be modified to discriminate between logs and results. I will look at this.

tests/test_app/test_sqlite_data_store.py
Nick-Foto: @GavinHuttley 
Because after calling `_write`, it adds new member to the list. if we return the existing member, it will add it as a new member.

GavinHuttley:
Merging despite failed test suite as the latter is due to codecov, not breaking code. We need to revisit the comments above after we've switched to using coverall as outlined in #1196

PR #1208, opened 2022-11-30
Issue1173

khiron:
resolves cogent3#1173

Implement Jaccard distance function jaccard() using set/frozenset/numpy.ndarray boolean set algebra  

Deprecate binary_dist_jaccard()


src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,47 @@ def binary_dist_hamming(datamtx, strict=True):
...
-def binary_dist_jaccard(datamtx, strict=True):
+def fast_jaccard(x: iter, y: iter):
+    xset = set(x)
+    yset = set(y)
+    return 1 - len(xset & yset) / len(xset | yset)
+
+
+@singledispatch
+def jaccard(x, y):
+    """Jaccard distance between two sets is a measure of similarity between two sets
```
GavinHuttley, line 1278: this is the only docstring required

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,47 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    (0) identical, (1) dissimilar"""
+    raise NotImplementedError(f"jaccard distance not implemented for {type(x)}")
+
+
+@jaccard.register
+def _(x: set, y: set):
+    """Jaccard distance between two sets is a measure of similarity between two sets
```
GavinHuttley, line 1285: this docstring not required.

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,47 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    (0) identical, (1) dissimilar"""
+    if not x or not y:  # two empty sets are identical
+        return 0
+    return 1 - len(x & y) / len(x | y)
+
+
+@jaccard.register
+def _(x: frozenset, y: frozenset):
+    """Jaccard distance between two sets is a measure of similarity between two sets
```
GavinHuttley, line 1294: docstring not required

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,47 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    (0) identical, (1) dissimilar"""
+    if not x or not y:  # two empty sets are identical
+        return 0
+    return 1 - len(x & y) / len(x | y)
+
+
+@jaccard.register
+def _(x: numpy.ndarray, y: numpy.ndarray):
+    """Jaccard distance between two sets is a measure of similarity between two sets
```
GavinHuttley, line 1303: docstring not required

GavinHuttley (changes requested):
The `singledispatch` decorator correctly handles the docstring, i.e. calling help on the function name returns that from the original decorated function, so it's safe to remove all the others.

khiron:
Updated the pull request to pass all tests fixing a trivial logic error, and removed redundant documentation.

khiron:
BTW: I came up with a generic jaccard function that would work with any two iterable classes

def iter_jaccard(x : iter, y: iter):
    # create dicts populated with the iterables which will use hash collision to de-dupe them
   
    xset = {i: None for i in x}
    yset = {i: None for i in y}
    intersection = sum(k in yset for k in xset)
    union = len(xset) + len(yset) - intersection    
    return 1 - intersection / union
 
However, it is about half the speed of using Python set algebra or Numpy array functions.  I don't know if there is any value to including that as a SingleDispatch fallback

src/cogent3/maths/distance_transform.py
```diff
@@ -61,6 +63,7 @@
     argsort,
     array,
     asarray,
+    intersect1d,
```
GavinHuttley, line 66: The LGTM service identified this as an unused import (see the "conversation" page for this PR).

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,41 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    xset = set(x)
+    yset = set(y)
+    return 1 - len(xset & yset) / len(xset | yset)
+
+
+@singledispatch
+def jaccard(x, y):
+    """Jaccard distance between two sets is a measure of similarity between two sets
+    (0) identical, (1) dissimilar"""
+    raise NotImplementedError(f"jaccard distance not implemented for {type(x)}")
```
GavinHuttley, line 1280: Change the error message to
```python
return NotImplementedError(f"jaccard distance handles sets or numpy arrays, not {type(x)}")
```

src/cogent3/maths/distance_transform.py
```diff
@@ -52,6 +52,8 @@
 
 """
 
+from functools import singledispatch
```
GavinHuttley, line 55: add an import
```python
from typing import Union
```

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,41 @@ def binary_dist_hamming(datamtx, strict=True):
     return dists
 
 
-def binary_dist_jaccard(datamtx, strict=True):
+def fast_jaccard(x: iter, y: iter):
+    xset = set(x)
+    yset = set(y)
+    return 1 - len(xset & yset) / len(xset | yset)
+
+
```
GavinHuttley, line 1275: define a union type
```python
T = Union[set,frozenset,ndarray]
```

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,41 @@ def binary_dist_hamming(datamtx, strict=True):
...
+@singledispatch
+def jaccard(x, y):
```
GavinHuttley, line 1277: fully type hint signature
```python
def jaccard(x: T, y: T) -> float:
```

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,41 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    """Jaccard distance between two sets is a measure of similarity between two sets
+    (0) identical, (1) dissimilar"""
+    raise NotImplementedError(f"jaccard distance not implemented for {type(x)}")
+
+
+@jaccard.register
+def _(x: set, y: set):
+    if not x and not y:  # two empty sets are identical
+        return 0
```
GavinHuttley, line 1286: make sure all returns are floats, so change `0` to `0.0`. This will actually have an impact in Python 3.11 (when we migrate to it) where there's a new builtin performance optimisation capability

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,41 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    return 1 - len(x & y) / len(x | y)
+
+
+@jaccard.register
+def _(x: frozenset, y: frozenset):
+    if not x and not y:  # two empty sets are identical
+        return 0
```
GavinHuttley, line 1293: change `0` to `0.0`

src/cogent3/maths/distance_transform.py
```diff
@@ -1253,7 +1267,41 @@ def binary_dist_hamming(datamtx, strict=True):
...
+    return 1 - len(x & y) / len(x | y)
+
+
+@jaccard.register
+def _(x: numpy.ndarray, y: numpy.ndarray):
+    if not x.any() and not y.any():  # two empty sets are identical
+        return 0
```
GavinHuttley, line 1300: change `0` to `0.0`

GavinHuttley (changes requested):
Very small changes, and thanks for the contribution! Last review comments, I promise!

On the suggestion of a fallback algorithm, it could be simplified to be nearly identical to your `fast_jaccard()`
```python
@singledispatch
def jaccard(x, y):
    # create dicts populated with the iterables which will use hash collision to de-dupe them
   return jaccard(set(x), set(y))
```
That looks like it's recursive, but I'll bet it isn't and that the last call triggers the set-handling version of the function.

However, I don't suggest we do it mainly because there are iterables that will fail with an uninformative message, e.g. if `x` and `y` are lists of lists they will raise a `TypeError` because they're not hashable. The user then has to figure out why that's happening. In our current case, they get an exception that the data type they've tried to use is not supported for jaccard, which is when I realised we need a better exception message!

src/cogent3/maths/distance_transform.py
khiron: Yes, that's much more informative.  Will make the change.

src/cogent3/maths/distance_transform.py
khiron: I think the only problem with this will be the implication that the function allows mixed types, eg: calculating the difference between a set and a NDArray, or vice versa.  And now I think of it, the current implementation won't handle calculating the difference between a set and a frozenset.  I suspect your deceptively not-recursive suggestion of casting to sets then calling the set,set function might handle this.  I'll have a play with it and see if I can't make something elegant do the job.

src/cogent3/maths/distance_transform.py
khiron: ah yes.  good call

src/cogent3/maths/distance_transform.py
GavinHuttley: `set()` and `frozenset()` instances can operate together. So no need to worry about that mixture.

I get what you mean about the type hint. I wonder if code editors like VS Code, PyCharm, will infer that `x` and `y` should have the same type and provide feedback to user when their usage doesn't match.

More importantly, we need to set limits to what we should try and handle.

For instance, there are limits to the `singledispatch` mechanism, mainly that it delegates based solely on the type of the first argument. For example, if a user invokes jaccard with `type(x)==set` and `type(y)==ndarray`, the set variant of `jaccard` will be invoked, and fail.

My view on cases such is that this reflects a fundamental error in the users code. Better to allow an exception happen and let them identify that error.

khiron: I tried probing singledispatch to verify that, and it's behaviour is perhaps even more disturbing.  If you have multiple prototypes with the same type for the first parameter, the last one is chosen.  I assume the internal data structure is a dict with the type of the first parameter as the key and a pointer to the function as the value, and subsequent implementations collide and overwrite the entry for the first parameter type.  That design is likely to cause developers relying upon duck typing magic to get into trouble.

I addressed the mixed type issue that can't be resolved programmatically using singledispatch by altering the doc comment

GavinHuttley: Document is the right way to go!

PR #1223, opened 2022-12-08
Data store refactor-Add lock unlock to sqlitedb

Nick-Foto:


src/cogent3/app/sqlite_data_store.py
```diff
@@ -295,6 +277,64 @@ def drop_not_completed(self) -> None:
...
+        # inadvertently overwrite something otherwise.
+        # BUT if mode=a, as the user expects to be modifying the data
+        # store then we have to update the value
+        if self.mode is not READONLY:
+            result = (
+                None
+                if self.mode is APPEND
+                else self._db.execute("SELECT state_id,lock_pid FROM state")
+            ).fetchall()
+            unlocked = False
```
GavinHuttley, line 293: `unlocked` seems pointless

src/cogent3/app/sqlite_data_store.py
```diff
@@ -295,6 +277,64 @@ def drop_not_completed(self) -> None:
...
+            if result and len(result):
+                unlocked = True
+            locked = result[0]["lock_pid"] if result else None
+            if locked and self.mode is OVERWRITE:
+                raise IOError(
+                    "You are trying to OVERWRITE a database which is locked. Use APPEND mode or unlock"
+                )
+            elif locked or unlocked:
```
GavinHuttley, line 301: this clause should be a separate `if`, not an `elif`. The expression is also strange. `locked` and `unlocked` are the only two states, so this seems like it will always be True

src/cogent3/app/sqlite_data_store.py
```diff
@@ -295,6 +277,64 @@ def drop_not_completed(self) -> None:
...
+                vals = [os.getpid(), state_id]
+            else:
+                cmnd = "INSERT INTO state(lock_pid) VALUES (?)"
+                vals = [os.getpid()]
+
+            self._db.execute(cmnd, tuple(vals))
+
+    @property
+    def is_locked(self):
+        """returns lock pid or None if unlocked or pid matches self"""
```
GavinHuttley, line 314: The expression returns a bool, so the docstring should be changed and a return type hint added

src/cogent3/app/sqlite_data_store.py
```diff
@@ -295,6 +277,64 @@ def drop_not_completed(self) -> None:
...
+        return (
+            self.db.execute("SELECT lock_pid from state where state_id=1").fetchall()[
+                0
+            ]["lock_pid"]
+            is not None
+        )
+
+    def unlock(self, force=False):
+        """remove a lock if pid matches. If force, ignores pid."""
```
GavinHuttley, line 323: update docstring, "ignored if mode is READONLY"

src/cogent3/app/sqlite_data_store.py
```diff
@@ -295,6 +277,64 @@ def drop_not_completed(self) -> None:
...
+                cmnd = "UPDATE state SET lock_pid=? WHERE state_id=?"
+                vals = [os.getpid(), state_id]
+            else:
+                cmnd = "INSERT INTO state(lock_pid) VALUES (?)"
+                vals = [os.getpid()]
+
+            self._db.execute(cmnd, tuple(vals))
+
+    @property
+    def is_locked(self):
```
GavinHuttley, line 313: rename to `locked`. This is equivalent to the file object property of `closed`.

GavinHuttley (changes requested):
Very close!

Refactor `is_locked` to `locked`.

Remove the `sample_not_locked.tinydb` data file. It's not being used. If we need that structure, add a fixture to produce it on demand.

src/cogent3/app/sqlite_data_store.py
Nick-Foto: The problem is to distinguish between 1- We don't have that record in `state` table, or 2- We have it and `lock_pid` field of that record is NULL.

src/cogent3/app/sqlite_data_store.py
Nick-Foto: @GavinHuttley 
I should change variable name `unlocked` to a better name.

src/cogent3/app/sqlite_data_store.py
Nick-Foto: @GavinHuttley 
I cannot make `not_locked` tinydb on demand, because we will remove tinydb support and I cannot change a tinydb file in program. I am using this file to test converting tinydb to sqlitedb. Now I added a new test for converting not_locked tinydb.

src/cogent3/app/sqlite_data_store.py
Nick-Foto: @GavinHuttley 
I changed this section of code completely. The name of variable `unlocked` didn't deliver my purpose correctly.

src/cogent3/app/sqlite_data_store.py
GavinHuttley: we can change the locked tinydb:
- int a new fixture, do the same imports as done in the `convert_tiny...` function
- copy locked file into a temp directory
- load tinydb
- remove the "LOC" record
- flush to disk
- etc...

Nick-Foto: OK

Nick-Foto: @GavinHuttley 
I know this method, but I didn't want to use anything from tinydb imports.
OK, I am going to change it.

src/cogent3/app/sqlite_data_store.py
```diff
@@ -277,50 +271,47 @@ def drop_not_completed(self) -> None:
         self.db.execute(f"DELETE FROM {_RESULT_TABLE} WHERE is_completed=0")
         self._not_completed = []
 
+    @property
+    def _lock_id(self):
+        """returns lock_pid"""
+        result = self.db.execute("SELECT lock_pid FROM state").fetchone()
+        return result[0] if result else result
```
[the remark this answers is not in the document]
Nick-Foto: It works with this form.

tests/test_app/test_data_store_new.py
```diff
@@ -143,6 +144,38 @@ def full_dstore(write_dir, nc_objects, completed_objects, log_data):
...
+
+
+@pytest.fixture(scope="function")
+def tinydbfile_notlocked(tmp_dir):
+    try:
+        from fnmatch import fnmatch, translate
+        from tinydb import Query, TinyDB
+        from tinydb.middlewares import CachingMiddleware
+        from tinydb.storages import JSONStorage
+        from cogent3.app.data_store import load_record_from_json
```
GavinHuttley, line 161: delete imports not used in this function, e.g. fnmatch, load_record..., DataStoreSq...

tests/test_app/test_data_store_new.py
```diff
@@ -143,6 +144,38 @@ def full_dstore(write_dir, nc_objects, completed_objects, log_data):
...
+        raise ImportError(
+            "You need to install tinydb to be able to migrate to new datastore."
+        ) from e
+    path = tmp_dir / "sample_locked.tinydb"
+    shutil.copy(DATA_DIR / path.name, path)
+    storage = CachingMiddleware(JSONStorage)
+    storage.WRITE_CACHE_SIZE = 50  # todo support for user specifying
+    db = TinyDB(path, storage=storage)
+    query = Query().identifier.matches("LOCK")
+    got = db.get(query)
```
GavinHuttley, line 173: delete not used
```python
got = ...
lock_pid...
```

GavinHuttley (changes requested):
simplify tests a bit

tests/test_app/test_data_store_new.py
```diff
@@ -143,6 +144,38 @@ def full_dstore(write_dir, nc_objects, completed_objects, log_data):
...
+        from tinydb import Query, TinyDB
+        from tinydb.middlewares import CachingMiddleware
+        from tinydb.storages import JSONStorage
+        from cogent3.app.data_store import load_record_from_json
+        from cogent3.app.sqlite_data_store import DataStoreSqlite
+    except ImportError as e:
+        raise ImportError(
+            "You need to install tinydb to be able to migrate to new datastore."
+        ) from e
+    path = tmp_dir / "sample_locked.tinydb"
```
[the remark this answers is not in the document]
Nick-Foto: 1 - The above rename you mentioned, renames on its own name. 
2 - When I use `tinydbfile_locked` fixture here instead of copying it, I don't have this line "path = tmp_dir / "sample_locked.tinydb"
3 - Now, I will use `rename` in defining the other fixture `tinydbfile_locked`, instead of `shutil.copy`.

src/cogent3/app/sqlite_data_store.py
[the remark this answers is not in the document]
GavinHuttley: Just to clarify, do you mean that `_lock_id()` returns an `int` or `None`? (Good to add a return type hint too.)

tests/test_app/test_data_store_new.py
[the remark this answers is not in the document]
GavinHuttley: There's a typo in my statement, it should say "sample_**un**locked.tinydb".

You need the `shutil.copy()`! Here's the process

1. copy the versioned "sample_locked.tinydb" file into a temporary directory
2. rename the copy to unlocked so at least the name reflects its state (that's the rename statement with "sample_unlocked.tinydb")
3. remove the lock
4. return the path to the renamed file

GavinHuttley (changes requested):
small edits

src/cogent3/app/sqlite_data_store.py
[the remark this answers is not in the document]
Nick-Foto: Yes, it returns int or None.

PR #1226, opened 2022-12-13
ENH: implement updated loader and writer apps

GavinHuttley:


src/cogent3/app/data_store_new.py
```diff
@@ -591,3 +595,29 @@ def convert_tinydb_to_sqlite(source: Path, dest: Optional[Path] = None) -> DataS
         dstore.db.execute(cmnd, (lock_id,))
 
     return dstore
+
+
+def make_record_for_json(identifier, data, completed):
```
Nick-Foto, line 600: Name of this function is a little ambiguous.

tests/test_app/test_io_new.py
```diff
@@ -54,7 +80,7 @@ def test_write_seqs(fasta_dir, tmp_dir):
     out_data_store = DataStoreDirectory(
         tmp_dir / "test_write_seqs", mode=Mode.w, suffix="fasta"
     )
-    writer = io_app_new.write_seqs(out_data_store, format="fasta")
+    writer = io_app.write_seqs(out_data_store, format="fasta")
```
Nick-Foto, line 83: Here , you have called old write_seqs.

src/cogent3/app/io_new.py
```diff
@@ -14,28 +45,263 @@
...
+    data = _read_it(path)
+    data = data.splitlines()
+    data = dict(iter(parser(data)))
+    seqs = klass(data=data, moltype=moltype)
+    seqs.info.source = str(path)
+    return seqs
+
+
+@define_app(app_type=LOADER)
+class load_aligned:
```
Nick-Foto, line 77: As you know, all classes have the same name as io.py. This may cause problems for users.

tests/test_app/test_io_new.py
GavinHuttley: Actually, I changed the imports at the top of the file to
```python
from cogent3.app import io_new as io_app
```

src/cogent3/app/io_new.py
GavinHuttley: Those are distinguished by name space, and I have an idea how to change how users get app instances that will make the impact on users smaller.

PR #1227, opened 2022-12-13
ENH: support for reading and writing from a db backend

GavinHuttley:


src/cogent3/app/io_new.py
```diff
@@ -45,6 +49,62 @@
 __status__ = "Alpha"
 
 
+@define_app
+def pickle_it(data: SerialisableType) -> bytes:
+    """attempts to call a to_rich_dict() method, fallback to
```
Nick-Foto, line 54: Do you thin we need type check here? What if user provides bad data?

src/cogent3/app/io_new.py
```diff
@@ -45,6 +49,62 @@
...
+    return pickle.dumps(data)
+
+
+@define_app
+def unpickle_it(data: bytes) -> SerialisableType:
+    return pickle.loads(data)
+
+
+@define_app
+class compressed:
```
Nick-Foto, line 68: I think name `compress` is better for this app.

Nick-Foto: Because the other app name is `pickle_it` not 'pickled'.

src/cogent3/app/io_new.py
```diff
@@ -45,6 +49,62 @@
...
+            function
+        """
+        self.decompressor = decompressor
+
+    def main(self, data: bytes) -> bytes:
+        return self.decompressor(data)
+
+
+@define_app
+class deserialised:
```
Nick-Foto, line 99: Do we need a `serialiser` app?

src/cogent3/app/io_new.py
GavinHuttley: Good call. If pickling doesn't work, I think it will result in a `NotCompleted` object, but a test should be added to confirm it.

src/cogent3/app/io_new.py
GavinHuttley: Perhaps `deserialised` is a bad name, since `pickle.loads()` is a deserialiser. This is specifically about "inflating" cogent3 objects. Will think on it.

src/cogent3/app/io_new.py
GavinHuttley: Done

src/cogent3/app/io_new.py
GavinHuttley: Addressed in new revisions

src/cogent3/app/io_new.py
GavinHuttley: Tested and confirmed in latest patch

PR #1229, opened 2022-12-14
move get_data_store to io_new.py and change it to open_data_store

Nick-Foto:


src/cogent3/app/io.py
```diff
@@ -139,42 +140,18 @@ def findall(base_path, suffix="fa", limit=None, verbose=False):
...
-    if zipfile.is_zipfile(base_path):
-        ds_suffix = ".zip"
-    elif base_path.suffix:
-        ds_suffix = base_path.suffix
-    else:
-        ds_suffix = None
-    klass = _datastore_reader_map[ds_suffix]
-    return klass(base_path, suffix=suffix, limit=limit)
+    """DEPRECATED, use io_new.open_data_store"""
+    warnings.warn(
```
GavinHuttley, line 144: please use the `cogent3` interface to warnings since I search for those before each release.

There are 2 convenience functions:

- `deprecated`
- `discontinued`

In this case, you want
```python
from cogent3.util.warning import deprecated
```
In this case, you would do
```python
deprecated("function", "get_data_store", "cogent3.open_data_store", "2023.3", "renamed and now a top-level import", stack_level=1)
```

The latter also means, of course, that in `cogent3.__init__` you should add the following statement
```python
from cogent3.app.io_new import open_data_store
```
Although there may be a circular import issue

tests/test_app/test_app_mpi.py
```diff
@@ -4,6 +4,7 @@
 
 from cogent3.app import align as align_app
 from cogent3.app import io as io_app
+from cogent3.app import io_new as io_app_new
```
GavinHuttley, line 7: delete the line with `io as io_app` and change `io_new ...` to` io_new as io_app`

tests/test_app/test_app_mpi.py
```diff
@@ -24,7 +25,7 @@ class MPITests(TestCase):
     @skipUnless(parallel.USING_MPI, reason="Not using MPI")
     def test_write_db(self):
         """writing with overwrite in MPI should reset db"""
-        dstore = io_app.get_data_store("data", suffix="fasta")
```
GavinHuttley, line 27: revert all uses of `io_app_new` to `io_app`

tests/test_app/test_composable.py
```diff
@@ -303,7 +303,7 @@ def main(self, val: int) -> int:
 
 def test_as_completed():
     """correctly applies iteratively"""
-    dstore = io_app.get_data_store(DATA_DIR, suffix="fasta", limit=3)
+    dstore = io_app_new.open_data_store(DATA_DIR, suffix="fasta", limit=3)
```
GavinHuttley, line 306: same issue with `io_app_new` here, except since only the `open_data_store` function is being used, you may as well use it via the new top-level import (i.e. `from cogent3 import pen_data_store`)

tests/test_app/test_io.py
```diff
@@ -52,53 +53,27 @@ def tearDown(self) -> None:
...
-        found = io_app.get_data_store(self.basedir, suffix=".fasta")
+        found = io_app_new.open_data_store(self.basedir, suffix=".fasta")
         self.assertTrue(len(found) > 1)
-        found = io_app.get_data_store(self.basedir, suffix=".fasta", limit=2)
+        found = io_app_new.open_data_store(self.basedir, suffix=".fasta", limit=2)
         self.assertTrue(len(found) == 2)
 
         # and with a suffix
-        found = list(io_app.get_data_store(self.basedir, suffix=".fasta*"))
+        found = list(io_app_new.open_data_store(self.basedir, suffix=".fasta*"))
```
GavinHuttley, line 64: use top-level import for `open_data_store`

tests/test_app/test_io_new.py
```diff
@@ -1,8 +1,11 @@
 import bz2
 import gzip
 import json
+import os
 import pickle
+import shutil
 
+from os.path import join
```
GavinHuttley, line 8: use `pathlib` only, not `os.path`

tests/test_app/test_io_new.py
```diff
@@ -445,3 +449,32 @@ def test_write_db_load_db(fasta_dir, tmp_dir):
         m = writer(orig, identifier=m.unique_id)
         read = reader(m)
         assert orig == read
+
+
+'''
+def test_findall():
```
GavinHuttley, line 455: delete these, since `findall` is to be discontinued and we will be dropping support for zip's

src/cogent3/app/io.py
GavinHuttley: Note, I never put a day on the deprecation version. Month is sufficient.

GavinHuttley (changes requested):
You need to delete the `cogent3.app.io.register_datastore_reader` class and the module level dictionary. Just import both from `io_new`.

PR #1230, opened 2022-12-15
Enhance the define_app decorator

GavinHuttley:


src/cogent3/app/io_new.py
```diff
@@ -432,9 +432,12 @@ def main(self, /, data: SerialisableType, *, identifier=None) -> T:
         identifier
         """
         identifier = identifier or get_data_source(data)
```
Nick-Foto, line 434: Here we should use `get_data_source`  or new `open_data_source` ?

GavinHuttley: Will move to `open_data_source` when that is merged in

src/cogent3/app/composable.py
```diff
@@ -1075,7 +1075,9 @@ def _main(self, arg, *args, **kwargs):
     return result
 
 
-def define_app(klass=None, *, app_type: AppType = GENERIC):
+def define_app(
+    klass=None, *, app_type: AppType = GENERIC, skip_not_completed: bool = True
+):
     """decorator for building callable apps
```
Nick-Foto, line 1081: I think default value for `skip_not_completed` should be False. Because I think most cases is False,

GavinHuttley: I interpret `skip_not_completed=True` to mean that NotCompleted's will not be processed, which is the most common use case. A developer has to take an action of setting it to False for it to change the existing behaviour.

That said, if you have a suggestion for a different name, I'm happy to hear it!

Nick-Foto: Its name is good, I thought default is to not skip.

src/cogent3/app/io_new.py
GavinHuttley: Actually, read that closer now. It's not `get_data_source` not **store**.

GavinHuttley: I get confused by those too! But we have `open_data_store` I won't get confused.

PR #1234, opened 2022-12-16
DEP: deprecating methods on _Annotatable, NEW: _AnnotationMixin class

KatherineCaley:
DEP: deprecated methods on `_Annotatable` class for `Sequence` instances.

[CHANGED] The methods that are retained on `Sequence` instances (`get_drawable`,
    `get_drawables`, `add_feature`, and `get_annotations_matching`) are moved to the
     new `_AnnotationMixin` class. Other methods will be retained on the annotation class.

### DIFF
--- a/src/cogent3/core/annotation.py
+++ b/src/cogent3/core/annotation.py
@@ -13,7 +13,7 @@ from .location import Map, as_map
 
 __author__ = "Peter Maxwell and Gavin Huttley"
 __copyright__ = "Copyright 2007-2022, The Cogent Project"
-__credits__ = ["Peter Maxwell", "Gavin Huttley"]
+__credits__ = ["Peter Maxwell", "Gavin Huttley", "Katherine Caley"]
 __license__ = "BSD-3"
 __version__ = "2022.10.31a1"
 __maintainer__ = "Gavin Huttley"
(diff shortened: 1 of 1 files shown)


src/cogent3/core/annotation.py
```diff
@@ -96,61 +194,54 @@ def __getitem__(self, index):
...
 
-    def get_drawables(self):
-        """returns a dict of drawables, keyed by type"""
-        result = defaultdict(list)
-        for a in self.annotations:
-            result[a.type].append(a.get_drawable())
-        return result
-
     def add_annotation(self, klass, *args, **kw):
+        from cogent3.core.sequence import Sequence
```
GavinHuttley, line 198: For all methods that are being flagged as being removed from the Sequence object, we need to refactor their tests.

For example, there all tests inside `test_alignment.py` that use `add_annotation`, e.g. `test_get_annotations_from_any_seq` which has uses like the following
```python
seqs.get_seq("seq1").add_annotation(Feature, "exon", "annotation1", [(3, 8)])
```
should be changed to the new method and name
```python
seqs.get_seq("seq1").add_feature("exon", "annotation1", [(3, 8)])
```

GavinHuttley (changes requested):
Look at the output of all tests, wherever there's a deprecation warning related to your change, change the test so the warning "goes away". In this case that means either using the new method name.

It may be necessary to refactor the tests and/or relocate them. Some tests in `test_alignment` or `test_sequence` may now need to be moved to `test_annotation` and their implementation changed so they're no longer based off a sequence. We can discuss those.

KatherineCaley:
Closing now, to be revisited once the annotation refactor has a clearer goal. Discussion on these changes is moved to Issue #1284!

PR #1235, opened 2022-12-16
ENH: connected up new data stores and apps

GavinHuttley:


src/cogent3/app/io.py
```diff
@@ -102,7 +102,7 @@ def get_data_store(
     return open_data_store(base_path, suffix, limit, verbose)
 
 
-def _load_seqs(path, klass, parser, moltype):
+def _load_seqs(path, klass, parser, moltype):  # pragma: no cover
     abs_path = str(path)
     if not hasattr(path, "read"):
         # we use a DataStoreMember as it's read() handles zipped compression
```
Nick-Foto, line 108: Does this function need deprecated warning?

src/cogent3/app/io_new.py
```diff
@@ -150,28 +132,33 @@ def open_data_store(
     -------
     ReadOnlyDirectoryDataStore or ReadOnlyZippedDataStore
     """
+    if not isinstance(suffix, (str, type(None))):
+        raise ValueError(f"suffix {type(suffix)} not one of string or None")
+
```
Nick-Foto, line 137: Here we can add `mode=Mode(mode)` as you always do.

src/cogent3/app/io_new.py
```diff
@@ -150,28 +132,33 @@ def open_data_store(
...
-        raise ValueError(f"'{base_path}' does not exist")
+    if base_path.suffix == ".tinydb":
+        kwargs["suffix"] = "json"
+        kwargs.pop("mode")
 
-    if type(suffix) != str:
-        raise ValueError(f"{suffix} is not a string")
-
-    if zipfile.is_zipfile(base_path):
+    if base_path.suffix == ".sqlitedb":
```
Nick-Foto, line 145: If we use `base.path.suffix`, why we have an argument `suffix` to this function?

src/cogent3/app/io_new.py
```diff
@@ -254,6 +241,7 @@ class tabular(Enum):
 
 
 def _read_it(path):
+    path = Path(path) if isinstance(path, str) else path
```
Nick-Foto, line 244: `isinstance(Path, str)` or `isinstance(path ,str)`?

tests/test_app/test_composable.py
```diff
@@ -1118,7 +1119,7 @@ def test_apply_to_only_appends(half_dstore1, half_dstore2):
     process1 = reader1 + min_length1 + writer1
     # create paths as strings
     dstore1 = open_data_store(half_dstore1.source, suffix="fasta")
-    dstore1 = [str(m) for m in dstore1]
+    dstore1 = [str(Path(m.data_store.source) / m.unique_id) for m in dstore1]
```
Nick-Foto, line 1122: I have problem with this style, if users have to use this type of access to data_store members.

src/cogent3/app/io.py
GavinHuttley: No. There is a modified version in `io_new.py`. I;ve left this one in place as it's behaviour is different due to working with the older data store classes.

src/cogent3/app/io_new.py
GavinHuttley: suffix is for the directory data store class, which identifies members  as those with the nominated ending

src/cogent3/app/io_new.py
GavinHuttley: converting a string to a `Path` instance so latter

tests/test_app/test_composable.py
GavinHuttley: They don't. This is just checking for the case where a string is provided in tests. A user should never need to do thjis since DataMember read will get them the data.

src/cogent3/app/composable.py
```diff
@@ -992,6 +992,8 @@ def _call(self, val, *args, **kwargs):
     # todo we should get the source information from val here
     if self.app_type is not LOADER and self.input:  # passing to connected app
         val = self.input(val, *args, **kwargs)
+        if isinstance(val, NotCompleted):
+            return val
```
Nick-Foto, line 996: Here, Do we need to check `skip_not_completed`?

Nick-Foto:
@GavinHuttley  
I have a suggestion about the function `def app_help(name: str)`. I think it is better to:
1 - Move this code to @define_app decorator
2 - Edit `__doc__` variable for app in the define_app decorator
3 - Change name of this function to `help` or `get_help`
4 - Change this function to return `app.__doc__`

GavinHuttley:
> @GavinHuttley I have a suggestion about the function `def app_help(name: str)`. I think it is better to: 1 - Move this code to @define_app decorator 2 - Edit `__doc__` variable for app in the define_app decorator 3 - Change name of this function to `help` or `get_help` 4 - Change this function to return `app.__doc__`

Can you post this in Zulip and we can follow up next week.

src/cogent3/app/composable.py
GavinHuttley: Good point!

Nick-Foto:
> > @GavinHuttley I have a suggestion about the function `def app_help(name: str)`. I think it is better to: 1 - Move this code to @define_app decorator 2 - Edit `__doc__` variable for app in the define_app decorator 3 - Change name of this function to `help` or `get_help` 4 - Change this function to return `app.__doc__`
> 
> Can you post this in Zulip and we can follow up next week.

Done

PR #1237, opened 2022-12-17
MAINT: more prep for deprecation of old-style data_store classes

GavinHuttley:


src/cogent3/app/data_store_new.py
```diff
@@ -502,15 +506,23 @@ def write_log(self, *, unique_id: str, data: str) -> None:
 def get_data_source(data) -> str:
```
Nick-Foto, line 506: When I look at this as a user, I get confused about naming. I think other names like `get_source_path` or `get_data_path` or a different name will be more informative.

src/cogent3/app/data_store_new.py
```diff
@@ -502,15 +506,23 @@ def write_log(self, *, unique_id: str, data: str) -> None:
...
     source = getattr(data, "source", None)
     if source is None:
```
Nick-Foto, line 508: I think our new datastores doesn't have `source` attribute. Why we use this attribute name here?

GavinHuttley: this is for data objects, it's what the unique ID is derived from

src/cogent3/app/data_store_new.py
GavinHuttley: Fair point. The biggest issue is not enough docstrings.

PR #1240, opened 2022-12-18
Rebase data_store_refactor and fix problem in returning NotCompleted in _validate_data_type

Nick-Foto:


src/cogent3/app/composable.py
```diff
@@ -1013,7 +1013,10 @@ def _call(self, val, *args, **kwargs):
             return val
 
     type_checked = self._validate_data_type(val)
-    if not type_checked:
```
GavinHuttley, line 998: I think there's a cleaner solution to this. Revert your changes in here and change the top conditional of `_validate_data_type` from
```python
def _validate_data_type(self, data):
    """checks data class name matches defined compatible types"""
    if isinstance(data, NotCompleted):
        return data
```
to
```python
def _validate_data_type(self, data):
    """checks data class name matches defined compatible types"""
    if isinstance(data, NotCompleted) and self._skip_not_completed:
        return data

```

GavinHuttley (changes requested):
Thanks for tracking this down. Please make the indicated change and then I can merge.

src/cogent3/app/composable.py
Nick-Foto: @GavinHuttley 
I changed this, but it doesn't work. It fails in one test.
Because it doesn't include one case that is NotCompleted and skip is false.

Nick-Foto: @GavinHuttley 
Done

PR #1310, opened 2023-03-07
TST: added tests of SeqView to address coverage gaps

KatherineCaley:
TST:
- added `test_subsequent_slice_neg_step` to ensure all conditions with negative steps are being tested
- added `test_seqview_step_0` to test that a slice with a step of 1 throws a `ValueError`
- added `test_seqview_repr` to test the representation of a SeqView object 

BUG: 
- removed untested condition that was both unnecessary and incorrect!


GavinHuttley (approved):
Looking good, thank you!

PR #1314, opened 2023-03-13
DEP: Discontinue Sequence.translate

AoboHill:
Discontinue cogent3.core.sequence.SequenceI.translate since this method is just delegating to the string method and it would be better if user just converts the sequence to string.


tests/lcov-3.10.info
GavinHuttley, line 1: Remove this file, it should not be included as part of repository. (Note to self, we needed to update our `.gitignore` file!)

GavinHuttley (changes requested):
- please run black and isort
- please remove the .info file
- please add to the end of your commit message ", fixes #<issue number>"

PR #1315, opened 2023-03-14
DEP: Discontinue Sequence.translate, fixes #1307

### LINKED ISSUE #1307: Discontinue Sequence.translate
It’s just delegating to the string method. Better if user just converts sequence to string.

As this is a discontinued function, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function will be removed) as 3 months from when you make the change.

AoboHill:
Discontinue Sequence.translate since this method is just delegating to the string method. Better if user just converts sequence to string.


src/cogent3/core/sequence.py
```diff
@@ -144,14 +144,22 @@ def to_json(self):
...
         -----
         This is a string method, nothing to do with translating into a
         protein sequence.
         """
+        from cogent3.util.warning import discontinued
+
+        discontinued(
+            "function",
+            "cogent3.core.sequence.SequenceI.translate",
+            "06/2023",
```
GavinHuttley, line 160: - change `"cogent3.core.sequence.SequenceI.translate"` to `"cogent3.core.sequence.Sequence.translate"` (note the deletion of `I`.
- the version should be `"2023.6"`

GavinHuttley (changes requested):
small changes

GavinHuttley (approved):

PR #1319, opened 2023-03-14
DEP: remove usages of deprecated Sequence.translate

GavinHuttley:


khiron (approved):
just a general comment about naming conventions for singledispatch functions.  I accept if the function name is irrelevant to the interpreter that it is more pythonic to use the underbar name ... I'm not convinced that _() and _() is more readable than gff_parser_using_string() and gff_parser_using_os_path()

GavinHuttley:
Thanks @khiron ! I understand the comment. I guess I'm relying in the fact that the decorator includes the name of the function, i.e. `@gff_parser.register`, and that having the same name repeated twice on consecutive lines seems even more odd (see below).

```python
@gff_parser.register
def _(...):
    ...
```
versus
```python
@gff_parser.register
def gff_parser(...):
   ...
```

PR #1321, opened 2023-03-15
DEP: `cogent3.parse.genbank.Location` refactor

### LINKED ISSUE #1193: refactor cogent3.parse.genbank.Location to be PEP8 compliant
The constructor arguments should all be deprecated in favour of pep8 compliant versions. The general approach needs to be
```python
class Location:
    def __init__(
        self,
        data,
        Ambiguity=None,
        ... # all the other args
       **kwargs
      ):
      # loop over key/value items of kwargs and assign key to corresponding pep8 compliant arg
      # collect deprecation messages
       # if any messages, call deprecated() with a date 6 months from the day you write the code
```
As this is a deprecated function/argument, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function/argument will be removed) as 6 months from when you make the change.

### LINKED ISSUE #1301: refactor genbank coordinate handling objects
Genbank coordinate format is 1-based and our current implementation is fragile, expecting the user to know they have to adjust for that. [Here's an example](https://github.com/cogent3/cogent3/blob/120288a6b6e9f6c1b21f18750cced23a8ed0bbf1/src/cogent3/parse/genbank.py#L679).

I recommend we implement two properties `.start` and `.stop` on the relevant objects -- `Location` and `LocationList` (in `cogent3.parse.genbank`). In both these cases, the existing `.first()` and `.last()` methods should be deprecated with users informed about new attribute and behaviour. Those methods should translate to 0-based coordinates on reading from GenBank.

KatherineCaley:
DEP: deprecating arguments on `cogent3.parse.genbank.Location` to be PEP8 compliant, fixes #1193.
DEP: deprecating coordinate handling methods on `cogent3.parse.genbank.Location`, `first` and `last` are replaced with `start` and `stop`, the former is 1-based and the latter is 0-based, fixes #1301.

### DIFF
--- a/src/cogent3/parse/genbank.py
+++ b/src/cogent3/parse/genbank.py
@@ -427,7 +427,7 @@ class Location(object):
         except TypeError:
             return self._data[-1].stop
 
-    def first(self):
+    def first(self):  # pragma: no cover
         from cogent3.util.warning import deprecated
 
         deprecated(
@@ -440,7 +440,7 @@ class Location(object):
 
         return self.start
 
-    def last(self):
+    def last(self):  # pragma: no cover
         from cogent3.util.warning import deprecated
 
         deprecated(
@@ -456,50 +456,28 @@ class Location(object):
 class LocationList(list):
     """List of Location objects."""
 
-    BIGNUM = 1e300
-
-    def start(self):
+    def first(self):  # pragma: no cover
         """Returns first base of self."""
-        curr = self.BIGNUM
-        for i in self:
-            start = i.start
-            if curr > start:
-                curr = start
-        return curr
+        from cogent3.util.warning import discontinued
 
-    def stop(self):
-        """Returns last base of self."""
-        curr = 0
-        for i in self:
-            stop = i.stop
-            if stop > curr:
-                curr = stop
-        return curr
-
-    def first(self):
-        from cogent3.util.warning import deprecated
-
-        deprecated(
+        discontinued(
             "method",
             "LocationList.first",
-            "LocationList.start",
             "2023.06.15",
-            "Warning: '.start' is changed to reflect 0-based coordinated, previous implementation reflected 1-based coordinates",
         )
 
-        return self.start + 1
+        return min(self, key=lambda x: x.start) + 1
 
-    def last(self):
-        from cogent3.util.warning import deprecated
+    def last(self):  # pragma: no cover
+        """Returns last base of self."""
+        from cogent3.util.warning import discontinued
 
-        deprecated(
+        discontinued(
             "method",
             "LocationList.last",
-            "LocationList.stop",
             "2023.06.15",
-            "Warning: '.stop' is changed to reflect 0-based coordinated, previous implementation reflected 1-based coordinates",
         )
-        return self.stop + 1
+        return max(self, key=lambda x: x.stop) + 1
 
     @property
     def strand(self):
--- a/src/cogent3/parse/genbank.py
+++ b/src/cogent3/parse/genbank.py
@@ -307,7 +307,7 @@ class Location(object):
     is_between should be False (the default), or True.
     is_bounds should be False(the default, indicates range), or True.
     accession should be an accession, or None (default).
-    Db should be a database identifier, or None (default).
+    db should be a database identifier, or None (default).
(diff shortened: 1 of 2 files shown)


src/cogent3/parse/genbank.py
```diff
@@ -348,85 +378,132 @@ def __str__(self):
...
+        return self.stop()
+
+
+class LocationList(list):
+    """List of Location objects."""
 
     BIGNUM = 1e300
 
-    def first(self):
+    def start(self):
```
GavinHuttley, line 459: can you make this a property
```
@property
def start(self):
    ...
```
and change all usages to `<obj>.start`

src/cogent3/parse/genbank.py
```diff
@@ -348,85 +378,132 @@ def __str__(self):
...
-            first = i.first()
-            if curr > first:
-                curr = first
+            start = i.start()
+            if curr > start:
+                curr = start
         return curr
 
-    def last(self):
+    def stop(self):
```
GavinHuttley, line 468: can you make this a property
```
@property
def stop(self):
    ...
```
and change all usages to `<obj>.stop`

src/cogent3/parse/genbank.py
```diff
@@ -348,85 +378,132 @@ def __str__(self):
...
+        deprecated(
+            "method",
+            "LocationList.last",
+            "LocationList.stop",
+            "2023.06.15",
+            "Warning: '.stop()' is changed to reflect 0-based coordinated, previous implementation reflected 1-based coordinates",
+        )
+        return self.stop() + 1
+
     def strand(self):
```
GavinHuttley, line 502: can you change this to being a property too. It will still work as a method, but at least it can be just used as `<obj>.strand`

tests/test_parse/test_genbank.py
```diff
@@ -505,29 +505,37 @@ def test_init(self):
...
-        l5 = Location([l4, l3], Strand=-1)
+        l5 = Location([l4, l3], strand=-1)
         self.assertEqual(str(l5), "complement(37^42..(37.42))")
 
 
+def test_Location_start():
+    """the start and stop should reflect 0-based indexing (python style), not 1-based indexing (genbank style).
+    This means they should be 1 less than the data given"""
+    l = Location(37)
+    assert str(l.start()) == "36"
```
GavinHuttley, line 535: errr, why not just `l.start == 36`? Does it need to be cast to a string?

GavinHuttley (changes requested):
couple of small tweaks!

tests/test_parse/test_genbank.py
KatherineCaley: lol, I was just copying the above tests

src/cogent3/parse/genbank.py
```diff
@@ -348,85 +378,135 @@ def __str__(self):
...
+
+class LocationList(list):
+    """List of Location objects."""
 
     BIGNUM = 1e300
 
-    def first(self):
+    def start(self):
         """Returns first base of self."""
         curr = self.BIGNUM
```
GavinHuttley, line 463: These methods are rather old. The `start()` is just finding the minimum value, so the `BIGNUM` attribute can be removed entirely and the method rewritten as
```python
    def start(self):
        """Returns first base of self."""
        return min(self, key=lambda x: x.start)

    def stop(self):
        """Returns last base of self."""
        return max(self, key=lambda x: x.stop)
```

I note that the methods on `LocationList` don't seem to be tested, can you just add tests -- check that the original code works and inspect the output being generated under it. Make sure it makes sense!

Then change to using your new properties and updater the tests to handle the 0-based counting.

GavinHuttley (changes requested):
address missing tests and old (inefficient) algorithms

src/cogent3/parse/genbank.py
GavinHuttley: In looking over these again I can see the `first()` and `last()` methods on `LocationList` are never used, so they should be discontinued. No tests needed!

KatherineCaley: Easy!

PR #1326, opened 2023-03-24
DEV: added scriv into tooling and used it

GavinHuttley:


khiron:
Other than the white space issue it looks good.  I'll fix that directly and check in a change to the template and any pending scrivs.

khiron (commented):
It's fine.  I'll fix the template's white spacing issue and fix any outstanding scrivs and check in an update

PR #1340, opened 2023-03-28
Issue#1194: deprecated function TreeAlign for tree_align

khiron:
Internal uses of TreeAlign changed to tree_align
Tests of TreeAlign maintained until deprecation

### DIFF
--- a/src/cogent3/align/progressive.py
+++ b/src/cogent3/align/progressive.py
@@ -150,33 +150,7 @@ def tree_align(
 
 
 def TreeAlign(*args, **kwargs):  # pragma: no cover
-    """Returns a multiple sequence alignment and tree.
-
-    Parameters
-    ----------
-    model
-        a substitution model or the name of one, see available_models()
-    seqs
-        a sequence collection
-    indel_rate, indel_length
-        parameters for the progressive pair-HMM
-    ests_from_pairwise
-        if no tree provided and True, the median value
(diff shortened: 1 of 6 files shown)


src/cogent3/align/progressive.py
```diff
@@ -147,3 +147,43 @@ def TreeAlign(
     )
     align.info["align_params"] = param_vals
     return align, tree
+
+
+def TreeAlign(*args, **kwargs):  # pragma: no cover
+    """Returns a multiple sequence alignment and tree.
```
GavinHuttley, line 153: change the docstring to """deprecated, used tree_align()"""

src/cogent3/parse/genbank.py
```diff
@@ -4,7 +4,10 @@
 from cogent3.core.info import Info
 from cogent3.core.moltype import get_moltype
 from cogent3.parse.record import FieldWrapper
-from cogent3.parse.record_finder import DelimitedRecordFinder, LabeledRecordFinder
+from cogent3.parse.record_finder import (
```
GavinHuttley, line 7: this is a black / isort issue. As this change is also made by the linter PR, remove it from here otherwise it will be a conflict.

GavinHuttley (changes requested):
All tests should be changed to using `tree_align` as there's no change to functionality.

PR #1343, opened 2023-03-29
Implement deprecate decorator

### LINKED ISSUE #1324: DEV add a deprecate_arguments decorator
There is a pattern to deprecating function / method arguments.

- change the signature to the new argument name(s)
- add a `**kwargs` parameter to the signature
- import the deprecated helper function
- inside the function / method, loop over `**kwargs` for deprecated args, assigning value of deprecated arg  to new parameter and trigger deprecation warning

After the warning period is finished, the signature, import and loop all need to be deleted. This is tedious. It's also quite an ask for new contributors to do all this.

Replace this functionality with a `deprecate_arguments` decorator added to `cogent3.util.misc. warning`. This decorator will allow us to change the method / function signature to its final state and removal

This decorator will:
- accept the same arguments as the utility function, but also a tuple of `old_new` pairs (old name, new name)
- intercept all calls to the wrapped method
-  assigning value of deprecated arg  to new parameter and trigger deprecation warning
- call the wrapped method with the correct signature

khiron:
fixes #1324


tests/test_util/test_warning.py
```diff
@@ -0,0 +1,38 @@
...
+__copyright__ = "Copyright 2007-2022, The Cogent Project"
+__credits__ = ["Gavin Huttley", "Richard Morris"]
+__license__ = "BSD-3"
+__version__ = "2023.2.12a1"
+__maintainer__ = "Gavin Huttley"
+__email__ = "gavin.huttley@anu.edu.au"
+__status__ = "Production"
+
+
+class WarningTests(TestCase):
```
GavinHuttley, line 20: don't use `unittest.TestCase`, just make valinall `test_<something> functions which will be exercised by `pytest`.

tests/test_util/test_warning.py
```diff
@@ -0,0 +1,38 @@
...
+    """Tests of functions in warnings"""
+
+    def test_deprecate(self):
```
GavinHuttley, line 23: use the pytest `capsys` fixture to check the deprecation message is generated.

src/cogent3/util/warning.py
```diff
@@ -68,3 +69,46 @@ def discontinued(_type, name, version, reason=None, stack_level=3):
...
+    def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
+        def wrapper(*args: Any, **kwargs: Any) -> Callable[..., Any]:
+
+            deprecated(
+                _type="function",
+                old=func.__name__,
+                new=alternate.__name__ if alternate else None,
+                version=version,
+                reason=reason,
+            )
```
GavinHuttley, line 104: the `stacklevel` argument will need to be specified, otherwise it will just point to where the decorator is used, not where the user employed the method.

tests/test_util/test_warning.py
```diff
@@ -0,0 +1,38 @@
...
+        """test that deprecated decorator functions"""
+
+        def new_function():
+            return True
+
+        @deprecate(new_function, "2023.3", "test deprecation")
+        def old_function():
+            return False
+
+        self.assertTrue(new_function())
```
GavinHuttley, line 33: check that the stdout correctly references the actual invoking line (might need to use the `inspect` module to get the line number for the current line and then call the function that triggers the deprecation, you can check that line number is used. [This stackoverflow answer](https://stackoverflow.com/questions/24438976/debugging-get-filename-and-line-number-from-which-a-function-is-called) might be useful.

GavinHuttley (changes requested):

tests/test_util/test_warning.py
```diff
@@ -0,0 +1,38 @@
...
+    """Tests of functions in warnings"""
+
+    def test_deprecate(self):
+        """test that deprecated decorator functions"""
+
+        def new_function():
+            return True
+
+        @deprecate(new_function, "2023.3", "test deprecation")
+        def old_function():
```
GavinHuttley, line 30: add a docstring to this function, and add a test that checks the decorated function has the original docstring, e.g.
```python
@deprecate(new_function, "2023.3", "test deprecation")
def old_function():
    """dummy docstring"""
    ...
```
then test as
```python
def test_old_func():
    assert old_func.__doc_ == "dummy docstring"
```

GavinHuttley: You'll need `functools.wraps` for this.

src/cogent3/util/warning.py
```diff
@@ -68,3 +69,46 @@ def discontinued(_type, name, version, reason=None, stack_level=3):
...
+
+    The decorator will call the alternate function when the decorated function is called.
+    If an alternate function is not provided, or is not a callable function
+    then the original function will be called instead.
+    """
+
+    def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
+        def wrapper(*args: Any, **kwargs: Any) -> Callable[..., Any]:
+
+            deprecated(
```
GavinHuttley, line 98: if it's to decorate functions, it should also decorate methods and that's where `functools.wraps` won't work in terms of correctly preserving the method name and other metadata (docstring, annotations).

tests/test_util/test_warning.py
```diff
@@ -0,0 +1,38 @@
...
+            return False
+
+        self.assertTrue(new_function())
+        self.assertTrue(old_function())
+
+
```
GavinHuttley, line 36: add another test, check that you can pickle the decorated function
```python
def test_pickling():
    import pickle

    assert type(pickle.dumps(old_function) == bytes
```
that will fail if the function cannot be pickled.

src/cogent3/util/warning.py
khiron: I am using a stacklevel of 2 figuring that from the wrapper of the decorated function back to the caller of the original function should be 2 frames.  Pending testing.

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
-__author__ = "Richard Morris"
-__copyright__ = "Copyright 2007-2022, The Cogent Project"
-__credits__ = ["Gavin Huttley", "Richard Morris"]
-__license__ = "BSD-3"
-__version__ = "2023.2.12a1"
-__maintainer__ = "Gavin Huttley"
-__email__ = "gavin.huttley@anu.edu.au"
-__status__ = "Production"
 
+def test_deprecated_args():
```
GavinHuttley, line 10: you'll want to trap the deprecation warning in this test, otherwise there'll be too much guff displayed after the test run.

Since testing the correctness of returned value is the purpose here, I suggest using the [@pytest.mark.filterwarnings](https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#pytest-mark-filterwarnings) decorator.

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
+        reason="x and y are not descriptive",
+    )
+    def test_function(a: int, b: int) -> int:
+        return a + b
 
-        @deprecate(new_function, "2023.3", "test deprecation")
-        def old_function():
-            return False
+    with pytest.deprecated_call():
+        test_function(x=5, y=3)
```
GavinHuttley, line 36: convert these different argument combinations into a parameterised test, e.g.
```python
@pytest.mark.parametrize("kwargs", (dict(x=5, y=3), dict(a=5,y=3))
def test_deprecated_args_emits_warning(kwargs):
    # setting up changed() function
    with pytest.deprecated_call():
        changed(**kwargs)

File: tests/test_util/test_warning.py [hunk 10]
```diff
@@ -1,38 +1,88 @@
...
-        def new_function():
-            return True
+def test_deprecated_args_emits_warning():
+    # Example target function to be decorated
+    @deprecated_args(
+        [("x", "a"), ("y", "b")],
+        version="a future release",
+        reason="x and y are not descriptive",
+    )
+    def test_function(a: int, b: int) -> int:
```
GavinHuttley, line 32: while this function is local, the name is confusing because it has the word "test". Suggest calling it `changed()`

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
-if __name__ == "__main__":
-    main()
+
+def test_function_deprecated_args_maintains_docstring():
+    @deprecated_args(
+        [("x", "a"), ("y", "b")],
+        version="a future release",
+        reason="x and y are not descriptive",
+    )
+    def test_function(a: int, b: int) -> int:
```
GavinHuttley, line 54: function name as above

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
+        """This is a test function"""
+        return a + b
+
+    assert test_function.__doc__ == "This is a test function"
+
+
+def test_method_deprecated_args():
+    class test_class:
```
GavinHuttley, line 62: name issue as above for both the class and the method, just call the class `foo` and method `changed`

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
+            version="a future release",
+            reason="x and y are not descriptive",
+        )
+        def test_method(self, a: int, b: int) -> int:
+            return a + b
+
+    test_object = test_class()
+    expected = test_object.test_method(a=5, b=3)
+    got = test_object.test_method(x=5, y=3)
+    assert got == expected
```
GavinHuttley, line 74: add a check on the method docstring

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
+
+
+def test_method_deprecated_args():
+    class test_class:
+        @deprecated_args(
+            [("x", "a"), ("y", "b")],
+            version="a future release",
+            reason="x and y are not descriptive",
+        )
+        def test_method(self, a: int, b: int) -> int:
```
GavinHuttley, line 68: add a docstring to the method for testing method docstrings also preserved

tests/test_util/test_warning.py
```diff
@@ -1,38 +1,88 @@
...
+        [("x", "a"), ("y", "b")],
+        version="a future release",
+        reason="x and y are not descriptive",
+    )
+    def test_function(a: int, b: int) -> int:
+        return a + b
+
+    myfunc = test_function(x=1, y=2)
+    pickled_func = pickle.dumps(myfunc)
+    assert type(pickled_func) == bytes
```
GavinHuttley, line 88: add a test for the class with decorated method too

GavinHuttley (changes requested):
edits suggested to reduce verbosity of tests and improve readability

khiron:
@deprecated decorator replaced with deprecated_args decorator which fixes #1343

tests/test_util/test_warning.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        def changed(self, a: int, b: int) -> int:
+            """This is a test function"""
+            return a + b
+
+    assert foo().changed.__doc__ == "This is a test function"
+
+
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
+def test_function_deprecated_args_pickled():
```
GavinHuttley, line 135: 3 copies of this same test name

GavinHuttley (changes requested):
need to fix text over-rides

tests/test_util/test_warning.py
khiron: I think I addressed the text over-rides comment

Issue #1364, opened 2023-04-16
DEV: add is_discontinued argument to deprecated_args decorator


GavinHuttley:
See the `cogent3.util.warning.deprecated_callable` decorator. It's modelled after your `deprecated_args` but includes this argument.

The use case is when we have arguments that are no longer going to be supported.

PR #1369, opened 2023-04-17
added is_discontinued to deprecated_args decorater

khiron:
fixes issue #1364

### DIFF
--- a/src/cogent3/util/warning.py
+++ b/src/cogent3/util/warning.py
@@ -1,7 +1,7 @@
 import functools
 import inspect
 
-from typing import Any, Callable, List, Optional, Tuple
+from typing import Any, Callable, List, Optional, Tuple, Union
 from warnings import catch_warnings, simplefilter
 from warnings import warn as _warn
 
@@ -70,10 +70,9 @@ def discontinued(_type, old, version, reason=None, stack_level=3):
 
 
 def deprecated_args(
-    mapping: List[Tuple[str, str]],
     version: str,
     reason: str,
-    is_discontinued: bool = False,
+    arguments: List[Union[Tuple[str, str], str]] = None,
 ) -> Callable[..., Any]:
     """
     A decorator that marks specific arguments of a function as deprecated.
@@ -84,15 +83,12 @@ def deprecated_args(
 
     Parameters
     ----------
-    mapping : List[Tuple[str, str]]
-        A list of 2-tuples specifying the mapping of old argument names to new argument names.
-        if the argument is being discontinued then the new argument name should be None
     version : str
         The version when the old arguments will be removed in calver format, e.g. 'YYYY.MM'
     reason : str
         Reason for deprecation or guidance on what to do
-    is_discontinued : bool
-        If True the callable is being discontinued.
+    mapping : List[Union[Tuple[str, str],str]]
+        A list of discontinued old argument names, or deprecated old and replacement new argument names.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


khiron:
This addresses the required functionality but I'm considering a change to the UX.  

Mapping:  list[Tuple[str,str]] in the case of is_discontinued = True the last item in the tuple is ignored.  Perhaps a better UX is for mapping to be of type List[Union[Tuple[str, str], str]] where a list item that is just a string is implied to be a discontinued argument.

eg: @deprecated_args([('x','a'),'y'],version="a future release", reason="a replaces x, y is discontinued")

src/cogent3/util/warning.py
```diff
@@ -105,7 +109,7 @@ def deprecated_args(
     Here's an example of how to use the `deprecated_args` decorator to mark the argument `old_name` as deprecated
     and replace it with the new name `new_name`.
 
-    >>> @deprecated_args(mapping=[('old_name', 'new_name')], version='2.0', reason='Use new_name instead')
+    >>> @deprecated_args(mapping=[('old_name', 'new_name')], version='2.0', reason='Use new_name instead', is_discontinues=True)
```
GavinHuttley, line 112: typo, should be `is_discontinued`

GavinHuttley (changes requested):
I think you're correct in pointing out we may have a mix of conditions, some arguments to be renamed, others to be discontinued.
I suggest the following, that `mapping` is now optional , i.e.
```python
mapping: Optional[List[Tuple[str, str]]]=None,
```
and `is_discontinued` is a name or series of names of arguments to discontinue
```python
is_discontinued: Optional[Union[str,List[str],Tuple[str]]]=None,
```
(may not have the type hint correct!! I'm basically saying allow the dev to just provide a single string or a series of strings as a list or tuple)

src/cogent3/util/warning.py
khiron: OK will make those changes and also clean up testing (similar to the way you did your tests).  I'm going to extract out a function to nominate a func ('method' or 'function'  + name) as I should use the same convention for args.

khiron: I decided to call the utility functions directly rather than messing with custom params.  Pretty happy with the UX from a developer POV right now.  

Check out the chained decorators test.

khiron:
deprecated_callable doesn't redirect the call to the new function.  

We could check if a new function is callable in scope and call that, allowing us to stub out the rest of the old function in the late stage process of the deprecation.

tests/test_util/test_warning.py
```diff
@@ -224,3 +224,22 @@ def test_method_deprecated_method_pickling(recwarn):
 def test_deprecated_callable_resolves_type(recwarn, func, _type):
     func(2)
     assert _type in recwarn.list[0].message.args[0]
+
+
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
+def test_function_deprecated_args_deprecated_callable_chained_decorators(recwarn):
+    @deprecated_args("2023.6", "x is not descriptive", [("x", "a")])
+    @deprecated_args("2023.6", "y is not descriptive", [("y", "b")])
+    @deprecated_callable(
```
GavinHuttley, line 233: This does not match the signature of `deprecated_callable`. The old function name is deduced from what is decorated and I think while it's possible to overload the usage of one variable to be either the new name or, in it's absence, implying it is being discontinued. What if the developer forgets to add the new name, but `is_discontinued==False` an exception should be raised (it's not currently, but your approach here makes me realise it should be).

I think the approach to using one argument in `deprecated_args()` as `old_to_new_or_discontinued` relies on implicit logic because one could have a mix of types on the same iterable (e.g. `[('a', 'x'),'drop_this',('b', 'y')]`. It can be done, but to me this is less clear than
```python
@deprecated_callable(old_new=[('a', 'x'), `('b', 'y')], discontinued='drop_this')`
```
So, parroting the zen ...explicit is better than implicit.

GavinHuttley (changes requested):
> deprecated_callable doesn't redirect the call to the new function.
> 
> We could check if a new function is callable in scope and call that, allowing us to stub out the rest of the old function in the late stage process of the deprecation.

There's no guarantee the new function is in `cogent3` and thus that it's even installed. Nor is there a guarantee the replacement function has the same API. (Changing a function provides that opportunity, we support the old until the indicated version but it's up to the user to read the specs on the suggested replacement and adapt their code themselves.)

So we won't do this. We won't even check if it's a valid reference for the same reason so it's possible there's a typo but them's the breaks!

Being explicit, replace the parameter argument with two.

tests/test_util/test_warning.py
khiron: That was a lot more fiddly than it should have been.  But it is now implemented as you expect.  It's now so explict that even duck typing can't be used to discriminate between old_new and discontinued.  Argument keywords must be used if discontinued is present in the arguments.

khiron:
Note Codacy appears to flag as a critical problem "No value for argument 'a' in function call" when calling changed(a,b) that has been decorated with deprecated_args(old_new('x','a')) using changed(x,b).

I am going to ignore the report in the tests as that is intentional/expected.  And we should not ignore the report if it pops up elsewhere.

PR #1392, opened 2023-04-26
ENH: updated relative and absolute position functions

KatherineCaley:
[CHANGED] absolute_index now called absolute_position, behaviour is pretty much maintained

[CHANGED] relative_index now called relative_position. No longer accepts negative indexing. No longer throws indexError for index corresponding to "stepped over" element. No longer returns a value that supports python index, as negative return value refers to positions prior to the start of the view.


KatherineCaley:
still todo, converts spans using relative index, when relative_index(x) < 0 or  relative_index(x) > len(self) then creating lostSpan()

PR #1424, opened 2023-05-09
ENH: sketch implementation of SequenceCollection.distance_matrix()

KatherineCaley:


src/cogent3/core/alignment.py
```diff
@@ -2285,6 +2298,103 @@ def get_features(
...
+        k : int, optional
+            The length of kmers to use in the distance calculation, by default 7.
+
+        Returns
+        -------
+        DistanceMatrix
+            Pairwise distances between sequences in the collection
+        """
+
+        from cogent3.evolve.fast_distance import DistanceMatrix
```
GavinHuttley, line 2320: I'm recommending we move distance related calculations into `cogent3.app.dist` to satisfy the "separation of concerns" principle, in terms of module organisation. In which case, this method just creates instances of the required apps using `cogent3.get_app()` (`cogent3` already imported above).

src/cogent3/core/alignment.py
```diff
@@ -2285,6 +2298,103 @@ def get_features(
                 # passing self only used when self is an Alignment
                 yield seq.make_feature(feature, self)
 
+    def distance_matrix(self, calc="jaccard", k: int = 7):
```
GavinHuttley, line 2301: We should not include `k` as an argument since the coefficients are derived from a specific value of `k`. The docstring needs to declare this. We have to hard-code the value of k that was used to obtain the coefficients.

`calc` is either `pdist` or `jc69` where the doctsring makes it clear the latter is a transformation of the former.
We should also raise a `TypeError` if moltype is not either DNA or RNA.

src/cogent3/core/alignment.py
```diff
@@ -2285,6 +2298,103 @@ def get_features(
...
+            for j in range(i):
+                name1, name2 = seq_names[i], seq_names[j]
+                dist = metric(kmers[name1], kmers[name2])
+                dists_dict[(name1, name2)] = dist
+                dists_dict[(name2, name1)] = dist
+
+        return DistanceMatrix(dists_dict)
+
+
+def _jaccard_transform(kmers_1: set, kmers_2: set):
```
GavinHuttley, line 2349: see my comment about splitting the computations,

GavinHuttley (changes requested):
I think we want to move all these calculations along with the polynomial coefficients into `cogent3.app.dist`. There should be 3 apps

1. `jaccard_dist()` app takes the `UnalignedSeqsType` (`SequenceCollection`) and returns a `PairwiseDistanceType` (`DistanceMatrix`) of jaccard distances. It should allow specifying the value of `k` 
2. `approx_pdist()` takes a `PairwiseDistanceType` of jaccard distances and returns a `PairwiseDistanceType` of predicted p-distances
3. `approx_jc69()` takes a `PairwiseDistanceType` of predicted p-distances distances and returns a `PairwiseDistanceType` of JC69 distances.

These apps can then be created and used within the `SequenceCollection.distance_matrix()` method.

src/cogent3/app/dist.py
```diff
@@ -125,3 +152,132 @@ def main(
...
+    -------
+    DistanceMatrix
+        returns pairwise Jaccard distance between sequences in the collection.
+    """
+
+    kmers = {name: set(seq.get_kmers(k)) for name, seq in seq_coll.named_seqs.items()}
+    seq_names = sorted(kmers.keys())
+    num_seqs = len(seq_names)
+
+    jaccard_dict = {}
```
GavinHuttley, line 179: Don't use Hungarian notation for variable names. Just call it `dists`

src/cogent3/app/dist.py
```diff
@@ -125,3 +152,132 @@ def main(
...
+    ----------
+    j_dists : DistanceMatrix
+    The pairwise Jaccard distance matrix
+
+    Returns
+    -------
+    DistanceMatrix
+    The pairwise approximate PDist matrix
+    """
+    j_dists_array = j_dists.array
```
GavinHuttley, line 209: avoid Hungarian notation, just `j_dists`.

src/cogent3/app/dist.py
```diff
@@ -125,3 +152,132 @@ def main(
...
+
+    # Initialise an array of the same size as j_dists_array with all values = 0.0
+    p_dists_array = zeros_like(j_dists_array)
```
GavinHuttley, line 212: just `p_dists`

src/cogent3/app/dist.py
```diff
@@ -125,3 +152,132 @@ def main(
...
+    # Convert only the upper indices from Jaccard distance to approximate PDist
+    upper_vals = polyval(JACCARD_PDIST_POLY_COEFFS, j_dists_array[upper_indices])
+    p_dists_array[upper_indices] = upper_vals
+
+    # Reflect the upper triangle to the lower triangle
+    lower_indices = (upper_indices[1], upper_indices[0])
+    p_dists_array[lower_indices] = upper_vals
+
+    # Set diagonal to 0.0
+    fill_diagonal(p_dists_array, 0.0)
```
GavinHuttley, line 228: isn't this already a zero array?

GavinHuttley (changes requested):
A small number of other requested changes, mainly about Hungarian notation.

Plus you need to test all the apps plus their integration. So test the jaccard app works as expected for simple cases, etc...

tests/test_app/test_dist.py
```diff
@@ -355,8 +450,8 @@ def test_approx_jc69_vals(_seqs1_collection):
     pdist_app = approx_pdist()
     pdists = pdist_app(jdists)
 
-    jc_app = approx_jc69()
-    jc_dists = jc_app(pdists)
+    jc_app = jaccard_dist(k=10) + approx_pdist() + approx_jc69()
+    jc_dists = jc_app(seqs)
 
     for i, j in itertools.combinations(range(len(names)), 2):
         seq1, seq2 = names[i], names[j]
```
GavinHuttley, line 457: need to add tests that NotImplementedError is raised when seqs have wrong moltype, not raised when they have RNA or DNA moltype
```python
@pytest.mark.parametrize("moltype", ("protein", "text", "bytes")
def test_fails(moltype):
    data = [("s1", "ACGTA"),
    ("s2", "ACGTA")]
    seqs = make_unaligned_seqs(data=data, moltype=moltype)
    with pytest.raised(NotImplementedError):
        seqs.distance_matrix()
```

add a test mirroring the above that does not fail when dna or rna are the moltypes.

GavinHuttley (changes requested):
needs two more tests

tests/test_app/test_dist.py
KatherineCaley: Is the second test you're suggesting not already tested for considering all other tests use DNA or RNA moltype?

KatherineCaley: no problems making it explicit though!

GavinHuttley: Your correct for DNA! Perhaps make one of those other tests use the rna moltype?

KatherineCaley: testing with RNA is already the case for the following:

@pytest.mark.parametrize("moltype", ("dna", "rna"))
 def test_collection_distance_matrix_same_seq(moltype):

but will add to other tests too!

PR #1463, opened 2023-06-02
MAINT: changes addressing speed of operations

KatherineCaley:


tests/test_core/test_annotation_db.py
```diff
@@ -457,6 +457,67 @@ def test_matching_conditions():
     assert got == expect
 
 
+def test_matching_conditions_IN():
+    got, cond = _matching_conditions({"biotype": ("CDS", "mRNA", "exon")}, allow_partial=True)
+    expect = "biotype IN (?,?,?)"
```
GavinHuttley, line 462: add a check for the value of `cond` too

GavinHuttley (changes requested):
very nice! tiny change, codacy identified an issue with one test (you didn't use the `cond` variable. I think it's good to test that value too.

GavinHuttley (approved):

PR #1470, opened 2023-06-06
MAINT: added method to merge two dbs 

KatherineCaley:


src/cogent3/core/alignment.py
```diff
@@ -914,8 +914,8 @@ def take_seqs(self, seqs, negate=False, **kwargs):
         result = self.__class__(result, names=seqs, info=self.info, **kwargs)
         if self.annotation_db:
             result.annotation_db = type(self.annotation_db)()
-            result.annotation_db.update(
-                annot_db=self.annotation_db, seqids=result.names
+            result.annotation_db._update_db_from_other_db(
+                other_db=self.annotation_db, seqids=result.names
```
GavinHuttley, line 918: Never use a private method in another class. Change the implementation of `annotation_db.update()` to using the new method instead of `_update_from_rich_dict()`

src/cogent3/core/annotation_db.py
```diff
@@ -988,6 +988,36 @@ def _update_db_from_rich_dict(self, data: dict, seqids: OptionalStr = None):
...
+        self, other_db: SupportsFeatures, seqids: OptionalStr = None
+    ):
+        if isinstance(seqids, str):
+            seqids = {seqids}
+        elif seqids is not None:
+            seqids = set(seqids) - {None}  # Make sure None is not part of this!
+
+        table_names = other_db.table_names
+
+        with other_db.db:
```
GavinHuttley, line 1001: Since this is a method on `SqliteAnnotationDbMixin`, using the current methods should greatly simplify this. Even though you're using private methods here, they're on instances of the same class.
```python
for table_name in other_db.table_names:
    sql, vals = _select_records_sql(table_name, conditions)
    for record in other_db._execute_sql(sql, vals):
        sql, vals = _add_record_sql(table_name, record)
        self.._execute_sql(sql, vals)
```
There may be a way to do a bulk insert operation in SQLite, rather than looping.

GavinHuttley (changes requested):
SOLID principles at play here! Plus I think there may be a command for doing a bulk insert which would be more performant

src/cogent3/core/annotation_db.py
GavinHuttley: Following on from the above, I think actually the whole thing can be much simpler. I found [this stackoverflow](https://stackoverflow.com/questions/18219779/bulk-insert-huge-data-into-sqlite-using-python#32239587) post about bulk inserts. In short, they use `<connection instance>.executemany(<sql>, <records>)`. As I'm not sure the order of columns will always the same I think we need to figure out the column order, so something like this.
```python
col_order = {tname: r["name"] for tname in other_db.table_names for r in other_db.execute(f"pragma table_info({tname!r})").fetchall()
```
Then, we iterate over the tables and get all the records
```python
for tname in other_db.table_names:
    sql = f"SELECT * FROM {tname}" # that will come back in the defined column order
    val_placeholder = ",".join("?" * len(col_order[tname]))
    data = list(other_db. _execute_sql(sql))
    # now make the insert sql statement
    sql = f"INSERT INTO {tname}({','.join(col_order[tname])}) ({val_placeholder})"
    self.db.executemany(sql, data)
```
Likely some errors in this, but it's a better way to do it than I suggested before.

GavinHuttley: Just goes to show I should read your code more carefully!

GavinHuttley (approved):
I should really have read your code more carefully, I would have learned something! Nice job :^)

PR #1473, opened 2023-06-07
MAINT: clean up _update_db_from_other_db code

KatherineCaley:


src/cogent3/core/annotation_db.py
```diff
@@ -998,25 +998,23 @@ def _update_db_from_other_db(
...
+                row[1]
+                for row in other_db.db.execute(f"PRAGMA table_info({tname})").fetchall()
+            ]
+            for tname in table_names
+        }
 
-                    self.db.commit()  # not sure if this is needed...
+        for tname in other_db.table_names:
+            sql, vals = _select_records_sql(
+                table_name=tname, conditions={"seqid": seqids}
```
GavinHuttley, line 1011: Should we just assign the default value for `conditions` if there are no `seqids` passed in?  That is, before the loop
```python
conditions = {"seqid": seqids} if seqids else {}
```

GavinHuttley (changes requested):
A question, the statement
```python
seqids = None if len(set(seqids) - {None}) == 0 else set(seqids) - {None}
```
is assigning a value to `seqids` that's not consistent with the declared types. Is this line being hit by a test? I make a suggestion in the file about what that might be changed to.=

src/cogent3/core/annotation_db.py
KatherineCaley: @GavinHuttley The funky handling of `seqid` input is to handle the edge case where `seqids=[None]`, because SQL treats None as a seqid. Not sure whether this is necessary to account for.

KatherineCaley: I also think the declared type should be `OptionalStrList = typing.Optional[typing.Union[str, typing.List[str]]]`, or even `OptionalStrContainer = typing.Optional[typing.Union[str, typing.Sequence]]`

GavinHuttley (approved):
Nice!

PR #1483, opened 2023-06-22
progressive_align app uses approximate distance to build tree when no tree is provided 

KatherineCaley:


KatherineCaley:
Converting to draft because the build fails on windows

KatherineCaley:
Re-ran test and now passing 🧐

GavinHuttley (approved):
This looks good !

PR #1489, opened 2023-06-27
MAINT: better NotCompleted result from sp_score

GavinHuttley:
[CHANGED] The alignment quality sp_score relies on computing pairwise
     genetic distanxces. This can fail due to NaN's in those distances.
     Return a more informative NotCompleted object.


KatherineCaley (approved):
Looks good!

PR #1505, opened 2023-07-05
TST: add test for make_dna_scoring_dict()

KatherineCaley:


GavinHuttley (changes requested):
add a transition != transversion check

KatherineCaley:
> add a transition != transversion check

I pushed one to my fork yesterday, but is hasn't been automatically be added to the PR? 

@GavinHuttley  I could cancel and reopen the PR?

GavinHuttley:
> > add a transition != transversion check
> 
> I pushed one to my fork yesterday, but is hasn't been automatically be added to the PR?
> 
> @GavinHuttley I could cancel and reopen the PR?

that is super odd. Yes to close and re-open. 

Actually, here's a better way than checking ts != tv. You explicitly define the values of match, ts and tv in constructing the scoring matrix. Just add those specific values to each tested condition.

KatherineCaley:
> 
Done. Also pushing up that change appears to have bundled up the rest of them.

src/cogent3/app/align.py
```diff
@@ -553,6 +559,65 @@ def main(self, seqs: UnalignedSeqsType) -> T:
...
+
+    Notes
+    -----
+    It records the score of the alignment (in addition to all other the
+    parameter values) as "sw_score" in ``alignment.info['align_params']``.
+    """
+
+    def __init__(
+        self,
+        score_matrix=None,
```
GavinHuttley, line 574: add type hints on the arguments

src/cogent3/app/align.py
```diff
@@ -553,6 +559,65 @@ def main(self, seqs: UnalignedSeqsType) -> T:
...
+        moltype
+            molecular type of sequences
+        """
+        moltype = get_moltype(moltype)
+        self.score_matrix = score_matrix or (
+            make_dna_scoring_dict(10, -1, -8)
+            if moltype.label == "dna"
+            else make_generic_scoring_dict(10, moltype)
+        )
+        self.insertion_penalty = insertion_penalty
```
GavinHuttley, line 598: make all the class attributes semi-private, to discourage the user from messing with them on an instance. The reason this matters is the serialisation of the instance uses `self._init_vals` which is defined on build, so any modification by the user makes the state inconsistent.

src/cogent3/app/align.py
```diff
@@ -553,6 +559,65 @@ def main(self, seqs: UnalignedSeqsType) -> T:
...
+        self.extension_penalty = extension_penalty
+
+    def main(self, seq1: SeqType, seq2: SeqType) -> AlignedSeqsType:
```
GavinHuttley, line 601: this won't work with the standard app approach, a single argument is required. Change to
```python
def main(self, seqs: UnalignedSeqsType) -> AlignedSeqsType:
      if seqs.num_seqs > 2:
        return NotCompleted("ERROR", ..., message="maximum number of two seqs per collection", ...)
     seq1, seq2 = seqs.seqs
   ```

File: tests/test_app/test_align.py [hunk 4]
```diff
@@ -634,3 +639,48 @@ def test_progressive_align_iters(seqs):
     assert isinstance(aln, ArrayAlignment)
     assert len(aln) == 42
     assert aln.moltype == aligner._moltype
+
+
+def test_smith_waterman_matches_local_pairwise(seqs):
+    aligner = smith_waterman()
+    got = aligner(seqs.get_seq("Human"), seqs.get_seq("Bandicoot"))
```
GavinHuttley, line 646: change test to using a sequence collection. Add a new test for calling with a sequence collection that does not have 2 sequences.

GavinHuttley (changes requested):
In reviewing this, I realise we need better documentation about the assumptions being made by the built-in apps!
So the major change (which is pretty small) is related to that.

tests/test_app/test_align.py
```diff
@@ -634,3 +639,68 @@ def test_progressive_align_iters(seqs):
     assert isinstance(aln, ArrayAlignment)
     assert len(aln) == 42
     assert aln.moltype == aligner._moltype
+
+
+def test_smith_waterman_matches_local_pairwise(seqs):
+    aligner = smith_waterman()
+    coll = make_unaligned_seqs(data=[seqs.get_seq("Human"), seqs.get_seq("Bandicoot")])
+    got = aligner(coll)
+    print(got)
```
GavinHuttley, line 648: delete print statement

tests/test_app/test_align.py
```diff
@@ -634,3 +639,68 @@ def test_progressive_align_iters(seqs):
...
+        insertion,
+        extension,
+        return_score=True,
+    )
+    assert got == expect
+
+
+def test_smith_waterman_generic_moltype():
+    """tests when the moltype is generic"""
+    test_moltypes = ["text", "rna", "protein", "protein_with_stop", "bytes", "ab"]
```
GavinHuttley, line 686: @pytest.mark.paramtrize!!

GavinHuttley (approved):
Two small changes, which I'll just do. But adding to this review so you see them. Thanks for finishing!

PR #1530, opened 2023-07-11
Add test for querying for non-existent parent/child

KatherineCaley:


src/cogent3/core/annotation_db.py
```diff
@@ -691,20 +691,21 @@ def get_feature_parent(
...
-                            column="name",
-                            name=name,
-                        )
-                    ):
-                        parent = parent[0]
-                        parent.pop(
-                            "parent_id"
-                        )  # remove invalid field for the FeatureDataType
-                        yield parent
+                if result["parent_id"]:
```
GavinHuttley, line 694: I suggest a slight change to make the code easier to read
```python
if not result["parentid"]:
    return

# de-indent the remaining code
```

GavinHuttley (changes requested):
nice!

1. Suggesting a slight change to code to make it less indented (and thus more readable).
2. In your tests, use an ID that's very clearly not in the db, like `"this id does not exist"`. Makes it clear the ID cannot exist, whereas how do I know the ones you created don't exist, maybe they do and the test passes because of a different error.

KatherineCaley:
> 2. In your tests, use an ID that's very clearly not in the db, like `"this id does not exist"`. Makes it clear the ID cannot exist, whereas how do I know the ones you created don't exist, maybe they do and the test passes because of a different error.

@GavinHuttley I chose IDs that do exist in the file! They just do not have a parent or a child. I think using an ID that does not exist in the file would be a different test.

GavinHuttley (approved):
Noice!

PR #1553, opened 2023-07-17
Update scriv to work with version number being moved to src/_version.py 

khiron:
Moved the location that scriv looks for __version__ from `src/__init__.py` to `src/_version.py` by updating scriv config in `pyproject.toml`

I've tested this locally.

Update `_version.py` with a new version number (after "2023.2.12a1"), then in the repo root invoke  ...

```sh 
scriv collect 
```

... commit changes, and push

Also changed 2 references in changelogs from "khiron" to "Richard Morris"


GavinHuttley (commented):
bonza!

PR #1562, opened 2023-08-03
MAINT: smith_waterman works when seqs have no moltype

KatherineCaley:


src/cogent3/app/align.py
```diff
@@ -618,6 +625,7 @@ def main(self, seqs: UnalignedSeqsType) -> AlignedSeqsType:
             extension_penalty=self._extension_penalty,
             sw_score=score,
         )
+        aln.moltype = self.moltype
```
GavinHuttley, line 628: change this to `aln.to_moltype(self.moltype)`

GavinHuttley (changes requested):
Looking good, just one change.

Alignments are immutable, so use the method, not direct assignment, to set the moltype.

GavinHuttley (approved):

Issue #1466, opened 2023-06-05
deprecate cogent3.util.table.tolist()


KatherineCaley:
(1) copy contents of `tolist` into new method `to_list`, 

(2) decorate old method with:

`@deprecated_callable(version='2023.9', reason='PEP8', new='to_list')`

(3)  replace `tolist` contents with calling the new function 

(4) remove internal calls of `tolist` with `to_list`

fredjaya:
I will work on this!

GavinHuttley:
Awesome! Please read out [contributor guidelines](https://github.com/cogent3/c3dev/wiki). Feel free to post any questions here.

fredjaya:
Hi all, I have made several commits in my forked repo but struggling to push them. This is part of the output when pushing: 
```
nox > Session testdocs-3.10 was successful.
nox > Running session testdocs-3.11
nox > Session testdocs-3.11 skipped: Python interpreter 3.11 not found.
nox > Ran multiple sessions:
nox > * test_slow-3.8: skipped
nox > * test_slow-3.9: failed
nox > * test_slow-3.10: success
nox > * test_slow-3.11: skipped
nox > * test-3.8: skipped
nox > * test-3.9: failed
nox > * test-3.10: success
nox > * test-3.11: skipped
nox > * testmpi-3.8: skipped
nox > * testmpi-3.9: failed
nox > * testmpi-3.10: failed
nox > * testmpi-3.11: skipped
nox > * testdocs-3.8: skipped
nox > * testdocs-3.9: failed
nox > * testdocs-3.10: success
nox > * testdocs-3.11: skipped
error: failed to push some refs to 'github.com:fredjaya/cogent3.git'
```

Is the git error related to the several tests failing? Or an unrelated git issue (which I've tried troubleshooting)?

In case it's useful, this is the output when running `nox -s test-3.10`:
```
========================================= short test summary info ==========================================
SKIPPED [1] test_app/test_app_mpi.py:14: Not using MPI
XFAIL test_align/test_align.py::test_tree_align_pwise_iter - fails on linux due to no effect of iters
XPASS test_app/test_composable.py::test_as_completed_progress[True] passes except when run in full test suit
e
XPASS test_app/test_composable.py::test_as_completed_progress[False] passes except when run in full test sui
te
====== 9461 passed, 1 skipped, 1 deselected, 1 xfailed, 2 xpassed, 566 warnings in 128.66s (0:02:08) =======
nox > Session test-3.10 was successful.
[...]

PR #1597, opened 2023-10-05
DEP: Deprecate cogent3.util.table.tolist()

fredjaya:
Addresses #1466


src/cogent3/util/table.py
```diff
@@ -1940,6 +1941,14 @@ def tolist(self, columns=None):
 
         return result
 
+    @c3warn.deprecated_callable(version="2023.9", reason="PEP8", new="to_list")
+    def tolist(self, **kwargs):  # pragma: no cover
+        """deprecated, use to_list"""
+
+        c3warn.deprecated("function", "tolist", "to_list", "2023.9")
```
GavinHuttley, line 1948: Delete this line as it's not required. The decorator is already doing this.

src/cogent3/util/table.py
```diff
@@ -1940,6 +1941,14 @@ def tolist(self, columns=None):
 
         return result
 
+    @c3warn.deprecated_callable(version="2023.9", reason="PEP8", new="to_list")
```
GavinHuttley, line 1944: The version is the released version by which this will be removed. Change it to 3 months from now, so the version when `Table.tolist()` will disappear is "2024.1"

tests/test_evolve/test_models.py
GavinHuttley, line 1: When you see a whole file being affected and you know you've only made a few changes, something has gone wrong. One likely candidate is an issue of a "linter", i.e. black was run differently on this file, or the file endings. Rerun the project pinned version of black (listed in pyproject.toml) on the file and see if that changes this.

GavinHuttley (changes requested):
Looking good! There's detailed comments on individual lines.

One additional remark is that the Codacy static checker "failed". Always click through and look at these. The real benefit is telling your edit means one import is no longer required, or you have an unused variable etc... But, if they don't appear to be things resulting from your changes, then you can make a comment in the PR about that. Codacy does sometimes only flag issues when a file containing it is changed, and sometimes the flagged issues are incorrect. If in doubt, just say it and we'll check it for you.

tests/test_evolve/test_models.py
GavinHuttley, line 1: still puzzled why this file has so many changes showing

GavinHuttley (changes requested):
any thoughts on what happened to `test_models.py`?

tests/test_evolve/test_models.py
fredjaya: no clue... looking into it now. The black version seems to be the same as the project pinned version. I'm going to edit that file from the last commit where it worked and see how it goes.

GavinHuttley: May be just replace your version directly with the original (download that file from main repo directly and copy on top of yours) and rerun the tests. If that works, commit, push and done!

GavinHuttley (approved):
Thanks for the contribution @fredjaya !

PR #1622, opened 2023-11-15
bug fixes for dotplot and related code, fixes #1619

### LINKED ISSUE #1619: dotplot doesn't use argument for kmer size, defualt k is buggy
`_SequenceCollectionBase.dotplot()`does not pass on the argument for kmer size to the resulting `Dotplot`. This is a simple fix. `Dotplot` has a `k` argument that should receive on the k from `_SequenceCollectionBase.dotplot()` 

As `Dotplot` does not propagate any user-defined k, it will always use `cogent3.align.pycompare._calc_seed_size()` to calculate the k size, and the logic in this function is dubious. 

```python3
def _calc_seed_size(w: int, t: int) -> int:
    """computes k-mer size

    Parameters
    ----------
    w : int
        window size
    t : int
        threshold for minimum number of matches

    Returns
    -------
    int
        returns the larger of (w // k) * k <= t OR w - t + 1
    """
    assert 0 < t <= w, f"threshold={t} > window size={w}"

    # k should be such that (w % k) * k < threshold
    denom = w - t + 1
    return max(min(w // denom, t), denom)
```

`    # k should be such that (w % k) * k < threshold` This is not maintained!! 

In fact, the smaller the proportion of `t/w`, the larger the returned `k` is!!

```python3
  _calc_seed_size(20, 15)
  # 6
```

```python3
  _calc_seed_size(20, 10)
  # 11
```

```python3
  _calc_seed_size(20, 5)
  # 16
```

In the latter cases, `k` is larger than the threshold!

KatherineCaley:
fixes #1619

### DIFF
--- a/src/cogent3/align/pycompare.py
+++ b/src/cogent3/align/pycompare.py
@@ -488,7 +488,7 @@ class MatchedSeqPaths:
         return trace
 
 
-def _calc_seed_size(w: int, t: int) -> int:
+def _calc_seed_size(w: int, t: int, min_val: int = 5) -> int:
     """computes k-mer size
 
     Parameters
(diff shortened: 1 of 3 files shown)


GavinHuttley (approved):
Looks good!

Issue #1633, opened 2023-11-21
`available_apps()` should take string to filter on


KatherineCaley:
Running `available_apps()` will only show you the top 5 and bottom 5 rows from 54 rows x 6 columns table! This means that you can't see the majority of apps. 

Running `available_apps()` should take a string argument that returns a filtered table with apps that match the query. 

i.e., `available_apps("load")` will return all apps with the string `"load"` in their name 

The documentation on how should be included when you print `available_apps()`. Perhaps also documentation on how to change the documentation to show the whole table?

PR #1637, opened 2023-11-21
Filter available_apps by name

cosmastech:
To resolve https://github.com/cogent3/cogent3/issues/1633

- [x] Confirm that test case passes

### DIFF
--- a/tests/test_app/test_init.py
+++ b/tests/test_app/test_init.py
@@ -69,6 +69,15 @@ class TestAvailableApps(TestCase):
         apps = available_apps()
         self.assertIsInstance(apps, Table)
         self.assertTrue(apps.shape[0] > 10)
+    
+    def test_available_apps_filter(self):
+        """availabe_apps can be filtered by name"""
+        filter: str = "sql"
+
+        apps = available_apps(filter)
+
+        self.assertIsInstance(apps, Table)
+        self.assertTrue(len(apps), 1)
 
     def test_composable_pairwise_applications(self):
         """Properly compose two composable applications"""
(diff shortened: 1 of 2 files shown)


tests/test_app/test_init.py
```diff
@@ -70,6 +70,15 @@ def test_available_apps(self):
         self.assertIsInstance(apps, Table)
         self.assertTrue(apps.shape[0] > 10)
 
+    def test_available_apps_filter(self):
```
GavinHuttley, line 73: we're migrating to all new tests being written as pytest functions.

This test can be converted by just moving the test method to being a test function at the bottom of the file, removing the `self` etc...

Also, we should modify the test so it exactly check the values in the output. So here's my suggested replacement for your test

```python
def test_available_apps_filterl()
    '''"availabe_apps can be filtered by name'''''
    app_name_filter: str = "load"
    apps = available_apps (app_name_filter)
    assert isinstance(apps, Table)
    # check every returned table row 'name' has query is in the
    assert len(apps) > 0
    assert sum(app_name_filter in n for n in t.columns["name"]) == len(table)
```

GavinHuttley (changes requested):
Thanks so much for your contribution! Can you fix the small logic error and the test.

tests/test_app/test_init.py
cosmastech: I'm very new to Python. What's the reason for the switch to pytest from unittest?

GavinHuttley: There are many reasons, but the 2 main reasons. `pytest` provides mechanisms that simplify:

1. checking combinations of parameters for a test, a small amount of code (`pytest.mark.parametrize`)
2. elegant test setup and teardown (`pytest.fixture`)

GavinHuttley (approved):
Awesome job, thanks for the contribution!

cosmastech:
> Awesome job, thanks for the contribution!

Thanks for your patience with a newbie.

PR #1650, opened 2023-12-01
ENH: AnnotationDb.subset returns a subset of rows matching condition, fixes #1641

### LINKED ISSUE #1641: AnnotationDb's should support subsetting
If we have annotated features from entire genomes, then selecting a slice of some sequences will be more performant if we create a new AnnotationDb with only the subset of features that map to the selected sequences.

The `subset()` method should be added to the `SupportsQueryFeatures` protocol with the API matching `get_records_matching()`.  The implementations should return a new instance of the same db class, so separate implementations for Basic, Gff and Genbank.

This will support development of EnsemblLite

GavinHuttley:
[NEW] uses same querying interface as AnnotationDb.get_records_matching(), but
    returns a new instance of self.

[NEW] added cookbook example too

[NEW] added scriv changelog entries

### DIFF
--- a/changelog.d/20231201_134621_Gavin.Huttley.md
+++ b/changelog.d/20231201_134621_Gavin.Huttley.md
@@ -1,0 +1,41 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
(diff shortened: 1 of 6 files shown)


KatherineCaley (approved):
Looks good to me!

PR #1700, opened 2024-01-28
ENH: added new properties on SeqView

GavinHuttley:
Plus two additional commits to deal with the impact of having no access to my GitHub repo for 14 days.

Ignore the merge and the commit related to `Alignment.get_degapped_relative_to()`.

### DIFF
--- a/src/cogent3/core/sequence.py
+++ b/src/cogent3/core/sequence.py
@@ -2207,6 +2207,9 @@ class SeqView:
         if len(self) == 0:
             return self
 
+        if segment.start is not None and segment.start == segment.stop:
+            return _zero_slice
+
         slice_step = 1 if segment.step is None else segment.step
 
         if slice_step > 0:
--- a/src/cogent3/core/sequence.py
+++ b/src/cogent3/core/sequence.py
@@ -1888,7 +1888,15 @@ def _input_vals_neg_step(seqlen, start, stop, step):
 class SeqView:
     __slots__ = ("seq", "start", "stop", "step", "_offset")
 
-    def __init__(self, seq, *, start: int = None, stop: int = None, step: int = None):
+    def __init__(
+        self,
+        seq,
+        *,
+        start: int = None,
+        stop: int = None,
+        step: int = None,
+        offset: int = 0,
+    ):
         if step == 0:
             raise ValueError("step cannot be 0")
         step = 1 if step is None else step
@@ -1898,7 +1906,7 @@ class SeqView:
         self.seq = seq
         self.start = start
         self.stop = stop
-        self._offset = 0
+        self._offset = offset
         self.step = step
 
     @property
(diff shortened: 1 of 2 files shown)


tests/test_core/test_sequence.py
KatherineCaley, line 1: explicitly test `absolute_start` and `absolute_stop` for cases where this is a `offset`?

At this point you are relying on the testing of of `annotation_offset` which now calls `absolute_start` to be complete.

src/cogent3/core/sequence.py
```diff
@@ -1909,6 +1909,41 @@ def offset(self) -> int:
     def offset(self, value: int):
         self._offset = value or 0
 
+    @property
+    def absolute_start(self) -> int:
+        """returns the plus strand absolute start
```
KatherineCaley, line 1914: Needs a more informative doctoring. It's complex for two reasons:
1. because "absolute" has alternate meanings in maths compared to general language. 
2.  because `absolute_start` is wrt to the underlying seq, unless there has been an offset set for the underlying annotations in which case it is wrt to the coordinates of the annotations.

src/cogent3/core/sequence.py
```diff
@@ -1909,6 +1909,41 @@ def offset(self) -> int:
...
+            assert self.stop < 0, "expected stop on reverse strand SeqView < 0"
+            start = self.stop + len(self.seq) + 1
+        else:
+            start = self.start
+
+        return self.offset + start
+
+    @property
+    def absolute_stop(self) -> int:
+        """returns the plus strand absolute stop
```
KatherineCaley, line 1932: see previous comment on adding a bit of detail to docstring

tests/test_core/test_sequence.py
```diff
@@ -2381,6 +2381,27 @@ def test_sliced_seqview_rich_dict(reverse):
     assert rd["offset"] == 2
 
 
+@pytest.mark.parametrize(
+    "sl",
+    (
+        slice(2, 5, 1),  # positive indices, positive step
```
KatherineCaley, line 2387: would be good to test edge cases, especially for slices at the ends of the sequence.
- if you slice the whole sequence (or the whole seq in reverse), do you get 0 for start, and the length of sequence for end?
- singleton slices, and singleton slices at the start and ends (and reversed).

KatherineCaley (changes requested):

KatherineCaley:
Looks good! some small changes requested for the documentation and testing.

src/cogent3/core/sequence.py
GavinHuttley: Good points. Changing the name is probably best bet. Which of the following seems best
- `parent_start`
- `full_start`
- `resolved_start`
- `adjusted_start`
or something else?

tests/test_core/test_sequence.py
GavinHuttley: by singleton slice do you mean `slice(i, i+1)`?

KatherineCaley: yep

src/cogent3/core/sequence.py
KatherineCaley: I think `parent_start`

KatherineCaley (approved):

PR #1704, opened 2024-01-31
ENH: added SeqView.copy()

GavinHuttley:
[NEW] sliced argument controls whether truncating the represented
     sequence or leaves as is.

### DIFF
--- a/tests/test_core/test_sequence.py
+++ b/tests/test_core/test_sequence.py
@@ -2173,6 +2173,24 @@ def test_relative_position_step_GT_one(integer_seq):
     assert got == 4
 
 
+@pytest.mark.parametrize("sliced", (False, True))
+@pytest.mark.parametrize("rev", (False, True))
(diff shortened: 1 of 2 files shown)


tests/test_core/test_sequence.py
```diff
@@ -2173,6 +2173,22 @@ def test_relative_position_step_GT_one(integer_seq):
     assert got == 4
 
 
+@pytest.mark.parametrize("sliced", (False, True))
+@pytest.mark.parametrize("rev", (False, True))
```
KatherineCaley, line 2177: might be good to also test that when `sliced=False` that the underlying seq is not modified

KatherineCaley (changes requested):

tests/test_core/test_sequence.py
GavinHuttley: done

KatherineCaley (approved):

PR #1709, opened 2024-02-07
DEV: added seqid arg to SeqView, fixes #1708

### LINKED ISSUE #1708: add attribute to `SeqView` to keep track of the original seqid
The problem: given a sequence with annotations, if it is renamed, the connection to the underlying annotations will be broken.

Solving this requires keeping track of the "parent" name, which we are calling the seqid, via the `SeqView` class. 

The first step is to add `seqid` as an argument to the constructor of `SeqView`:

```py
class SeqView:
    def __init__(
        self,
        seq,
        *,
        start: int = None,
        stop: int = None,
        step: int = None,
        offset: int = 0,
        seqid: str = None,
    ):
```
and adding `_seqid` as an attribute of the class 

```py 
self._seqid = seqid
```

Then following class methods will need to be modified to cope:

- `to_rich_dict` 
- `from_rich_dict`
- `copy`

@GavinHuttley 

The expected behaviour is that when a `SeqView` instance is created, and is attached to a `Sequence` with a name, upon initialisation the `seqid` will be assigned the value of the `name`?

KatherineCaley:


src/cogent3/core/sequence.py
```diff
@@ -1932,6 +1933,7 @@ def __init__(
         self.stop = stop
         self._offset = offset
         self.step = step
+        self._seqid = seqid
 
     @property
     def offset(self) -> int:
```
GavinHuttley, line 1939: add a "read only" seqid property to the `SeqView` class
```python
@property
def seqid(self):
    return self._seqid
```

tests/test_core/test_sequence.py
```diff
@@ -2613,3 +2619,27 @@ def test_parent_coordinates(rev):
         seq = seq.rc()
 
     assert seq.parent_coordinates() == (0, 0, 1)
+
+
+def test_seqview_seqid():
+    sv = SeqView("ACGGTGGGAC")
+    assert sv._seqid is None
```
GavinHuttley, line 2626: change all references to the private attribute `_seqid` to the property `.seqid`

GavinHuttley (changes requested):
small tweak, we want the property for accessing the value

KatherineCaley:
> small tweak, we want the property for accessing the value

Duh! added @GavinHuttley

GavinHuttley (changes requested):
Can you add a test showing that when the SeqView is sliced, the seqid value is propagated to the new object

KatherineCaley:
> Can you add a test showing that when the SeqView is sliced, the seqid value is propagated to the new object

yep!

GavinHuttley (approved):
🎉 thank you!

PR #1719, opened 2024-02-16
DEV: Coverage only runs when arguments passed through nox command, fixes #1718

### LINKED ISSUE #1718: amend noxfile to take pytest args from command line
The coverage settings should only be automatically applied in GitHub actions. @rmcar17 can you apply the solution you showed me to fixing this.

rmcar17:
Addresses #1718


GavinHuttley (approved):
thanks @rmcar17 !

Issue #1715, opened 2024-02-13
Add `seq_len` attribute to `SeqView` class


fredjaya:
Currently, `SeqView` can only provide a view of a string, which is assigned to `self.seq`.

We require the ability to have a reference to a string rather than the string itself (i.e. positions in the parent seq). This will be later used for the WIP `SeqDataView` class, which will contain the`seq` data and provide a mechanism to return only a segment of the sequence.

At present, `SeqView` uses `len(self.seq)` in multiple places -- limiting the use of this class to cases where `.seq` is a Python iterable.

We can most easily remove this constraint by adding a `seq_len` argument (and corresponding `self.seq_len` attribute) to the `SeqView` class. 
For the current case of `SeqView.seq` as a string , `seq_len` should default to `None`. If None, assign `len(self.seq)` to `self.seq_len`  in `__init__()`.

Then, replace all uses within `SeqView` code of `len(self.seq)` with `self.seq_len`.

**Note:** because we use slots, we need to add `"seq_len"` to the slots variable too.

fredjaya:
Should `seq_len` be added to the outputs of special methods (i.e. __repr__) and when self is returned?

GavinHuttley:
> Should seq_len be added to the outputs of special methods (i.e. repr)

yes

> and when self is returned?

I don't understand, can you give a specific example?

GavinHuttley:
So in those cases, add the `seq_len=self.seq_len` argument for sure. Since we are planning for the case where subclasses cannot infer that from the value of `self.seq`

PR #1725, opened 2024-02-19
DEV: Add `seq_len` attribute to `SeqView`

fredjaya:
Addresses #1715.

- Add `seq_len` attribute and property
- Add tests for initialisation and `.copy()`
- Replaces `SeqView` use of `len(self.seq)` with `seq.seq_len`

### DIFF
--- a/src/cogent3/core/sequence.py
+++ b/src/cogent3/core/sequence.py
@@ -2141,6 +2141,7 @@ class SeqView:
             step=self.step * step,
             offset=self.offset,
             seqid=self.seqid,
+            seq_len=self.seq_len,
         )
 
     def _get_forward_slice_from_reverse_seqview_(self, slice_start, slice_stop, step):
@@ -2167,6 +2168,7 @@ class SeqView:
             step=self.step * step,
             offset=self.offset,
             seqid=self.seqid,
+            seq_len=self.seq_len,
         )
 
     def _get_reverse_slice(self, segment, step):
@@ -2222,6 +2224,7 @@ class SeqView:
             step=self.step * step,
             offset=self.offset,
             seqid=self.seqid,
+            seq_len=self.seq_len,
         )
 
     def _get_reverse_slice_from_reverse_seqview_(self, slice_start, slice_stop, step):
@@ -2260,6 +2263,7 @@ class SeqView:
             step=self.step * step,
             offset=self.offset,
             seqid=self.seqid,
+            seq_len=self.seq_len,
         )
 
     def __getitem__(self, segment):
@@ -2272,6 +2276,7 @@ class SeqView:
                 step=step,
                 offset=self.offset,
                 seqid=self.seqid,
+                seq_len=self.seq_len,
             )
 
         if segment.start is segment.stop is segment.step is None:
@@ -2308,6 +2313,7 @@ class SeqView:
                 step=self.step,
                 offset=self.offset,
                 seqid=self.seqid,
+                seq_len=self.seq_len,
             )
 
         return self.__class__(new_seq)
@@ -2326,7 +2332,7 @@ class SeqView:
         return (
             f"{self.__class__.__name__}(seq={seq!r}, start={self.start}, "
             f"stop={self.stop}, step={self.step}, offset={self.offset}, "
-            f"seqid={self.seqid!r})"
+            f"seqid={self.seqid!r}, seqlen={self.seq_len})"
         )
 
     def to_rich_dict(self):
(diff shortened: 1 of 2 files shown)


src/cogent3/core/sequence.py
```diff
@@ -1928,6 +1928,7 @@ def __init__(
         step: int = None,
         offset: int = 0,
         seqid: str = None,
+        seq_len: int = None,
```
GavinHuttley, line 1931: change to
```python
  start: Optional[int] = None,
  stop: Optional[int] = None,
  step: Optional[int] = None,
  offset: int = 0,
  seqid: Optional[str] = None,
  seq_len: Optional[int] = None
```
(Optional already imported)

src/cogent3/core/sequence.py
```diff
@@ -1941,6 +1942,7 @@ def __init__(
         self.step = step
         self._offset = offset
         self._seqid = seqid
+        self._seq_len = seq_len or len(self.seq)
```
GavinHuttley, line 1945: change this code to
```python
if seq_len is not None and seq_len != len(seq):
    raise AssertionError(f"{seq_len} != {len(self.seq)}")
```

tests/test_core/test_sequence.py
```diff
@@ -2720,3 +2726,36 @@ def test_empty_seqview_translate_position():
...
     assert sv.relative_position(0) == 0
+
+
+@pytest.mark.parametrize("start", (None, 0, 1, 10, -1, -10))
+@pytest.mark.parametrize("stop", (None, 10, 8, 1, 0, -1, -11))
+@pytest.mark.parametrize("step", (None, 1, 2, -1, -2))
+@pytest.mark.parametrize("length", (1, 8, 999))
+def test_seqview_seqlen_init(start, stop, step, length):
+    """seq_len is length of seq when None"""
+    seq_data = "0123456789"
```
GavinHuttley, line 2737: this test will break due to the new assertion error. Since we don't care what the sequence is change the seq data to
```python
seq_data = "A" * length
```

tests/test_core/test_sequence.py
```diff
@@ -2720,3 +2726,36 @@ def test_empty_seqview_translate_position():
...
+    got = SeqView(seq="ACTG", seq_len=length).seq_len
+    assert got == length
+
+
+def test_seqview_copy_propagates_seq_len():
+    seq = "ACGGTGGGAC"
+    sv = SeqView(seq)
+    copied = sv.copy()
+    assert copied.seq_len == len(seq)
+
```
GavinHuttley, line 2754: add a new test that triggers the assertion warning

GavinHuttley (changes requested):
thanks @fredjaya ! A few amendments and the type hinting catch from @rmcar17

src/cogent3/core/sequence.py
fredjaya: Should
 ```python
self._seq_len = seq_len or len(self.seq)
```
still be included? Or will `seq_len` always need to be provided?

File: src/cogent3/core/sequence.py [hunk 5]
```diff
@@ -1942,6 +1942,8 @@ def __init__(
         self.step = step
         self._offset = offset
         self._seqid = seqid
+        if seq_len and seq_len != len(seq):
```
GavinHuttley, line 1945: this should still be `is None` because `seq_len=0` and a `seq="AGT"` would not be caught by this check

src/cogent3/core/sequence.py
GavinHuttley: Good question, that should still be included!

The check is just to make sure if they are both provided that they're at least consistent.

tests/test_core/test_sequence.py
```diff
@@ -2732,18 +2732,17 @@ def test_empty_seqview_translate_position():
...
     sv = SeqView(seq_data, start=start, stop=stop, step=step)
     expect = len(seq_data)
     # Check property and slot
     assert sv.seq_len == expect
     assert sv._seq_len == expect
 
-    # Expect input seq_len to be correct if provided
-    got = SeqView(seq="ACTG", seq_len=length).seq_len
-    assert got == length
+    with pytest.raises(AssertionError):
```
GavinHuttley, line 2744: Make this a new test (make sure test function name is unique) and give it 3 different conditions using `pytest.mark.parametrize` (look at examples in the module for using this).
We want to check that the following conditions all fail
- `seq_len=0, seq="A"`
- `seq_len=01, seq=""`
- `seq_len=2, seq="A"`

GavinHuttley (changes requested):
Getting cloe, a few more test conditions

GavinHuttley (approved):
And great job catching those `seqlen` instead of  `seq_len` cases!

PR #1753, opened 2024-03-13
feature/issue 1731

### LINKED ISSUE #1731: mark binomial_exact function for removal
**What should be changed**

`cogent3.maths.stats.binomial_exact()`

**Why**

Not used in cogent3, available in scipy.stats?

**How much notice**

6 months notice. This should be represented in the user warning as a specific version. e.g. <current year>.<current month + *x*> where *x* is from the PR date.  

**Guidelines for deprecation**

As this is a function that will be removed, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function will be removed) as 3 months from when you make the change.

ChangsenJiang:
DEP: remove the binomial_exact() function and the related tests , fixes #1731  
Remove the `cogent3.maths.stats.binomial_exact()` function, and removed the related `test_binomial_series()`,  `test_binomial_exact()`, `test_binomial_exact_floats()`, and `test_binomial_exact_errors()`

### DIFF
--- a/src/cogent3/maths/stats/distribution.py
+++ b/src/cogent3/maths/stats/distribution.py
@@ -74,9 +74,7 @@ def poisson_exact(successes, mean):
 
 
 @c3warn.deprecated_callable(
-    version="2024.9",
-    reason="use scipy.stats.binom.pmf()instead", 
-    is_discontinued=True
+    version="2024.9", reason="use scipy.stats.binom.pmf()instead", is_discontinued=True
 )
 def binomial_exact(successes, trials, prob):
     """Returns binomial probability of exactly X successes.
--- a/src/cogent3/maths/stats/distribution.py
+++ b/src/cogent3/maths/stats/distribution.py
@@ -5,7 +5,7 @@ which is (c) Stephen L. Moshier 1984, 1995.
 
 from numpy import arctan as atan
 from numpy import array, exp, sqrt
-from scipy.stats import binom, f, norm, t
+from scipy.stats import f, norm, t
 from scipy.stats.distributions import chi2
 
 from cogent3.maths.stats.special import (
@@ -23,8 +23,8 @@ from cogent3.maths.stats.special import (
     log1p,
     ndtri,
 )
-from cogent3.util import warning as c3warn
 
+from cogent3.util import warning as c3warn
 
 # ndtri import b/c it should be available via this module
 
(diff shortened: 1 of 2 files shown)


src/cogent3/maths/stats/distribution.py
```diff
@@ -72,20 +73,12 @@ def poisson_exact(successes, mean):
         return pdtrc(successes - 1, mean) - pdtrc(successes, mean)
 
 
+@c3warn.deprecated_callable(
+    version="2024.9", reason="use scipy.stats.binom.pmf()instead", is_discontinued=True
+)
 def binomial_exact(successes, trials, prob):
```
GavinHuttley, line 79: The point here is not to replace our function with that from Scipy; rather, just use the decorator as indicated. Revert the change within the function to the original.

src/cogent3/maths/stats/distribution.py
```diff
@@ -5,7 +5,7 @@
 
 from numpy import arctan as atan
 from numpy import array, exp, sqrt
-from scipy.stats import f, norm, t
+from scipy.stats import binom, f, norm, t
```
GavinHuttley, line 8: delete the new import

GavinHuttley (changes requested):
Thanks for the contribution!

When marking a function to be discontinued we limit our changes to applying the decorator and giving the reason (including referring users to a replacement in another project). In other words, we do not actually change our code within the function since we are going to be deleting it.

src/cogent3/maths/stats/distribution.py
ChangsenJiang: Thanks for your reply! Does that mean what I'm going to do is keep the @c3warn.deprecated_callable message at the top, revert everything inside the function (including return), and then delete the new modules I imported?

src/cogent3/maths/stats/distribution.py
ChangsenJiang: Now I revert the change within the function to the original and delete the new imports. Please have a look : )

ChangsenJiang:
Now I revert the change within the function to the original and delete the new imports. Please have a look : )

ChangsenJiang:
Now I revert the change within the function to the original and delete the new imports. Please have a look : )

GavinHuttley (changes requested):
Thanks for your effort, but I can't accept this PR as is.

First, the replacement function in scipy you have identified is incorrect. It should be `scipy.stats.binomtest`. This means you would need to update your deprecation reason and, in doing so, correct the spacing between words.

Second, you would need to identify all uses of `binomial_exact()` in `cogent3` and replace them with correct usage of the scipy function.

Finally, be sure and delete imports of `binomial_exact()` since it is no longer being used.

GavinHuttley:
Closing due to lack of response to review. But thanks for your effort!

PR #1757, opened 2024-03-13
feature/issue 1632

### LINKED ISSUE #1632: improve error message for invalid sequence file format
The so-called "relaxed" phylip format is not handled by the cogent3's phylip parser. But worse is that the error message is not very informative.
For the attached minimal file [locus-0.phy.zip](https://github.com/cogent3/cogent3/files/13421981/locus-0.phy.zip), I get the following error
```python
from cogent3 import load_aligned_seqs

aln = load_aligned_seqs("/Users/gavin/Desktop/Inbox/locus-0.phy", format="phylip", moltype="dna")
```
gives
```
...
   [3840](https://untitled+.vscode-resource.vscode-cdn.net/~/miniconda3/envs/c311/lib/python3.11/site-packages/cogent3/core/alignment.py:3840) seq_lengths = set(len(s) for s in seqs)
   [3841](https://untitled+.vscode-resource.vscode-cdn.net/~/miniconda3/envs/c311/lib/python3.11/site-packages/cogent3/core/alignment.py:3841) if len(seq_lengths) != 1:
-> [3842](https://untitled+.vscode-resource.vscode-cdn.net/~/miniconda3/envs/c311/lib/python3.11/site-packages/cogent3/core/alignment.py:3842)     raise ValueError("not all sequences have same length")

ValueError: not all sequences have same length
```

KyleSu12:
ENH: Improved error message for sequence length validation, fixed #1632
Added a clear and informative error message that triggers if input sequences are not of the same length, detailing the unique lengths found.


GavinHuttley (changes requested):
Thanks for the contribution! This needs to be accompanied by a test before it can be accepted.

In this case, the test should be a new function at the end of `tests/test_core/test_alignment.py` and the test should check that the error is triggered when mixed length sequences are provided and that the error message matches, e.g.
```python
def test_array_align_error_with_mixed_length():
    data = dict(s1="ACGG", s2="A-G")
    with pytest.raises(ValueError, match=".* not all the same length.*"):
        make_aligned_seqs(data=data)
```
that test checks that both the error type AND the message match. This is important because the error type has not changed compared to the original code, so the only way to be sure you're patch is correct is to include the "match" condition.

KyleSu12:
Thanks Gavin, I've added the this verification and passed the updated pytest.

Issue #1740, opened 2024-02-24
mark identity() function for removal


GavinHuttley:
**What should be changed**

`cogent3.util.misc.identity()`

**Why**

Not used in cogent3.

**How much notice**

6 months notice.

**Guidelines for deprecation**

Checking no internal uses will be tricky numpy and scipy also have functions called `identity()`.

As this is a function that will be removed, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function will be removed) as <current year>.<PR month + *x*> where *x* is defined above.

sharonli126:
I would like to do this issue

GavinHuttley:
awesome! let me know if you run into any difficulties.

sharonli126:
Hi Gavin, I have fixed the issue and synced the change. Please have a look. thanks.

PR #1759, opened 2024-03-13
feature/ issue #1740

sharonli126:
it is marked

### DIFF
--- a/compose-dev.yaml
+++ b/compose-dev.yaml
@@ -1,0 +1,12 @@
+services:
+  app:
+    entrypoint:
+    - sleep
+    - infinity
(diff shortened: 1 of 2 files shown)


compose-dev.yaml
GavinHuttley, line 1: You should never commit changes to development related processes to an open source project!

GavinHuttley (changes requested):
Thanks for the contribution. It looks like you have accidentally committed a change you made to your local Docker environment into the repository.

You also need to remove the tests of this deprecated function. But thanks for the contribution!

GavinHuttley:
Thanks for your effort, but we cannot accept this PR as it doesn't meet the project standards.

PR #1761, opened 2024-03-13
Feature/issue 1739

### LINKED ISSUE #1739: mark tprob() function for removal
**What should be changed**

`cogent3.maths.distribution.tprob()`

**Why**

Not used in cogent3, available in scipy.stats?

**How much notice**

6 months notice.

**Guidelines for deprecation**

As this is a function that will be removed, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function will be removed) as <current year>.<PR month + *x*> where *x* is defined above.

YouzhiHuang:
DEP: discontinued cogent3.maths.distribution.tprob(), fixes #1739 
Mark tprob() function for removal and delete test for tprob() Not used in cogent3, available in scipy.stats

### DIFF
--- a/src/cogent3/maths/stats/distribution.py
+++ b/src/cogent3/maths/stats/distribution.py
@@ -41,7 +41,6 @@ def zprob(x):
 @c3warn.deprecated_callable(
     version="2024.9", reason="use scipy.stats.t.sf() instead", is_discontinued=True
 )
-
 def tprob(x, df):
     """Returns both tails of t distribution (-infinity to -x, infinity to x)"""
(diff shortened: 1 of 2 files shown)


GavinHuttley (commented):
Unfortunately, `test_compare_by_tip_distances()` fails. This test fails because ultimately, there's a call to `tprob()` in `cogent3.maths.stats.test`. You deleted the import, but did not check that the tests still passed. You would need to replace that call with the `scipy.stats.t.sf()` so that the tests pass.

But thanks for the effort!

PR #1762, opened 2024-03-13
Feature/issue 1737

Firimp:
Remove `cogent3.maths.special.combinations()`, `cogent3.maths.special.combinations_exact()`, `cogent3.maths.special.ln_combinations()` and correlated test functions
Closes https://github.com/cogent3/cogent3/issues/1737


Firimp:
It seems I made some mistakes when writing commit messages, like 'STY: improve indentation in special.py' should be 'STY: improve indentation in special.py', and 'DEP: REV: revert the changes within functions in issue#1737' should be 'STY: change the docstring of deprecated functions to the original version'.
The lateist version of the branch has met all the requirements in https://github.com/cogent3/cogent3/issues/1737 and it runs well in local, I just wonder what I can do to compensate for these wrong commit messages.

GavinHuttley (changes requested):
I note that the changes cause thousands of warnings (displayed after running pytest). To really address the issue, we need to replace all uses of the deprecated functions with the named replacement from `scipy`.

Firimp:
> I note that the changes cause thousands of warnings (displayed after running pytest). To really address the issue, we need to replace all uses of the deprecated functions with the named replacement from `scipy`.

I used the "search for text" feature to search the entire repository, but I didn't find any of the three functions from special.py being called, which leaves me puzzled about why so many warnings are being generated. However, I did discover another function also named combinations(), from the itertools module, being widely called. I'm wondering if this could be the primary reason for the warnings.

src/cogent3/maths/stats/special.py
```diff
@@ -179,8 +184,9 @@ def ln_binomial(successes, trials, prob):
     Note: no error checking (expects to be called through binomial_exact())
     """
     prob = fix_rounding_error(prob)
+    result_of_ln_combinations = lgam(trials + 1) - lgam(successes + 1) - lgam(trials - successes + 1)
```
GavinHuttley, line 187: While this no longer calls the deprecated function, it should use `scipy.special.comb()` instead.

GavinHuttley (changes requested):
Thanks for amending, but this should use the nominated replacement from scipy rather than just duplicating lines of code from elsewhere.

src/cogent3/maths/stats/special.py
Firimp: I don't really think the `lgam()` function is deprecated, and the function is sophisticated enough that it can't simply be replaced by `scipy.special.comb()`.

GavinHuttley: I take your point that `comb` is not a dropin replacement, however `lgam()` will eventually be deprecated. I suggest the following then:
```python
from scipy.special import loggamma # import statement inside the function, since we will be removing this function too

result_of_ln_combinations = loggamma(trials + 1) - loggamma(successes + 1) - loggamma(trials - successes + 1)
```

GavinHuttley (changes requested):
Use scipy functions instead of cogent3 ones where possible as we seek to eliminate this entire module eventually.

Firimp:
Done！

GavinHuttley (approved):
Thanks for your persistence in making the necessary changes!

I've mad a couple of small edits, but happy to accept and merge now!

PR #1764, opened 2024-03-13
feature/issue-1628

GaoPeizhong:
I for path: / workspaces/Python/cogent3 / SRC/cogent3 / app/data_store. Py file ReadOnlyDataStoreZipped has added new features, this can let the bug is eliminated

### DIFF
--- a/issue.py
+++ b/issue.py
@@ -1,4 +1,3 @@
 from cogent3 import open_data_store
 ds = open_data_store("loci.zip", suffix="phy", mode="r")
-ds.members[:5]
 print(ds.members[:5])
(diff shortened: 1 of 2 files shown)


GavinHuttley (commented):
Unfortunately, I can't accept this contribution as it is quite incomplete. Here are a few points that will hopefully help you with your next contribution to an open source project.

- When you're unsure about how you should do something, please ask the project developers!
- Based on our chat history you were able to exclude names that begin with a ".". It is best, however, to make a minimal set of amendments. In your case, a small number of changes to the existing `_iter_matches()` method is much better than deleting it and then reinstating it in another place.
- A solid contribution also includes new tests that fail without your code amendments and pass with them.
- Unless called for, don't add new files into a repository.

But I thank you for your effort.

GavinHuttley:
Did not meet contribution standards. But thanks for the effort!

Issue #1733, opened 2024-02-24
mark poisson_low() function for removal


GavinHuttley:
**What should be changed**

`cogent3.maths.stats.poisson_low()`

**Why**

Not used in cogent3, available in scipy.stats?

**How much notice**

6 months notice. This should be represented in the user warning as a specific version. e.g. <current year>.<current month + *x*> where *x* is from the PR date.  

**Guidelines for deprecation**

As this is a function that will be removed, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function will be removed) as the number above from when you make the change.

Zongjing-Han:
I would like to do this issue

GavinHuttley:
fixed in https://github.com/cogent3/cogent3/commit/535a2992b16a832e828f4304b59f3d300c3e8973

PR #1768, opened 2024-03-13
Feature/issue 1733

Zongjing-Han:
Remove the cogent3 poisson_low() function, and removed the related test_poisson_low()


Zongjing-Han:
DEP: remove the poisson_low() function and the related tests, fixes #1733
Remove the cogent3 poisson_low() function, and removed the related test_poisson_low()

PR #1769, opened 2024-03-13
feature/issue 1734

Tesla686:
poisson_exact() function for removal #1734


Tesla686:
Because this is a marker that this function will be removed, but it has not actually been removed, so I am thinking whether I need to delete its corresponding test function?

GavinHuttley:
Hi @Tesla686,
Yes you should delete the corresponding test function. Thanks!

PR #1777, opened 2024-03-14
BUG: fix representation of numpy arrays in sqlite databases

### LINKED ISSUE #1776: BUG: GenbankAnnotationDb and GffAnnotationDb files are not OS independent
**Describe the bug**
In [annotation_db.py](https://github.com/cogent3/cogent3/blob/develop/src/cogent3/core/annotation_db.py) when a numpy array is saved to or loaded from the sqlite database the [bytes representation](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html) of the array is used. This representation doesn't store information about the type of the array.

My Ubuntu installation uses a default of `np.int64` for the array, whereas my Windows installation uses a default of `np.int32`. This means when I load on Windows a `GenbankAnnotationDb` that was saved on Ubuntu, the spans column is padded with extra zeros. This breaks OS independence of the saved files.

**To Reproduce**
Load a `GenbankAnnotationDb` on an OS that uses a different default `numpy` data type from the one where it was saved.

**Expected behavior**
These files should be OS independent.

**Desktop (please complete the following information):**
 - OS: Windows 22H2 19045.4170, Ubuntu 22.04.4 LTS

rmcar17:
Stores arrays in the sqlite database in the [.npy format](https://numpy.org/devdocs/reference/generated/numpy.lib.format.html) to restore OS independence.

Importantly, this change is not backwards compatible with previously saved databases that include numpy arrays. Notably, previous versions of `GffAnnotationDb` and `GenbankAnnotationDb` will not be able to be loaded.

Fixes #1776

### DIFF
--- a/src/cogent3/core/annotation_db.py
+++ b/src/cogent3/core/annotation_db.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections
 import copy
 import inspect
+import io
 import json
 import os
 import pathlib
@@ -37,13 +38,16 @@ _is_ge_3_11 = (sys.version_info.major, sys.version_info.minor) >= (3, 11)
 # Define custom types for storage in sqlite
 # https://stackoverflow.com/questions/18621513/python-insert-numpy-array-into-sqlite3-database
 def array_to_sqlite(data):
-    return data.tobytes()
+    out = io.BytesIO()
+    numpy.save(out, data)
+    out.seek(0)
+    return out.read()
 
 
 def sqlite_to_array(data):
-    result = numpy.frombuffer(data, dtype=int)
-    dim = result.shape[0] // 2
-    return result.reshape((dim, 2))
+    out = io.BytesIO(data)
+    out.seek(0)
+    return numpy.load(out)
 
 
 def dict_to_sqlite_as_json(data: dict) -> str:
@@ -72,14 +76,11 @@ class FeatureDataType(typing.TypedDict):
 
 @typing.runtime_checkable
 class SerialisableType(typing.Protocol):  # pragma: no cover
-    def to_rich_dict(self) -> dict:
-        ...
+    def to_rich_dict(self) -> dict: ...
 
-    def to_json(self) -> str:
-        ...
+    def to_json(self) -> str: ...
 
-    def from_dict(self, data):
-        ...
+    def from_dict(self, data): ...
 
 
 @typing.runtime_checkable
@@ -96,18 +97,15 @@ class SupportsQueryFeatures(typing.Protocol):  # pragma: no cover
         strand: OptionalStr = None,
         attributes: OptionalStr = None,
         on_alignment: OptionalBool = None,
-    ) -> typing.Iterator[FeatureDataType]:
-        ...
+    ) -> typing.Iterator[FeatureDataType]: ...
 
     def get_feature_children(
         self, *, name: str, start: OptionalInt = None, end: OptionalInt = None, **kwargs
-    ) -> typing.List[FeatureDataType]:
-        ...
+    ) -> typing.List[FeatureDataType]: ...
 
     def get_feature_parent(
         self, *, name: str, start: OptionalInt = None, end: OptionalInt = None, **kwargs
-    ) -> typing.List[FeatureDataType]:
-        ...
+    ) -> typing.List[FeatureDataType]: ...
 
     def num_matches(
         self,
@@ -118,8 +116,7 @@ class SupportsQueryFeatures(typing.Protocol):  # pragma: no cover
         strand: OptionalStr = None,
         attributes: OptionalStr = None,
         on_alignment: OptionalBool = None,
-    ) -> int:
-        ...
+    ) -> int: ...
 
     def subset(
         self,
@@ -131,8 +128,7 @@ class SupportsQueryFeatures(typing.Protocol):  # pragma: no cover
         end: OptionalInt = None,
         strand: OptionalStr = None,
         attributes: OptionalStr = None,
-    ):
-        ...
+    ): ...
 
 
 @typing.runtime_checkable
@@ -149,8 +145,7 @@ class SupportsWriteFeatures(typing.Protocol):  # pragma: no cover
         attributes: OptionalStr = None,
         strand: OptionalStr = None,
         on_alignment: bool = False,
-    ) -> None:
-        ...
+    ) -> None: ...
 
     def add_records(
         self,
@@ -158,8 +153,7 @@ class SupportsWriteFeatures(typing.Protocol):  # pragma: no cover
         records: typing.Sequence[dict],
         # seqid required for genbank
         seqid: OptionalStr = None,
-    ) -> None:
-        ...
+    ) -> None: ...
 
     def update(self, annot_db, seqids: OptionalStrList = None) -> None:
         # update records with those from an instance of the same type
--- a/src/cogent3/core/annotation_db.py
+++ b/src/cogent3/core/annotation_db.py
@@ -11,6 +11,7 @@ import re
 import sqlite3
 import sys
 import typing
+import warnings
 
 import numpy
 import typing_extensions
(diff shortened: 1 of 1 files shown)


src/cogent3/core/annotation_db.py
```diff
@@ -72,14 +76,11 @@ class FeatureDataType(typing.TypedDict):
 
 @typing.runtime_checkable
 class SerialisableType(typing.Protocol):  # pragma: no cover
-    def to_rich_dict(self) -> dict:
-        ...
+    def to_rich_dict(self) -> dict: ...
 
-    def to_json(self) -> str:
-        ...
+    def to_json(self) -> str: ...
```
GavinHuttley, line 81: eek! this looks like a black version issue. Can you please check you run the current pinned version against the repo, it should revert these.

GavinHuttley (changes requested):
This will fail with a `ValueError` if an array was previously stored using `tobytes()`.
```python
import io
import numpy

a = numpy.arange(4)
b = a.tobytes()

out = io.BytesIO(b)
out.seek(0)
numpy.load(out)
```
Add a test of this and amend the `sqlite_to_array()` function to trap that exception and try the "old" approach.

src/cogent3/core/annotation_db.py
rmcar17: Whoops! Looks like vscode is trying to use the right version, but something else is going on. I'll run the commands manually for now.

rmcar17:
> Add a test of this and amend the `sqlite_to_array()` function to trap that exception and try the "old" approach.

One concern is that this doesn't convert the values in the `spans` column from the old to the new format, meaning such old files remain OS dependent. Should we provide a warning when we try the "old" approach that suggests to the user a new function to convert the format of that column? An alternative would be that we silently update the entries in that column, similar to how we automatically rename the `end` column to `stop`.

I don't like the latter option as much as for one it assumes a user is running the program on their usual OS - if they don't it would lead to an incorrect representation in the new format.

GavinHuttley:
> ...Should we provide a warning when we try the "old" approach that suggests to the user a new function to convert the format of that column?

I think this is the right solution. As you point out, could be bad side effects. One thought is to include a reference to the issue this is solving so the user can go and see the cross-platform nature of this.

src/cogent3/core/annotation_db.py
```diff
@@ -1603,3 +1631,77 @@ def load_annotations(
...
+    data : bytes
+        The sqlite3 representation of the numpy array.
+
+    Returns
+    -------
+    bytes
+        The new sqlite3 representation of the numpy array.
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
```
GavinHuttley, line 1655: does this stop the DeprecationWarning from showing?

GavinHuttley (commented):
clarifying one issue with PR

src/cogent3/core/annotation_db.py
```diff
@@ -1622,3 +1631,77 @@ def load_annotations(
...
+        new_data = array_to_sqlite(sqlite_to_array(data))
+    return new_data
+
+
+def update_file_format(
+    source_file: os.PathLike,
+    backup_file: typing.Optional[os.PathLike],
```
GavinHuttley, line 1662: Another way to do this is make backup "opt-out" by the user, so make `backup_file=True` the default. We can just append `.bak` to the original file path.

src/cogent3/core/annotation_db.py
```diff
@@ -56,6 +57,14 @@ def sqlite_to_array(data):
         result = numpy.frombuffer(data, dtype=int)
         dim = result.shape[0] // 2
         result = result.reshape((dim, 2))
+
+        warnings.warn(
+            "Old OS dependent file format detected. "
+            "Please update file format to latest version. "
```
GavinHuttley, line 63: Change the second line to explicit advice `Update the file format using cogent3.core.annotation_db.update_file_format()`

src/cogent3/core/annotation_db.py
```diff
@@ -56,6 +57,14 @@ def sqlite_to_array(data):
...
+            "See: https://github.com/cogent3/cogent3/issues/1776.",
```
GavinHuttley, line 64: Add `For reason see https://...`

GavinHuttley (changes requested):
tweaks to `update_file_format()` plus mods to warning message

src/cogent3/core/annotation_db.py
rmcar17: Good idea!

src/cogent3/core/annotation_db.py
rmcar17: Will do!

src/cogent3/core/annotation_db.py
```diff
@@ -1659,36 +1659,50 @@ def _update_array_format(data: bytes) -> bytes:
...
+    backup : bool, optional
+        If True (default), performs a backup of the database before updating.
+        Otherwise does not perform a backup prior to update (not recommended).
     """
     anno_db = db_class(source=source_file)
 
-    if backup_file is not None:
-        anno_db.write(backup_file)
+    if backup:
+        backup_path = source_file + ".bak"
```
GavinHuttley, line 1697: This won't work if `source_file` is a `Path` object. Also, doesn't cover the case where a user may provide a path with `"~"`,  I suggest the following:
```python
source_file = pathlib.Path(source_file)
...
backup_path = source_file.parent / f"{source_path.name}.bak"
if backup_path.exists():
...
```

GavinHuttley (changes requested):
handle possible `source_path` type issues

src/cogent3/core/annotation_db.py
```diff
@@ -37,13 +39,33 @@
...
+        result = numpy.load(out)
+    except ValueError:
+        # array is not stored in the numpy.save format
+        # attempt to read from the old format where the
+        # array was saved using numpy.ndarray.tobytes
+        result = numpy.frombuffer(data, dtype=int)
+        dim = result.shape[0] // 2
+        result = result.reshape((dim, 2))
+
+        warnings.warn(
```
rmcar17, line 61: I've been scratching my head for the last little bit on this. When this code is run through a WIP test I have going on in pytest, pytest is able to capture and display the warning. However, if I take the equivalent code in pytest and run it in its own file, the warnings are not displayed - unless the warning is also specifically given the parameter stacklevel=4.

src/cogent3/core/annotation_db.py
```diff
@@ -37,13 +39,33 @@
...
+            "Old OS dependent file format detected. "
+            "Update the file format using cogent3.core.annotation_db.update_file_format() "
+            "For reason see https://github.com/cogent3/cogent3/issues/1776.",
+            DeprecationWarning,
+        )
```
rmcar17, line 66: I've been scratching my head for the last little bit on this. When this code is run through a WIP test I have going on in pytest, pytest is able to capture and display the warning. However, if I take the equivalent code in pytest and run it in its own file, the warnings are not displayed - unless the warning is also specifically given the parameter stacklevel=4.

src/cogent3/core/annotation_db.py
```diff
@@ -37,13 +39,33 @@
...
+        # attempt to read from the old format where the
+        # array was saved using numpy.ndarray.tobytes
+        result = numpy.frombuffer(data, dtype=int)
+        dim = result.shape[0] // 2
+        result = result.reshape((dim, 2))
+
+        warnings.warn(
+            "Old OS dependent file format detected. "
+            "Update the file format using cogent3.core.annotation_db.update_file_format() "
+            "For reason see https://github.com/cogent3/cogent3/issues/1776.",
```
rmcar17, line 64: I've been scratching my head for the last little bit on this. When this code is run through a WIP test I have going on in pytest, pytest is able to capture and display the warning. However, if I take the equivalent code in pytest and run it in its own file, the warnings are not displayed - unless the warning is also specifically given the parameter stacklevel=4.

rmcar17: It does work if I use a `UserWarning` instead of a `DeprecationWarning`, so perhaps something weird going on with the filters.

rmcar17: I'm also not entirely convinced this is the correct place to put the warning. For instance, the warning contains no information about the name of the bad file and would only be raised lazily when they attempt to do something that involves reading from the spans column. I think ideally, users should be warned when they load the database, but this would involve verifying all rows in every array type column on load.

GavinHuttley: This is difficult. As you point out, doing it properly is a significant effort. It's better to have some warning rather than none. Let's leave it at that for now.

GavinHuttley (approved):
thank you!

PR #1779, opened 2024-03-14
Cogent3 plugin

khiron:
Pull request into a new branch `cogent3-plugin`

replaced app_register for stevedore.ExtensionManager


tests/test_app/test_init.py
```diff
@@ -181,6 +174,7 @@ def main(self, val: int) -> int:
         return val + self.constant
 
 
+@pytest.mark.xfail(reason="Constructing apps on the fly is no longer supported")
```
GavinHuttley, line 177: This comment not really directed at this test, but your comment raises an interesting issue. I think we should still support constructing apps on the fly, BUT not support them being hooked into the extension system. So they would have to be explicitly imported.

GavinHuttley: Now to address this xfail. We do need to be testing that, for a given app, we correctly generate docs. IO think this means we should write a new function called `_make_apphelp_docstring(app)`. This should basically execute the commands within `app_help()`
```python
def _make_apphelp_docstring(app):
    app_doc = app.__doc__ or ""
    init_doc = app.__init__.__doc__ or ""
    if init_doc.strip():
        docs.extend(["", _clean_params_docs(init_doc)])

    types = _make_types(app)
    docs.extend([""] + _make_head("Input type") + [", ".join(types["_data_types"])])
    docs.extend([""] + _make_head("Output type") + [", ".join(types["_return_types"])])

    return "\n".join(docs)

def app_help(...):
    ...
    print(_make_apphelp_docstring(app))
```
This function can be tested on any app.

src/cogent3/app/__init__.py
```diff
@@ -6,34 +6,27 @@
...
-    "io",
-    "sample",
-    "translate",
-    "tree",
-]
-
+def _get_app_attr(name):
+    """
+    returns app details for display
+    NB: this loads the module the app is in
```
GavinHuttley, line 19: use `numpydoc` style, i.e. Notes is a rst heading...

src/cogent3/app/__init__.py
```diff
@@ -155,9 +155,7 @@ def _get_app_matching_name(name: str):
         raise NameError(f"no app matching name {name!r}")
     elif table.shape[0] > 1:
         raise NameError(f"too many apps matching name {name!r},\n{table}")
-    modname, _ = table.to_list(columns=["module", "name"])[0]
-    mod = importlib.import_module(modname)
-    return getattr(mod, name)
+    return apps()[name].plugin
```
GavinHuttley, line 158: if this generates a `KeyError` when `name` is incorrect, I think we might want to trap that call and raise a `ValueError` instead, with a custom message making it clear users should check for typos

GavinHuttley (changes requested):
Great progress! A few amendments and then we can start trying to fix the test errors.

src/cogent3/app/composable.py
```diff
@@ -704,7 +699,7 @@ def _as_completed(self, dstore, parallel=False, par_kw=None, **kwargs) -> Genera
 
 def is_composable(obj):
```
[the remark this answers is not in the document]
khiron: I think for most uses of is_compsable, is_app is sufficient.  

But one test of concat specificly is testing an app that is not composable, so I added a function `is_app_composable()`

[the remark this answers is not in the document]
khiron: I also added an optional boolean argument to `def apps(force: bool = False):` which forces a reload of the ExtensionManager, which is otherwise lazy loaded and cached.

khiron:
coverage dropped because I don't test that multiple apps with the same name don't trigger the exception for when that is the case.  I'm still working out how I can test that without littering the production code base with test artifacts

src/cogent3/app/__init__.py
```diff
@@ -15,20 +15,21 @@
...
 
-    is_composable = hasattr(obj, "app_type") and obj.app_type != App_type.NON_COMPOSABLE
-
     _types = _make_types(obj)
 
     return [
         obj.__module__,
         name,
-        is_composable,
+        hasattr(obj, "app_type"),
```
GavinHuttley, line 32: Oh dear, I'm about to contradict myself! So sorry, I now see where you were applying that call to validate the app is composable. You were right!!

I think the issue I was concerned about relates to line 25. At this point we have a putative app .. some developer indicates this is a cogent3 app. I think we should introduce a check on line 26 as
```python
if not is_app(app):
    warnings.warn(f"{obj!r} from {obj.__module__!r} is not a valid cogent3 app, not skipping")
```
that is, we warn the user that app is invalid, we do not include it, but we shouldn't fail. What do you think?

GavinHuttley (changes requested):
sorry, but this review is incomplete due to me running out of time. Just wanted to confess my mistake so you can make progress!

src/cogent3/app/__init__.py
khiron: Yes - if an object return from stevedore is not is_app, it shouldn't appear in available_apps.  

I think we still need is_composable in the 3rd row of the attr list.  Otherwise the third column will be meaningless in available_apps (all will be True)

khiron: I can't test this adequately without a mechanism to on-the-fly create entry_points that stevedore will load.

khiron: Working on a strategy to do that by creating a temp dir, creating a setup.py and a module, pip installing it, running a test, then uninstalling it, forcing refresh of the ExtensionManager, and then deleting the temp dir.

src/cogent3/app/composable.py
```diff
@@ -702,9 +697,14 @@ def _as_completed(self, dstore, parallel=False, par_kw=None, **kwargs) -> Genera
     return ui.series(to_do, count=len(mapped), **kwargs)
 
 
-def is_composable(obj):
-    """checks whether obj has been registered by the composable decorator"""
-    return __app_registry.get(get_object_provenance(obj), False)
+def is_app_composable(obj):
+    """checks whether obj has been decorated by define_app and it's app_type attribute is not NON_COMPOSABLE"""
+    return hasattr(obj, "app_type") and obj.app_type is not NON_COMPOSABLE
```
GavinHuttley, line 702: make this function more robust to potential future changes in decorator behaviour, change line to
```python
return is_app(obj) and obj.app_type is not NON_COMPOSABLE
```

GavinHuttley (changes requested):
awesome job! one very small tweak and done (until the next bit)

tests/test_app/test_composable.py
```diff
@@ -1129,3 +1132,110 @@ def delme2(val: c3types.SerialisableType) -> dict:
...
+        return data.upper()
+
+
+@pytest.mark.parametrize("install_temp_app", [temp_app_uppercase], indirect=True)
+def test_add_app_with_result(install_temp_app):
+    appercase = get_app("temp_app_uppercase")
+    assert appercase("hello") == "HELLO"
+
+
+@pytest.fixture
```
khiron, line 1150: Test fixture for spinning up a temp app and tearing it down afterwards

khiron:
I think we're good for merging this PR, and probably merging it into branch: plugin-support which is a better description.  

I have another branch locally working on registering multiple apps with the same name, and testing dynamic app registration for testing that will require code in this PR to be merged.

GavinHuttley (approved):
happy for you to merge this where you feel it should be to facilitate your next patch -- just not into the develop branch yet

PR #1780, opened 2024-03-15
Refactor AnnotationDb to not use SQL keywords for column names, fixes #1750

### LINKED ISSUE #1750: Refactor AnnotationDb to not use SQL keywords for column names
`end` is a SQL keyword but has been embedded in the Sqlite3 based AnnotationDb classes, without raising any errors.

We should convert all methods on the `AnnotationDb` classes / subclasses to replace `end` with `stop`. If they method is public, the argument will need to be deprecated (use the `cogent3.util.warning.deprecate_argument` decorator).
On loading a db from disk, if a column named `end` exists it should be renamed as `stop`.

rmcar17:
Fixes #1750.

Replaces `end` as a column name with `stop` and deprecates related arguments.

On loading a database, columns named `end` are replaced with `stop` (though in conjunction with #1777 old databases would not be able to be loaded due to the new numpy format - maybe that's something that needs to be added to that PR?).


rmcar17:
Codacy is detecting a false positive as the variable, as before, is being used with `locals()`

src/cogent3/core/annotation_db.py
```diff
@@ -448,6 +480,39 @@ def _compatible_schema(db: sqlite3.Connection, schema: dict[str, set]) -> bool:
...
+
+    # Check if the column exists
+    for column_info in table_info:
+        if column_info["name"] == old_column:
+            break
+    else:
+        return  # There is no column to rename
+
+    sql = f"""ALTER TABLE {table_name}
+RENAME COLUMN "{old_column}" TO {new_column}"""
```
GavinHuttley, line 510: why split this across lines? If this is about needing to stay within line limits, you can do the following instead
```python
    sql = (f"ALTER TABLE {table_name} "
           f'RENAME COLUMN "{old_column}" TO {new_column}')
```

src/cogent3/parse/gff.py
```diff
@@ -87,32 +87,32 @@ def _gff_parser(
...
-        attributes = attribute_parser(attributes, (start, end))
+        attributes = attribute_parser(attributes, (start, stop))
 
         yield {
             "SeqID": seqid,
             "Source": source,
             "Type": type_,
             "Start": start,
-            "End": end,
+            "Stop": stop,
```
GavinHuttley, line 115: I don't think we should change the name here, because the GFF3 specs define that column as End. Sorry!

GavinHuttley (changes requested):
Fantastic contribution, thank you! One minor issue and a more significant one. In the [format description](https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md), columns 4 and 5 are defined as "start" and "end". (I've seen them as captialised in other contexts, hence why I've done "Start", "End".) So I think we should retain that naming in the returned dict from the parser for consistency with the standard.

src/cogent3/core/annotation_db.py
rmcar17: Ah, just an old unnecessary habit. Fixed!

src/cogent3/parse/gff.py
rmcar17: Done! The renaming now happens in `_merged_data` of `GffAnnotationDb` just like `"Type"` is renamed to `"biotype"`

PR #1778, opened 2024-03-14
DEP: remove the erf() and erfc() function, fixes #1738

ShunyiYang:

Issue #1738, opened 2024-02-24
mark erf(), erfc(), functions for removal


GavinHuttley:
**What should be changed**

`cogent3.maths.species.erf()`, `cogent3.maths.species.erfc()`

**Why**

Not used in cogent3, available in scipy.stats?

**How much notice**

6 months notice.

**Guidelines for deprecation**

As this is a function that will be removed, see [dev guidelines for deprecation](https://github.com/cogent3/c3dev/wiki/Deprecating-Code) of code. Indicate the limiting version (for when the old function will be removed) as <current year>.<PR month + *x*> where *x* is defined above.

ShunyiYang:
I would like to do this issue.

PR #1782, opened 2024-03-18
DEP: remove the erf() and erfc function fixes #1778

ShunyiYang:
This time, I used GitHub Desktop. Previously, I used the Git command line, but there was a line ending issue: I am on a Windows system, and after downloading cogent3 to my computer, all the files had \r\n line endings instead of \n, which resulted in over a million modifications when compared to the original repository. This problem doesn't occur with GitHub Desktop. However, I still want to know how to solve this issue if using the Git command line.

### DIFF
--- a/src/cogent3/maths/stats/special.py
+++ b/src/cogent3/maths/stats/special.py
@@ -1,8 +1,9 @@
 """Translations of functions from Release 2.3 of the Cephes Math Library,
 (c) Stephen L. Moshier 1984, 1995.
 """
-from cogent3.util import warning as c3warn
 from numpy import exp, floor, log, sin, sqrt
+
+from cogent3.util import warning as c3warn
 
 
 log_epsilon = 1e-6  # for threshold in log/exp close to 1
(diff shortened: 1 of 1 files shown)


ShunyiYang:
Oh, I get it now. After changing the VSCode settings, I need to redownload cogent3 so that VSCode can open files in LF mode.

src/cogent3/maths/stats/special.py
```diff
@@ -295,6 +297,8 @@ def erfc(a):
         return y
 
 
+
```
GavinHuttley, line 300: you should run black and isort against the code. In the root directory for the project
```
$ black .; isort .
```
these ensure formatting, white space etc.. all match the project standards

GavinHuttley (changes requested):
Thanks so much for the contribution!

One small task left before wrapping up your first time contribution to Cogent3 🚀! Run the linters against the code (see comment on the file).

GavinHuttley:
fixes #1738

PR #1788, opened 2024-03-21
TST: add tests for update_file_format

rmcar17:
Stacked on #1787.

Adds tests for the update_file_format function.

Also fixes an issue where update_file_format proceeded without error, creating a new file, when the file to be updated did not exist.


GavinHuttley (approved):
very thorough tests @rmcar17 !

PR #1797, opened 2024-03-29
ENH: improve accessibility of ENS statistic from likelihood functions

GavinHuttley:


GavinHuttley:
codacy issue is the Markdown heading level. Safe to ignore since scriv is doing it's own formatting

Issue #1789, opened 2024-03-21
Enh: allow apps to be referenced by scope to avoid namespace collisions


khiron:
`_get_app_matching_name` (and `get_app`) will take a full scope but only use the app name when requesting app attributes to construct the app instance so get_app will return the first app that matches the name without reference to scope.  

Modify get_app and available_apps and app_help to support scope in the app name, and get the following test to pass

```python
@define_app
class to_json:
    def main(self, data: c3types.SerialisableType) -> str:
        """Convert primitive python types to json string."""
        return json.dumps(data)
    @classmethod
    def _requires_imports(cls):
        return ["from cogent3.app import typing as c3types"]   

from stevedore.extension import ExtensionManager


@pytest.mark.parametrize("install_temp_app", [to_json], indirect=True)
def test_app_namespace_collision(install_temp_app):
    em = ExtensionManager(namespace="cogent3.app", invoke_on_load=False)
    to_json_extensions = [ext for ext in em.extensions if ext.name == "to_json"]
    assert len(to_json_extensions) == 2

    with pytest.raises(NameError):
        to_json = get_app("to_json")

    assert available_apps("to_json").shape[0] == 2

    internal_to_json = get_app("cogent3.app.io.to_json")
    added_to_json = get_app("temp_app_to_json.to_json")
    data = {"a": 1, "b": 2}

    assert internal_to_json(data) == added_to_json(data) == json.dumps(data)
```

khiron:
Create a new `test_plugin.py` for testing plugin apps, that includes the fixture for dynamically loading and unloading test apps.  Move tests into this file from `test_composable.py` that specifically test registration of apps and remove the xfail decoration from these tests.

Test help documentation using a test app and `_make_apphelp_docstring(app)`

PR #1801, opened 2024-04-01
App namespace resolution  #1789

khiron:
Added the ability to dynamically create and register apps for the purpose of testing name space collisions 

Added the ability to mock up the stevedore.extension.ExtensionManager so we can directly put extensions in without having to change the virtual environment's site-packages

Added tests for namespace collision


khiron:
`pkg_resources` has been deprecated for Python 3.12

Unfortunately I am using it to verify when a new package I have pip installed appears to the environment.  Working on an alternative strategy for dynamically loading packages.  For now I'll back out the dynamic installation code as it's not needed to test namespace collision as we can use a mocked ExtensionManager

khiron:
`pkg_resources` can be mostly replaced with `import_lib` with the exception of rebuilding the working set.  

`import_lib.invalidate.caches()` does some of what rebuilding the working set does, but it's not guaranteed to see a change to the environment.  Stevedore apparently switched to `import_lib.metadata` recently which results in the stevedore cache not being updated immediately a new package is installed.  The recommended approach is to restart the python interpreter which is not feasible in the context of cogent3 acquiring the latest packages.

For now invalidate_caches will be sufficient for at least one test.  Which is sufficient.

PR #1803, opened 2024-04-01
Dynamic plugin loading

khiron:
Add back in dynamic plugin loading for the purposes of testing.

Note as this uses `import_lib` the available plugins will be only immediately update for the fist plug_in loaded using this mechanism.  Subsequence `pip install .` commands will only be reflected when `import_lib` regularly invalidates it's caches.


src/cogent3/app/__init__.py
```diff
@@ -52,36 +56,67 @@ def _make_types(app) -> dict:
     return _types
 
 
-def available_apps(name_filter: str | None = None) -> Table:
+# private global to hold an ExtensionManager instance
+__apps = None
+
+
+def apps(force: bool = False) -> stevedore.ExtensionManager:
```
GavinHuttley, line 63: rename to `get_app_manager()`. Otherwise you could have code like `cogent3.apps.apps` which does not convey it's returning an extension manager.

src/cogent3/app/__init__.py
```diff
@@ -52,36 +56,67 @@ def _make_types(app) -> dict:
...
+
+
+def available_apps(name_filter: str | None = None, force: bool = False) -> Table:
     """returns Table listing the available apps
 
     Parameters
     ----------
     name_filter
         include apps whose name includes name_filter
+    force : bool, optional
```
GavinHuttley, line 97: I think leave all type hint info for the function signature, otherwise chance of out-of-sync

tests/test_app/test_install_plugins.py
```diff
@@ -0,0 +1,256 @@
...
+        if is_package_installed(package_name=package_name):
+            break
+        elif time.time() - start_time > timeout:
+            raise TimeoutError(
+                f"Package {package_name!r} not found after waiting for {timeout} seconds."
+            )
+        time.sleep(1)
+
+    # check we can import it
+    timeout = 60  # maximum time to wait in seconds
```
GavinHuttley, line 106: duplicate definition of timeout, and below too...

tests/test_app/test_install_plugins.py
```diff
@@ -0,0 +1,256 @@
...
+        mod = f"mod_{cls.__name__}"
+    # Use subprocess to run pip uninstall -y in the temporary directory
+    package_name = mod.replace("_", "-")  # replace underscores with hyphens
+    check_call([sys.executable, "-m", "pip", "uninstall", "-y", package_name])
+    # force the app cache to reload
+    _ = available_apps(force=True)
+
+
+@contextmanager
+def temp_app(cls, module_name: str = None):
```
GavinHuttley, line 148: Nice!

tests/test_app/test_install_plugins.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Usage
+    -----
+    with temp_app(cls):
+        myapp = get_app(cls.__name__)
+        # test myapp
+    """
+    if module_name is None:
+        module_name = f"mod_{cls.__name__}"
+    temp_dir = mkdtemp()
```
GavinHuttley, line 182: is there a reason you can't use pytest temp_path fixture here so you don't manually create `mkdtemp()`? This approach can be flaky if the default temp directory is very small (e.g. on a VM with a secondary and much larger mount). Use `pytest` so any changes to settings can be done in one place

tests/test_app/test_install_plugins.py
```diff
@@ -0,0 +1,256 @@
...
+    -----
+    The app is installed in a temporary directory, which is deleted after testing.
+
+    Usage
+    -----
+
+    @pytest.mark.parametrize("install_temp_app", [MyAppClass], indirect=True)
+    def test_function(install_temp_app):
+        myapp = get_app('MyAppClass')
+        # test myapp
```
GavinHuttley, line 227: comment line needed?

tests/test_app/test_plugins.py
```diff
@@ -0,0 +1,163 @@
...
+        name=name,
+        entry_point=EntryPoint(
+            name=name, value=f"{module_name}:{name}", group="TESTING"
+        ),
+        plugin=plugin,
+        obj=None,
+    )
+
+
+def test_Install_app_class(mock_extension_manager):
```
GavinHuttley, line 65: style violation! all lower case

tests/test_app/test_plugins.py
```diff
@@ -0,0 +1,163 @@
...
+            return data.upper()
+
+    mock_extension_manager([create_extension(uppercase)])
+
+    appercase = get_app("uppercase")
+    assert appercase("hello") == "HELLO"
+    assert appercase.__doc__ in _make_apphelp_docstring(appercase.__class__)
+
+
+def test_Install_app_function(mock_extension_manager):
```
GavinHuttley, line 80: style!! 👀

tests/test_app/test_plugins.py
```diff
@@ -0,0 +1,163 @@
...
+    """available_apps robust to local scope apps"""
+
+    @define_app
+    def dummy(val: int) -> int:
+        return val
+
+    mock_extension_manager([create_extension(dummy)])
+    apps = available_apps()
+    assert isinstance(apps, Table)
+    apps.filtered(lambda x: dummy.__name__ == x, columns="name")
```
GavinHuttley, line 162: this is having effect. Table's are immutable, so you'd need to be assign the result of the `filtered` call to `apps`

GavinHuttley (changes requested):
Looking very comprehensive and thorough, but it's OK to not so comprehensively `docstring` test functions! Given they'll never be examined using `help(func)`, making it a comment is fine.

tests/test_app/test_install_plugins.py
khiron: That was just laziness on my part to not revert to pytest temp_path.

I was seeing what appeared to be a race condition with stevedore's cache not being updated after the second installation of a temp app.  So I was playing with when I removed the temp directory in case that was the problem.  Turns out the problem was not a race condition, but a change in stevedore to use importlib instead of pkg_resources, and an implicit assumption in the python "use pattern" that the interpreter will be reloaded between package installation.  Which leaves this as a live problem for multiple dynamic loading of apps for testing purposes.

Will change to use pytest temp_path.  Looking into strategies for hiving off dynamic loading of apps that can't be addressed with a mocked ExtensionManager (ie: collision with existing tests) to a special test suite (like we do with "slow" tests).

tests/test_app/test_init.py
```diff
@@ -151,51 +131,14 @@ def test_get_app_kwargs():
...
-    app_help("blah")
-    got = capsys.readouterr().out
-    if app_doc:
-        assert "Overview" in got
-
-    if init_doc:
-        assert "Options" in got
-
-
+@pytest.mark.xfail(reason="Constructing apps on the fly is no longer supported")
```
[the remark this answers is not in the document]
khiron: Yes, I have implemented this in test_plugins using the mocked ExtensionManager

tests/test_app/test_install_plugins.py
khiron: cut and paste antipattern.  I have an idea for a more declarative approach

src/cogent3/app/__init__.py
GavinHuttley, line 1: this is using a private attribute of stevedore, definitely don't do this!

src/cogent3/app/__init__.py
```diff
@@ -60,43 +62,42 @@ def _make_types(app) -> dict:
...
+        __apps = stevedore.ExtensionManager(
+            namespace=APP_ENTRY_POINT, invoke_on_load=False
+        )
+        importlib.invalidate_caches()  # reset the cache of entry points
+        cache = __apps.ENTRY_POINT_CACHE
+        # technique for clearing cache from https://github.com/openstack/cloudkitty/blob/e552c3851fa4bb98096228a89613ceb823d45e72/cloudkitty/utils.py#L198
+        if APP_ENTRY_POINT in cache:
+            del cache[APP_ENTRY_POINT]
+        else:
+            cache.clear()
```
GavinHuttley, line 85: why do this if `APP_ENTRY_POINT` is not present?

tests/test_app/test_install_plugins.py
```diff
@@ -1,256 +1,283 @@
...
+from importlib import metadata
+from inspect import getsourcelines
+from pathlib import Path
+from subprocess import check_call
+
+import pytest
+import stevedore
+
+from stevedore import ExtensionManager
+from stevedore._cache import _get_cache_dir
```
GavinHuttley, line 37: OK, it's in a test so not so vulnerable for users if stevedore changes their API, which they may do without warning for private attributes... There must be a better way. Have you thought of posting a question on their repo on whether have a mechanism for supporting this type of evaluation?

tests/test_app/test_install_plugins.py
```diff
@@ -1,256 +1,283 @@
...
+            lambda: is_entry_point_registered(module_name=mod, app_name=cls.__name__)
+        )
+
+
+def reset_stevedore():
+    importlib.invalidate_caches()
+    importlib.reload(metadata)
+    importlib.invalidate_caches()
+    importlib.reload(stevedore)
+    importlib.reload(stevedore._cache)
```
GavinHuttley, line 171: and again...

GavinHuttley (changes requested):
Getting close, but we can never use private attributes for third party libraries within cogent3 modules. Since they could be changed at any time a user updating stevedore could break their cogent3 install. Better for us to document that users should restart the interpreter or notebook kernel if they install new apps than run the risk of difficult to track down bugs.

I'm less bothered within the tests since cogent3 developers should have the skills to identify the problem but, even then, best to avoid this at all costs and seek an alternate strategy.

src/cogent3/app/__init__.py
khiron: it was an openstack suggestion to handle the feature of implementing dynamic plugins in stevedore.  I was getting desperate at this point and I figgured openstack might  know something I don't about how stevedore works.  Turns out stevedore is just not designed to change it's extensions while the interpreter isn't being reloaded.

khiron:
Closing pull request to pull this into plugin-support instead of develop

PR #1767, opened 2024-03-13
feature/issue 1754

Lizhaozhe123:
Add code example snippet to cogent3.app.sample.trim_stop_codons

PR #1804, opened 2024-04-04
DOC: Docstring examples for cogent3.app.sample.trim_stop_codons

### LINKED ISSUE #1754: Add code example snippet to cogent3.app.sample.trim_stop_codons
The code should be executable by doctesting. See `omit_degenerates()`, `omit_gap_pos()`, `concat()` docstrings for examples. The purpose of these examples is a user gets help on an app and the help displays working code. In other words, the sample code you write should work in an interactive interpreter once the preceding `>>> ` have been removed. It should also not load any data from file, but create sequence collections (in this case) from dictionaries with small sequences.

This app removes trailing stop codons from sequences in a sequence collection or alignment. The output will vary based on whether the input is a sequence collection or an alignment.

For examples, see the `test_trim_stop_codons()` method in `tests/test_app/test_sample.py`. You should only use a minimal data set to demonstrate how it works, not the entire data set used for testing.

fredjaya:
Edits to @Lizhaozhe123 contributions on PR https://github.com/cogent3/cogent3/pull/1767.

Fixes #1754.

### DIFF
--- a/src/cogent3/app/sample.py
+++ b/src/cogent3/app/sample.py
@@ -820,13 +820,13 @@ class omit_duplicated:
 class trim_stop_codons:
     """Removes terminal stop codons."""
 
-    def __init__(self, gc=1):
+    def __init__(self, gc: Union[str, int] = 1):
         """
         Parameters
         ----------
         gc
             identifier for a genetic code or a genetic code instance, defaults
-            to standard genetic code
+            to standard genetic code. See https://cogent3.org/doc/cookbook/what_codes.html
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


src/cogent3/app/sample.py
```diff
@@ -820,18 +820,46 @@ def main(self, seqs: SeqsCollectionType) -> T:
...
+        collection or alignment.
+
+        Create a sample unaligned sequence data and an app to remove trailing
+        stop codons.
+
+        >>> from cogent3 import make_unaligned_seqs, make_aligned_seqs, get_app
+        >>> ualn = make_unaligned_seqs(data={"s1": "AAATTTCCC", "s2": "AAATTTTAA"}, moltype="dna")
+        >>> app = get_app("trim_stop_codons")
+        >>> result = app(ualn)
+        >>> print(result.to_dict())
```
GavinHuttley, line 849: change to pretty_print()

src/cogent3/app/sample.py
```diff
@@ -820,18 +820,46 @@ def main(self, seqs: SeqsCollectionType) -> T:
...
+        Remove trailing stop codons from aligned sequences, using the genetic
+        code for "Vertebrate Mitochondrial".
+
+        For a list of all genetic codes, see
+        https://cogent3.org/doc/cookbook/what_codes.html.
+
+        >>> aln = make_aligned_seqs(data={"s1": "AAATTTCCC", "s2": "AAATTTTAA"}, moltype="dna")
+        >>> app = get_app("trim_stop_codons", gc=2)
+        >>> result = app(aln)
+        >>> print(result.to_dict())
```
GavinHuttley, line 861: as above

GavinHuttley (changes requested):
small tweaks

src/cogent3/app/sample.py
fredjaya: `AttributeError: 'SequenceCollection' object has no attribute 'to_pretty'`. Works for the next example where it's aligned. Suggestions?

fredjaya: `AttributeError: 'SequenceCollection' object has no attribute 'to_pretty'`. Works for the next example that returns an `Aligned` instance. Suggestions?

GavinHuttley: Nah, you're right. `to_dict()` is fine.

PR #1822, opened 2024-04-24
ENH: seqcoll.to_fasta supports block_size

rmcar17:
Addresses #1795


rmcar17:
Looks like Python 3.9 is not supported by setup-python on macos-14 (https://github.com/actions/setup-python/issues/850).
A result of the [rollout to macos-14](https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/), may need to downgrade to macos-13 for now,

ZeroIntensity:
`macos-12` is what `macos-latest` used to point to, maybe it's a better idea to just use that.

rmcar17:
Thanks Zero! Looks like it's been resolved now anyway, so we'll be moving back to latest.

GavinHuttley (approved):
Awesome, thank you!

PR #1827, opened 2024-04-25
API: refactor of map into IndelMap and FeatureMap, fixes #1726

### LINKED ISSUE #1726: Split core.location.Map with specialist classes
Following on from #1690, cogent3.core.location.Map is doing double duty as a map of indels in aligned sequences and of sequence segments represented by features. I will split this class into 2; `IndelMap` and `FeatureMap`. These two classes should no longer record the slice operation history of the associated sequence as this is done by SeqView. Therefore all methods / properties for whether a feature us reversed or not need to be changed so that they derive from the bound Sequence if possible. Although Alignment's may require that be relaxed for FeatureMap.

The original `Map` class will be deprecated.

GavinHuttley:
[NEW] IndelMap uses numpy arrays to store gap locations as gap_pos and cum_gap_lengths.
    It has no memory of slicing, unlike the previous Map class, so is much simpler. It has methods
    for converting sequence indices into alignment indices and vice versa.

[NEW] FeatureMap is largely the same as the original Map class, but also no longer stores history.
    As the name implies, it is intended for Features only.


GavinHuttley:
I note that codacy is identifying two issues. One concerns usage of the name `map` as a variable / attribute. The latter will be resolved when we replace `Aligned.map` and `Aligned.data` with the single aligned data view. This is planned for the Q2 release.

The second concerns a difference in the signature of the `from_spans()` method in a subclass. This method is in place to enable migration from the original Map class. When that class is fully removed, this method will be deprecated.

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        if (
+            self.num_gaps
+            and self.gap_pos[-1] + self.cum_gap_lengths[-1] < self.parent_length
+        ):
+            coords.append((last, self.parent_length))
+        return coords
+
+    def get_gap_coordinates(self) -> SeqCoordTypes:
+        """returns [(gap pos, gap length), ...]"""
+        cum_lengths = self.cum_gap_lengths.copy()
```
rmcar17, line 1673: Reuse get_gap_lengths instead of the first three lines?

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+
+        Parameters
+        ----------
+        coords
+            sequence insert gap coordinates [(gap start, gap end), ...]
+        Returns
+        -------
+
+        """
+        coords = list(sorted(coords))
```
rmcar17, line 1726: sorted already returns a new list

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        )
+
+    def joined_segments(self, coords: SeqCoordTypes) -> "IndelMap":
+        """returns new map with disjoint gapped segments joined
+
+        Parameters
+        ----------
+        coords
+            sequence insert gap coordinates [(gap start, gap end), ...]
+        Returns
```
rmcar17, line 1722: Returns section empty

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+    def nongap(self) -> Iterator[SpanTypes]:
+        """ungappeed segments in this map in aligned coordinates"""
+        # we want to know the coordinates of the ungapped segments on
+        # the aligned sequence. The gap_pos attribute is in sequence
+        # coordinates
+        prev_pos = 0
+        for i, pos in enumerate(self.gap_pos):
+            if pos == 0:
+                # we start with a gap
+                prev_pos = pos
```
rmcar17, line 1594: This is not covered by tests

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+
+            cum_length = 0 if i == 0 else self.cum_gap_lengths[i - 1]
+            start = 0 if i == 0 else prev_pos
+            start += cum_length
+            end = self.gap_pos[i] + cum_length
+            yield Span(start, end)
+            prev_pos = pos
+
+        if self.num_gaps and self.gap_pos[-1] + self.cum_gap_lengths[-1] < len(self):
+            yield Span(prev_pos, len(self))
```
rmcar17, line 1605: This is not covered by tests

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        # using a dict here because joining can produce a gap merge
+        gaps = {}
+        cum_length = 0
+        cum_parent_length = 0
+        dtype = self.gap_pos.dtype
+        for start, end in coords:
+            im = self[start:end]
+            for i in range(im.num_gaps):
+                pos = im.gap_pos[i] + cum_parent_length
+                if pos in gaps:
```
rmcar17, line 1736: Simplify to `gaps[pos] = gaps.get(pos, cum_length) + im.cum_gap_lengths[i]`?

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+
+        Notes
+        -----
+        discards reverse attribute on both spans and self
+        """
+        new_pos = self.gap_pos.copy()
+        lengths = self.get_gap_lengths()
+
+        if len(new_pos):
+            new_pos = self.parent_length - new_pos
```
rmcar17, line 1765: This feels off to me, `self.parent_length` refers to the number of non-gaps no?

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        """returns sequence coordinates of ungapped segments
+
+        Returns
+        -------
+        [(start, end), ...]
+        """
+        coords = []
+        last = 0
+        for pos, cum_length in zip(self.gap_pos, self.cum_gap_lengths):
+            pos, cum_length = int(pos), int(cum_length)
```
rmcar17, line 1657: `cum_length` unused

rmcar17 (changes requested):

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        Parameters
+        ----------
+        align_feature_map
+            with alignment coordinates
+        include_gaps
+            whether to include gaps from self as LostSpan's
+        """
+        gap_spans = {}
+        if include_gaps and self.num_gaps:
+            last = 0
```
rmcar17, line 1828: This section is not covered by any tests

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+            for pos, cum_length in zip(self.gap_pos, self.cum_gap_lengths):
+                gap_spans[pos] = LostSpan(cum_length - last)
+                last = cum_length
+        spans = []
+        for span in align_feature_map.spans:
+            if span.lost:
+                spans.append(span)
```
rmcar17, line 1835: This section is not covered by any tests

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+                continue
+
+            start = self.get_seq_index(span.start)
+            end = self.get_seq_index(span.end)
+            if lost := gap_spans.pop(span.start, None):
+                spans.append(lost)
```
rmcar17, line 1841: This section is not covered by any tests

src/cogent3/app/align.py
```diff
@@ -251,7 +251,8 @@ def _gaps_for_injection(other_seq_gaps: dict, refseq_gaps: dict, seqlen: int) ->
     # sequence coordinates
     # we probably need to include the refseq gap union because we need to
     # establish whether a refseq gap overlaps with a gap in other seq
-    # and
+    # todo convert these functions to using IndelMap and the numpy set
```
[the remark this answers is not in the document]
GavinHuttley: yep

src/cogent3/core/location.py
```diff
@@ -1099,7 +1100,1120 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        ...
+
+
+def _spans_from_locations(locations, parent_length) -> Tuple[Union[Span, _LostSpan]]:
+    if not len(locations):
+        # using len() because locations can be a numpy array
+        return ()
+
+    if locations[0][0] > locations[-1][1]:
+        raise ValueError("locations must be ordered smallest-> largest")
```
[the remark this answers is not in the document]
GavinHuttley: not used elsewhere

src/cogent3/core/location.py
```diff
@@ -1099,7 +1100,1120 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        seq_length = aligned_length - cum_lens[-1]
+
+        return cls(
+            gap_pos=gap_pos,
+            cum_gap_lengths=cum_lens,
+            parent_length=seq_length,
+        )
+
+    @functools.singledispatchmethod
+    def __getitem__(self, item):
```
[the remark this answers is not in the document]
GavinHuttley: cannot have a forward reference (ie str) on a singledispatch decorated method, get a NameError

src/cogent3/core/location.py
```diff
@@ -1099,7 +1090,1101 @@ def RangeFromString(string, delimiter=","):  # pragma: no cover
...
+        gap_data = numpy.array([self.gap_pos, self.cum_gap_lengths]).T
+        return f"{gap_data.tolist()!r}/{self.parent_length}"
+
+    def get_gap_lengths(self) -> IntArrayTypes:
+        lengths = self.cum_gap_lengths.copy()
+        lengths[1:] = numpy.diff(lengths)
+        return lengths
+
+    @property
+    def offsets(self) -> IntArrayTypes:
```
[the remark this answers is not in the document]
GavinHuttley: deleting this method and associated tests since its not used elsewhere

src/cogent3/core/location.py
GavinHuttley: When we reverse a nucleic acid sequence, the gaps shift. For example with
```
#   01  2
seq AC--G
```
the gap is at sequence index 2. Now when we reverse (and I'm doing the reverse complement of the DNA sequence for "completeness"
```
#   0  12
seq C--GT
```
The gap sequence position is now 3-2=1

rmcar17: What about a case like:
```
#   0 1 2
seq A-A-A
```
which would have gap positions at 1 and 3. We would expect the gap positions of the reverse complement
```
#   0 1 2
seq T-T-T
```
to still be 1 and 3, but the current algorithm would return 3-3=0 and 3-1=2 right?

I got the impression that this function should rely on the length of the entire sequence, as well as the length of the gaps (as we care about the position of the last character in the gap, not the first).

rmcar17: As another example for
```
-A-AA----A
```
we should expect [1, 7, 9], but I believe we would instead get [-1, 2, 4]

rmcar17: As another example, for

```
#    0 12    3
seq -A-AA----A
```

we should expect [1,7,9] but I believe we end up getting [-1,2,4]

rmcar17: Disregard the above, I was indexing the gap positions incorrectly. For the first, the gap positions are originally [1,2] not [1,3]. For the second it is [0, 1, 3] for the original gap positions, not [0, 2, 5]

PR #1843, opened 2024-05-01
ENH: apps now work as intended for the `|` union syntax

rmcar17:


GavinHuttley (changes requested):
we need a test that breaks on the original code

rmcar17:
Added tests for Unions constructed with `|`, seems to only have been an issue re: the old code with builtin types.

GavinHuttley (approved):
Awesome, thanks!

Issue #1844, opened 2024-05-02
`doctest-modules` doesn't recognise docstring examples in decorated functions


fredjaya:
Running ``pytest --doctest-modules -v`` doesn't find docstrings in functions decorated with `@define_app` (e.g. `app.align.cogent3_score`). 

However, it works fine in `@define_app` classes where the docstrings are in the ``__init__`` (e.g. `app.align.progressive_align`).

Specfically, `composable._class_from_func` does not seem to be correctly propagating docstrings.

PR #1847, opened 2024-05-02
add test to check for #1844

khiron:
This test verifies that a @define_app decorated function generates apphelp from the functions docstring


khiron:
Small PR to add a test to verify that @define_app decorated functions generate help containing text from the function's  docstring.

Sorry to add to the load

PR #1850, opened 2024-05-02
Fix `__repr__` handling of limit in Alignment/SeqCol related objects

rmcar17:
Also adds tests for each object's `__repr__`


tests/test_core/test_alignment.py
```diff
@@ -3738,8 +3744,63 @@ def test_sequence_collection_repr():
...
+        "d": "D" * 11,
+        "e": "E" * 11,
+    }
+    seqs = cls(data=data, moltype=ASCII)
+    assert (
+        repr(seqs)
+        == "5 x 11 text alignment: a[AAAAAAAAAA...], b[BBBBBBBBBB...], c[CCCCCCCCCC...], ..."
+    )
+
+    if (
```
GavinHuttley, line 3801: All sequence collection types will soon not allow construction with empty, delete this final clause

GavinHuttley (changes requested):

src/cogent3/core/alignment.py
```diff
@@ -564,6 +564,11 @@ def __init__(
         if isinstance(data, (typing.Generator, typing.Iterator)):
             data = tuple(data)
 
+        if len(data) == 0:
```
rmcar17, line 567: I assume the conversion above ensures that data has a `__len__` attribute, though I note that the coercion in `conversion_f` checks again for whether it is a Generator/Iterator.

GavinHuttley: Correct and the redundancy in this code will be addressed with the "big rewrite" of backends to sequence collections.

GavinHuttley (approved):
did the trick!

PR #1864, opened 2024-05-14
Mixture of patches

GavinHuttley:


src/cogent3/core/annotation_db.py
```diff
@@ -1275,18 +1275,19 @@ def add_records(self, reduced: dict) -> None:
...
             # our Feature code assumes start always < stop,
             # we record direction using Strand
+            record = reduced[key]
             spans = numpy.array(sorted(record["spans"]), dtype=int)  # sorts the rows
             spans.sort(axis=1)
             record["start"] = int(spans.min())
             record["stop"] = int(spans.max())
             record["spans"] = spans
-            rows.append(tuple(record.get(c) for c in col_order))
+            self.db.execute(sql, tuple(record.get(c) for c in col_order))
```
rmcar17, line 1288: lgtm. Though, I wonder about the performance implications of many individual inserts rather than one large batched insert when there are many rows. Potentially could have some kind of batch size, though likely overcomplicating it for now.

src/cogent3/util/io.py
```diff
@@ -343,8 +344,81 @@ def remove_files(list_of_filepaths, error_on_missing=True):
...
-    with contextlib.suppress(ValueError, TypeError):
-        return os_path.exists(str(path))
+    with contextlib.suppress(Exception):
+        return Path(path).exists()
     return False
+
+
+def iter_splitlines(
+    path: PathType, chunk_size: Optional[int] = 1_000_000
+) -> Generator[str, None, None]:
```
rmcar17, line 356: Fine as it is, [docs](https://docs.python.org/3/library/typing.html#typing.Generator) list that when there is no `SendType` or `ReturnType`, it could also be simplified to either `Iterator[YieldType]` or `Iterable[YieldType]`

src/cogent3/util/io.py
```diff
@@ -343,8 +344,81 @@ def remove_files(list_of_filepaths, error_on_missing=True):
...
+    with open_(path) as infile:
+        last = ""
+        while True:
+            data = infile.read(chunk_size)
+            if not data:  # end of file
+                break
+
+            data = last + data
+            # we check for POSIX or Windows line endings
+            end_is_newline = data.endswith(("\n", "\r\n"))
```
rmcar17, line 385: Should be equivalent to just `data.endswith("\n")`

src/cogent3/core/annotation_db.py
GavinHuttley: True, if we wind up doing batches of lines, this change will not be worth it. I'll revert that.

src/cogent3/util/io.py
GavinHuttley: if you read in all text from a text file that was created on Windows (thus "\r\n" endings), on Windows what line endings are in the text?

rmcar17: It's complicated. Though in either case a line which ends with "\r\n" still ends with "\n" so there is redundancy.

rmcar17:
Looks good to me!

PR #1870, opened 2024-05-17
NEW: abstract base class for views, fixes #1865

### LINKED ISSUE #1865: The New SeqView foundation
Do this in the `develop` branch in current location for `SeqView` class (i.e. `core.sequence`).

An abstract base class foundation for the different types of view. Core attributes for `SliceRecordABC` are.

```
("start", "stop", "step", "_offset", "_seq_len")
```

Needs the following as concrete methods:

- [x] `absolute_position`
- [x] `relative_position`
- [x] ALL methods involved with `__getitem__`
  - [x]  `__getitem__`
  - [x] all related methods

Needs concrete properties for:

- [x] `is_reversed` (just renaming the `reverse` property)
- [x] `parent_start`
- [x] `parent_stop`

All attributes beginning with `_` need a corresponding concrete setter / getter properties without the `_`.
- [x] `_offset` setter and getter
- [x] `_seq_len` setter only at this point

Misc:
- [x] ensure wording in docstring remains consistent with the abstraction

### Updating current SeqView

Inherits from `SliceRecordABC`, deleting methods that are in the ABC but retaining special attributes and methods.

This issue is finished when all current tests pass!

KatherineCaley:
added abstract base class `SliceRecordABC`, for recording the history of operations to be applied to some underlying data. Modified `SeqView` to inherit from `SliceRecordABC`

### DIFF
--- a/src/cogent3/core/sequence.py
+++ b/src/cogent3/core/sequence.py
@@ -2074,7 +2074,7 @@ class SliceRecordABC(ABC):
             )
 
         if segment.start is segment.stop is segment.step is None:
-            return self.copy(sliced=False)
+            return self.copy()
 
         if len(self) == 0:
             return self
@@ -2402,7 +2402,7 @@ class SeqView(SliceRecordABC):
         sv = cls(**init_args)
         return sv
 
-    def copy(self, sliced=False):
+    def copy(self, sliced: bool = False):
         """returns copy
 
         Parameters
--- a/src/cogent3/core/sequence.py
+++ b/src/cogent3/core/sequence.py
@@ -1912,8 +1912,9 @@ class SliceRecordABC(ABC):
 
     Notes
     -----
-    seq_len refers to a Python typing.Sequence object, e.g. array, str, list, etc
+    seq_len refers to a Python typing.Sequence object, e.g. array, str, list.
     """
+
     start: int
     stop: int
     step: int
@@ -1921,11 +1922,23 @@ class SliceRecordABC(ABC):
     _offset: int
 
     @abstractmethod
+    def _get_init_kwargs(self) -> dict:
+        """return required arguments for construction that are unique to the 
+        subclass"""
+        ...
+
+    @abstractmethod
     def copy(self): ...
 
     @property
     @abstractmethod
     def _zero_slice(self): ...
+
+    @abstractmethod
+    def to_rich_dict(self): ...
+
+    @abstractmethod
+    def from_rich_dict(self): ...
 
     @property
     def offset(self) -> int:
@@ -1986,8 +1999,8 @@ class SliceRecordABC(ABC):
 
         Returns
         -------
-        the absolute index with respect to the coordinates of the self
-        including offset
+        the absolute index with respect to the coordinates of the original 
+        sequence (including offset if present).
         """
         if not self:
             return 0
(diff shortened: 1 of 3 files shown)


src/cogent3/core/sequence.py
```diff
@@ -1906,35 +1906,35 @@ def _input_vals_neg_step(seqlen, start, stop, step):
...
+    step: int
+    _seq_len: int
+    _offset: int
+
+    @abstractmethod
+    def copy(self): ...
+
+    @property
+    def _zero_slice(self):
+        return self.__class__(seq=type(self.seq)())
```
GavinHuttley, line 1933: Make this an abstract property and leave it empty in this abstract base class. For example...
```python
@property
@abstractmethod
def _zero_slice(self): ...
```

src/cogent3/core/sequence.py
```diff
@@ -2017,26 +2009,29 @@ def absolute_position(self, rel_index: int, include_boundary=False):
...
+        if self.is_reversed:
             abs_index = offset + self.seq_len + seq_index + 1
         else:
             abs_index = offset + seq_index
 
         return abs_index
 
     def relative_position(self, abs_index, stop=False):
-        """converts an index relative to annotation coordinates to be with respect to the current sequence view
+        """converts an index relative to annotation coordinates to be with
```
GavinHuttley, line 2020: drop the reference to annotations. This is about converting an index on the original "python sequence" into an index on this "view"

src/cogent3/core/sequence.py
```diff
@@ -2056,6 +2051,48 @@ def relative_position(self, abs_index, stop=False):
...
         return rel_pos
 
+    def __getitem__(self, segment, **kwargs):
+        """subclasses requiring unique arguments for construction should implement
+        ``__getitem__`` proving unique args as keyword args, e.g.,
+
+        .. code-block:: python
+
+            def __getitem__(self, index):
+                subclass_uniq = {..., ...}
```
GavinHuttley, line 2061: that looks like a set, change to
```python
subclass_uniq = {"<attr name>": <attr_value>, ...}
```

src/cogent3/core/sequence.py
```diff
@@ -2119,25 +2158,27 @@ def _get_forward_slice_from_forward_seqview_(self, slice_start, slice_stop, step
 
         # if -ve, it's an invalid slice
         if start < 0 or stop < 0:
-            return _zero_slice
+            return self._zero_slice
```
GavinHuttley, line 2161: change all uses to `self._zero_slice()`

src/cogent3/core/sequence.py
```diff
@@ -2245,50 +2290,56 @@ def _get_reverse_slice_from_reverse_seqview_(self, slice_start, slice_stop, step
...
-        if slice_step > 0:
-            return self._get_slice(segment, slice_step)
-        elif slice_step < 0:
-            return self._get_reverse_slice(segment, slice_step)
-        else:
-            raise ValueError(
-                f"{self.__class__.__name__} cannot be sliced with a step of 0"
-            )
+    def __getitem__(self, segment):
+        seqview_unique = {"seqid": self.seqid}
```
GavinHuttley, line 2341: add "seq" to this unique attribute dict

GavinHuttley (changes requested):
Great start! Channeling my inner Robvert "type hints throughout"..

More broadly I don't think the abstract base class should have a reference to a sequence object. It should be decoupled from whatever it is that the slice will be applied to. That's the job of the subclasses.

src/cogent3/core/sequence.py
KatherineCaley: @GavinHuttley, your above comment indicates that `_zero_slice` is a property

KatherineCaley:
fyi codacy is yelling at `def __getitem__(self, segment: Union[int, slice], kwargs: dict):` for having two params

GavinHuttley (approved):
Awesome! Thank you!!

PR #1876, opened 2024-05-24
DEV: Update `SeqDataView` with `SliceRecordsABC` 

fredjaya:
Draft PR to get current progress reviewed before going further. Mainly changes to `SeqData` and `SeqDataView` to be compatible with the `SeqView`/`SliceRecordABC` refactor.

A preliminary attempt at adding `SeqData._set_seq_maker`.

Lots of tests required for `SeqData` after copying over `SeqView` methods, and for `SeqData.__getitem__` when `self._seq_maker` is provided.


src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+    def __str__(self):
+        return self.str_value
+
+    def __array__(self):
+        return self.array_value
+
+    def __bytes__(self):
+        return self.bytes_value
+
+    def __len__(self):
```
KatherineCaley, line 179: candidate to move to `SliceRecordABC`

tests/test_core/test_new_alignment.py
```diff
@@ -0,0 +1,420 @@
...
+
+@pytest.mark.parametrize("idx", (0, 1))
+def test_getitem_int(simple_dict, alpha, idx):
+    sd = SeqData(simple_dict, alphabet=alpha)
+    got = sd[idx]
+    assert got.seq == sd
+    assert got.seqid == list(simple_dict)[idx]
+
+
+# @pytest.mark.parametrize("seqid", ("seq1", "seq2"))
```
KatherineCaley, line 254: can mark xfail if failing or delete, don't want to accidentally commented out code to sneak into develop

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+        self._alphabet = alphabet
+        self._name_order = process_name_order(data, name_order)
+        self._seq_maker = seq_maker
+        # Convert to moltype alphabet indicies
+        self._data: dict[str, numpy.ndarray] = {
+            k: seq_index(v, self._alphabet) for k, v in data.items()
+        }
+
+    @property
+    def set_seq_maker(self) -> type:
```
KatherineCaley, line 262: call this property `seq_maker`, also not sure about the return type being type - if` _seq_maker` is a callable then it could be `Callable`?

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+        return self._seq_maker
+
+    @set_seq_maker.setter
+    def set_seq_maker(self, value: tuple[type, CharAlphabet]) -> type:
```
KatherineCaley, line 266: also change name to `seq_maker` here, and this is not returning anything so drop the type hint for returned value

```py 
@seq_maker.setter
    def seq_maker(self, value: tuple[type, CharAlphabet]):
``

File: src/cogent3/core/new_alignment.py [hunk 5]
```diff
@@ -0,0 +1,449 @@
...
+        if alphabet != self._alphabet:
+            raise ValueError("Alphabets must be the same")
+        assert alphabet == self._alphabet
+        self._seq_maker = seq_maker
+
+    def _make_seq(self, seqid) -> Union[SeqDataView, type]:
+        seqdataview = self.get_seq_view(seqid=seqid)
+        if self._seq_maker is None:
+            return seqdataview
+        return self.set_seq_maker(seqdataview, seqid=seqid)
```
KatherineCaley, line 277: ` return self.seq_maker(seqdataview, seqid=seqid)`

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+    def iter_seq_view(self, *, name_order: tuple[str] = None) -> Iterator:
+        # Should this output SeqView or SeqDataView?
+        seqids = process_name_order(self._name_order, name_order)
+        for seqid in seqids:
+            yield self.get_seq_view(seqid=seqid)
+
+
+@singledispatch
+def seq_to_gap_coords(
+    seq: Union[str, numpy.ndarray], moltype: MolType
```
KatherineCaley, line 323: im confused, dont you implement a dispatch for both `str` and `numpy.ndarray` below?

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+    gap_pos = parent_coords.T[0] - numpy.append(0, cum_lengths[:-1, numpy.newaxis])
+
+    indel_map = IndelMap(
+        gap_pos=gap_pos, cum_gap_lengths=cum_lengths, parent_length=parent_len
+    )
+
+    return ungapped, indel_map
+
+
+class AlignedDataView(SeqDataView):
```
KatherineCaley, line 384: I think this should inherit from `SliceRecordABC`, but not confident

fredjaya: Yep I think so too, but leaving AlignedData/View until SeqData is (mostly) finished

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+
+    @__getitem__.register
+    def _(self, value: int) -> SeqDataView:
+        seqid = self._name_order[value]
+        return self._make_seq(seqid=seqid)
+
+    def get_seq_array(
+        self, *, seqid: str, start: int = None, stop: int = None
+    ) -> numpy.ndarray:
+        return self._data[seqid][start:stop]
```
KatherineCaley, line 295: Since `get_seq_array` is a class method, would it be better not to require all the args to be provided (like `stop` and `start`) and rely on `self.stop`, etc, or do we foresee wanting flexible indexing?

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+
+@dataclass
+class SeqData:
+    __slots__ = ("_data", "_alphabet", "_name_order", "_seq_maker")
+
+    def __init__(
+        self,
+        data: dict[str, SeqTypes],
+        alphabet: CharAlphabet,
+        name_order: Optional[Union[tuple[str], None]] = None,
```
GavinHuttley, line 250: Refactor throughout, changing `name_order` to `names` and assign to `_names`.

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+        return self._alphabet.from_indices(
+            self.get_seq_array(seqid=seqid, start=start, stop=stop)
+        )
+
+    def get_seq_bytes(
+        self, *, seqid: str, start: int = None, stop: int = None
+    ) -> bytes:
+        return self.get_seq_str(seqid=seqid, start=start, stop=stop).encode("utf8")
+
+    def iter_names(self, *, name_order: tuple[str] = None) -> Iterator:
```
GavinHuttley, line 307: change to
```python
    @property
    def names(self) -> Iterator[str]:
        yield from self._names
```

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+        yield from process_name_order(self._name_order, name_order)
+
+    def get_seq_view(self, seqid: str) -> SeqDataView:
+        seq_len = len(self._data[seqid])
+        return SeqDataView(seq=self, seqid=seqid, seq_len=seq_len)
+
+    def iter_seq_view(self, *, name_order: tuple[str] = None) -> Iterator:
```
GavinHuttley, line 314: Delete this method for now. (I suspect we may get it "for free" having implemented `__getitem__`)

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+        if self._seq_maker is None:
+            return seqdataview
+        return self.set_seq_maker(seqdataview, seqid=seqid)
+
+    @singledispatchmethod
+    def __getitem__(self, value: Union[str, int]) -> SeqDataView:
+        raise NotImplementedError(f"__getitem__ not implemented for {type(value)}")
+
+    @__getitem__.register
+    def _(self, value: str) -> SeqDataView:
```
GavinHuttley, line 284: This is the method that needs to use `self.make_seq` if defined!
```python
@__getitem__.register
def _(self, index: str) -> Union[SeqDataView, Sequence]:
    sv = self.get_seq_view(seqid=index)
    if self._make_seq is None:
        return sv
    return self.make_seq(sv, seqid=seqid)
```

src/cogent3/core/new_alignment.py
GavinHuttley: Actually, let's use the name we've used elsewhere, which is `make_seq`. Change throughout.

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+
+    @set_seq_maker.setter
+    def set_seq_maker(self, value: tuple[type, CharAlphabet]) -> type:
+        seq_maker, alphabet = value
+        if alphabet != self._alphabet:
+            raise ValueError("Alphabets must be the same")
+        assert alphabet == self._alphabet
+        self._seq_maker = seq_maker
+
+    def _make_seq(self, seqid) -> Union[SeqDataView, type]:
```
GavinHuttley, line 273: The code in this method should be in the `__getitem__` method for the `str` type

src/cogent3/core/new_alignment.py
GavinHuttley: Agree, we don't worry about Aligned stuff yet.

GavinHuttley (changes requested):
Thanks for your comments @KatherineCaley !
Some big picture comments first. A few name refactors to make attributes / methods consistent with naming patterns elsewhere. Suggest changing those throughout and making sure your tests pass as well as before. The next step is a refactor of a few methods (largely simplification, or outright deletion). Let me know if things aren't clear!

src/cogent3/core/new_alignment.py
fredjaya: Yes, type hinted the base function to indicate the supported types

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+    def __array__(self):
+        return self.array_value
+
+    def __bytes__(self):
+        return self.bytes_value
+
+    def __len__(self):
+        return abs((self.start - self.stop) // self.step)
+
+    def __iter__(self):
```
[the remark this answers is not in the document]
fredjaya: Specifically for `__iter__`:
- Should it return only one type? (thinking iter(self.str_value))
- Return one of str/array/bytes according to e.g. an argument supplied by the user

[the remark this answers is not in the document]
KatherineCaley: If it matters, iteration over an `array` is much faster than iteration over a `str`.

src/cogent3/core/new_alignment.py
```diff
@@ -0,0 +1,449 @@
...
+        self.step = step
+        self._offset = offset
+        self._seqid = seqid
+
+    def _checked_seq_len(self, seq_len) -> int:
+        assert seq_len is not None
+        return seq_len
+
+    @property
+    def _zero_slice(self):
```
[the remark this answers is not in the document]
fredjaya: Shouldn't `self.seq_len` always be the length of the original/parent sequence?

`start=0, end=0` already added so that the `str` representation of a `SeqData._zero_slice matches the `SeqView.seq = ''` equivalent (i.e. all str-representation slices go through `SeqDataView.str_value()`)

tests/test_core/test_new_alignment.py
```diff
@@ -0,0 +1,436 @@
...
+
+@pytest.mark.parametrize("idx", (0, 1))
+def test_getitem_int(simple_dict, alpha, idx):
+    sd = SeqData(simple_dict, alphabet=alpha)
+    got = sd[idx]
+    assert got.seq == sd
+    assert got.seqid == list(simple_dict)[idx]
+
+
+@pytest.mark.xfail
```
KatherineCaley, line 269: could add parametrization for `make_seq`

```python 
@pytest.mark.xfail(reason="'SeqDataView' object has no attribute 'replace'")
@pytest.mark.parametrize("idx", (0, 1))
@pytest.mark.parametrize("make_seq", (True, False))
def test_getitem_int(simple_dict, alpha, idx, make_seq):
    sd = SeqData(simple_dict, alphabet=alpha)
    ms = get_moltype("dna").make_seq if make_seq else None
    sd.make_seq = ms
    got = sd[idx]
    assert got.seq == sd
    assert got.seqid == list(simple_dict)[idx]


@pytest.mark.xfail(reason="'SeqDataView' object has no attribute 'replace'")
@pytest.mark.parametrize("seqid", ("seq1", "seq2"))
@pytest.mark.parametrize("make_seq", (True, False))
def test_getitem_str(sd_demo, seqid, make_seq):
    ms = get_moltype("dna").make_seq if make_seq else None
    sd_demo.make_seq = ms
    got = sd_demo[seqid]
    assert got.seq == sd_demo
    assert got.seqid == seqid
```

GavinHuttley (approved):
Thanks @fredjaya ! We will work off this. Onto the next steps...

Issue #1867, opened 2024-05-15
begin new_alignment.SequenceCollection


GavinHuttley:
Waiting on #1866

- [x] Create a SequenceCollection class
- [x] SequenceCollection(seq_data=<seq data instance>, moltype=..)
- [x] type hints the constructor
- [x] Assigns seq_data to a `_seq_data` attribute
- [x] Assigns `self._seq_data.seq_maker = self.moltype.make_seq`
- [x] Implements a `seqs` property which returns the `_seq_data`
- [x] Ports relevant tests from `test_alignment.py` into `test_new_alignment.py`, converting them into pytest style (sorry!)
- [x] Implements a `named_seqs` property, which just returns `.seqs` . Deprecate this method and point to just using `.seqs` (give this 12 months).

KatherineCaley:
@GavinHuttley will the new class be inheriting from `_SequenceCollectionBase`?

GavinHuttley:
No. `_SequenceCollectionBase` is an artefact of needing core methods on both `SequenceCollection` (from which `Alignment` directly inherits) and `ArrayAlignment` (which inherits directly from `_SequenceCollectionBase`).

In `new_alignment.py`, define a `SequenceCollection` class without any reference to `alignment.py` classes. All it has to begin with is a constructor and `.seqs` attribute.

GavinHuttley:
closing this as there are now specific issues that are more granular

PR #1881, opened 2024-05-30
Start on SequenceCollections refactor

KatherineCaley:
working towards #1867

questions wrt direction are commented in place


src/cogent3/core/new_alignment.py
```diff
@@ -199,7 +208,10 @@ def to_rich_dict(self):
         else:
             start, stop = self.start, self.stop
```
KatherineCaley, line 210: line 213 changes `.seq` from a `SeqData` instance, to a `str`. My instinct is that it should be a new `SeqData` instance where the seq corresponding to `seqid` is sliced

src/cogent3/core/new_alignment.py
```diff
@@ -357,6 +369,101 @@ def _(seq: numpy.ndarray, moltype: MolType) -> tuple[str, IndelMap]:
...
+    label_to_name: Callable = None,
+    info: dict = None,
+    source: Union[str, Path] = None,
+    **kwargs,
+):
+    """Initialize an unaligned collection of sequences.
+
+    Parameters
+    ----------
+    data
```
KatherineCaley, line 422: let's decide what forms of `data` we are going to support being passed to `make_unaligned_seqs`. currently I am supporting: 
- `dict{name: seq, ...}`
- `[[name, seq]...]` (lists or sets) 
- `[seq, ...]` (lists or sets) 
- `2D numpy array`
which of these are actually necessary and which have I missed?

src/cogent3/core/new_alignment.py
```diff
@@ -357,6 +369,101 @@ def _(seq: numpy.ndarray, moltype: MolType) -> tuple[str, IndelMap]:
...
+    **kwargs
+        other keyword arguments to be passed to SequenceCollection
+    """
+    moltype = get_moltype(moltype)
+    alphabet = moltype.alphabet
+
+    if len(data) == 0:
+        raise ValueError("data must be at least one sequence.")
+    prepped_data = prep_for_seq_data(data, alphabet)
+
```
KatherineCaley, line 443: naming consistency (using `validate_names`) is currently the responsibility of `SeqData` but I think it should be moved into the `SequenceCollection`

src/cogent3/core/new_alignment.py
KatherineCaley: Should we support creating a `SequenceCollection` where data is already a `SeqData` instance?

tests/test_core/test_new_alignment.py
```diff
@@ -9,6 +9,9 @@
     AlignedDataView,
     SeqData,
     SeqDataView,
+    SequenceCollection,
+    _names_list_tuple,
+    make_unaligned_seqs,
     seq_index,
```
GavinHuttley, line 15: Change all imports to use the new versions of the MolType sequence and alphabet modules.

src/cogent3/core/new_alignment.py
```diff
@@ -357,6 +369,101 @@ def _(seq: numpy.ndarray, moltype: MolType) -> tuple[str, IndelMap]:
...
+    seq_data = SeqData(data=prepped_data, alphabet=alphabet)
+
+    return SequenceCollection(seq_data=seq_data, moltype=moltype)
+
+
+class SequenceCollection:
+    def __init__(self, seq_data: SeqData, moltype: MolType):
```
GavinHuttley, line 450: We should require these to be keyword arguments. So `(self, *, ...)`

GavinHuttley (changes requested):
Great progress!
I'm going to get all radical here and suggest that we rename `SeqData` to `SeqsData` because it can contain many sequences.
I also need to update the testing code so that it uses the new versions that I just pushed in for molecular type, alphabets and sequence.

GavinHuttley (changes requested):
Great progress, and some good issues raised. A few small comments to address small comments to address.

src/cogent3/core/new_alignment.py
```diff
@@ -222,49 +172,40 @@ def from_rich_dict(cls, data: dict):
...
-                seq=self.seq,
-                start=self.start,
-                stop=self.stop,
-                step=self.step,
-                offset=self.offset,
-                seqid=self.seqid,
-                seq_len=self.seq_len,
-            )
-        return self.from_rich_dict(self.to_rich_dict())
+        return self.__class__(
```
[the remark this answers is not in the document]
KatherineCaley: @GavinHuttley can you dumb it down to me what you mean by "just return self" - is it in the context of supporting sliced, or to replace the current code?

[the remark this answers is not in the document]
GavinHuttley: Sorry, too brief a comment! I meant the following
```python
def copy(self):
    return self
```

PR #1884, opened 2024-06-04
TST: Port `SeqView` tests and ruff fixes

fredjaya:
Updated tests use e.g. `new_moltype.DNA.make_seq()` replacing `Sequence(...)`.

### DIFF
--- a/src/cogent3/core/moltype.py
+++ b/src/cogent3/core/moltype.py
@@ -964,14 +964,14 @@ class MolType:
         """Returns True if sequence contains no items that are not in self."""
         try:
             return self.first_invalid(sequence) is None
-        except Exception:
+        except:
             return False
 
     def is_strict(self, sequence):
         """Returns True if sequence contains only items in self.alphabet."""
         try:
             return (len(sequence) == 0) or (self.first_non_strict(sequence) is None)
-        except Exception:
+        except:
             return False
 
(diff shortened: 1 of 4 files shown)


tests/test_core/test_sequence.py
```diff
@@ -2596,6 +2596,7 @@ def test_gapped_by_map_segment_iter():
     moltype = get_moltype("dna")
     m, seq = moltype.make_seq("-TCC--AG").parse_out_gaps()
     g = list(seq.gapped_by_map_segment_iter(m, allow_gaps=True, recode_gaps=False))
+    assert g == ["-", "TCC", "--", "AG"]
```
fredjaya, line 2599: Test was missing an assertion! Not 100% sure if this is correct

GavinHuttley: I think that's correct!

src/cogent3/core/moltype.py
```diff
@@ -964,14 +964,14 @@ def is_valid(self, sequence):
         """Returns True if sequence contains no items that are not in self."""
         try:
             return self.first_invalid(sequence) is None
-        except:
+        except Exception:
```
GavinHuttley, line 967: ignore changes to modules being deprecated, but do make those changes in the `new_<module name>`!

GavinHuttley (approved):

PR #1886, opened 2024-06-04
DEV: more refactoring of SequenceCollection

KatherineCaley:


src/cogent3/core/new_alignment.py
```diff
@@ -297,25 +310,62 @@ def __init__(
         *,
         seqs_data: SeqsData,
         moltype: new_moltype.MolType,
+        names: list[str] = None,
```
GavinHuttley, line 313: Let's add `annotation_db` as an optional argument to the `SequenceCollection` constructor, and also add it to the `make_unaligned_seqs()` function. The default value should be configured the same way as it is in `alignment.py` (ie `DEFAULT_ANNOTATION_DB = BasicAnnotationDb`)

src/cogent3/core/new_alignment.py
```diff
@@ -297,25 +310,62 @@ def __init__(
...
         self.info = info
         self._repr_policy = dict(num_seqs=10, num_pos=60, ref_name="longest", wrap=60)
+        self._annotation_db = None
 
     @property
     def names(self) -> Iterator[str]:
         return self._names
 
+    @names.setter
+    def names(self, names: list[str]):
```
GavinHuttley, line 331: add a `# refactor: design` comment (or something better if you can think of it). I'm not sure if this will be a good idea and I know I suggested it, but definitely worth trialing.

src/cogent3/core/new_alignment.py
```diff
@@ -356,35 +395,72 @@ def take_seqs(
...
-            selected_seqs = {
-                name: self.seqs.get_seq_str(seqid=name)
-                for name in names
-                if name in self.names
-            }
+            names = [name for name in names if name in self.names]
 
-        if not selected_seqs:
+        if not names:
             return {}
```
GavinHuttley, line 414: not a good return value. Better to raise an exception, like `ValueError(f"{names=} and {negate=} resulted in no names")`

src/cogent3/core/new_alignment.py
```diff
@@ -356,35 +395,72 @@ def take_seqs(
...
 
-        moltype = kwargs.pop("moltype", self.moltype)
+        seqs_data = self.seqs.subset(names)
 
-        seqs_data = SeqsData(data=selected_seqs, alphabet=moltype.alphabet)
+        moltype = kwargs.pop("moltype", self.moltype)
```
GavinHuttley, line 418: Delete this. Don't support indirect ways of changing moltype. A `to_moltype()` method should be the only way to transform a moltype of a collection. In doing so, we need to add an abstract method of `to_alphabet()` on `SeqDataABC` and implement that on `SeqData`. This can be easy if the alphabet's are nearly identical, like DNA and RNA, but slightly more involved if they are not. An example of the latter case will be someone has created a SeqData instance without defining the correct moltype. If we default to a general one (like "text") then it's a superset of the others. So we need to convert the stored data into the new index coordinates. There are classes in `new_alphabet.py` to assist with this.

src/cogent3/core/new_alignment.py
```diff
@@ -356,35 +395,72 @@ def take_seqs(
...
         result = self.__class__(
             seqs_data=seqs_data, moltype=moltype, info=self.info, **kwargs
         )
-        # todo: kath, update annotation_db and assign to result when we have annotations working
+        if self.annotation_db:
```
GavinHuttley, line 422: copying an annotation_db could be quite an expensive operation. Let's add an optional argument `copy_annotations=False`. In that case we just assign the same annotation db.

src/cogent3/core/new_alignment.py
```diff
@@ -356,35 +395,72 @@ def take_seqs(
...
+            result.annotation_db = type(self.annotation_db)()
+            result.annotation_db.update(
+                annot_db=self.annotation_db, seqids=result.names
+            )
         return result
 
+    def get_seq_indices(
```
GavinHuttley, line 429: change method name to `get_seq_names_if()`

src/cogent3/core/new_alignment.py
```diff
@@ -356,35 +395,72 @@ def take_seqs(
...
+        self, f: Callable[[new_seq.Sequence], bool], negate: bool = False
+    ):
+        """Returns list of names of seqs where f(seq) is True."""
+        get = self.seqs.__getitem__
+
+        if negate:
+
+            def new_f(x):
```
GavinHuttley, line 437: I think this use case is likely common enough that we should add a class to `cogent3.util.misc` like the following and apply it in all those cases. The main reason to avoid such local functions is they cause problems for pickling (required for parallel computations)
```python
import typing

class negate_condition:
    """negates the result of calling func."""
    def __init__(self, func: typing.Callable[[typing.Any], bool]) -> None:
        self.func = func
    
    def __call__(self, *args: typing.Any, **kwargs: typing.Any) -> bool:
        return not bool(self.func(*args, **kwargs))
```

tests/test_core/test_new_alignment.py
```diff
@@ -77,32 +72,51 @@ def ragged():
...
+    return [
+        ("A---CTG-C", [[1, 3], [4, 1]]),
+        ("-GTAC--", [[0, 1], [4, 2]]),
+        ("---AGC--TGC--", [[0, 3], [3, 2], [6, 2]]),
+    ]
+
+
+def test_seqs_data_default_attributes(dna_sd: new_aln.SeqsData):
+    assert dna_sd.names == ["seq1", "seq2", "seq3"]
     assert isinstance(dna_sd._alphabet, new_alpha.CharAlphabet)
```
GavinHuttley, line 96: change `dna_sd._alphabet` to `dna_sd.alphabet`

tests/test_core/test_new_alignment.py
```diff
@@ -552,8 +598,81 @@ def test_set_wrap_affects_repr_html():
     assert got.count(token) == 3 * orig.count(token)
 
 
+@pytest.mark.xfail(reason="todo: kath, new SequenceCollection does not support slicing")
```
GavinHuttley, line 601: neither did the old one, so this test only appropriate for Alignments

GavinHuttley (changes requested):
Make the arguments to `SeqData` keyword only, hence `def __init__(self, *, ...)`

We need a `SeqsDataABC` which defines all the methods on the current implementation as `abstractmethod`. This ABC will be modified as our design evolves, but it's the key to us opening cogent3 up to having such core functionality replaced. I will also be the base for the AlignedSeqsData` to be implemented.

One method that should be present is `seq_lengths` as a property which returns a dict of string name as the key and the length of the sequences. For our current implementation, this should be computed on the fly.

also, for the `_repr_html_`, can you just confirm visually within a jupyter notebook that a collection is displaying correctly

src/cogent3/core/new_alignment.py
```diff
@@ -198,10 +196,16 @@ def get_seq_bytes(
 
     def get_seq_view(self, seqid: str) -> SeqDataView:
         seq_len = len(self._data[seqid])
-        return SeqDataView(seq=self, seqid=seqid, seq_len=seq_len)
+        return SeqDataView(seqs=self, seqid=seqid, seq_len=seq_len)
+
+    def subset(self, names: Union[str, typing.Sequence[str]]) -> SeqsData:
+        """Returns a new SeqsData object with only the specified names."""
+        names = [names] if isinstance(names, str) else names
```
fredjaya, line 203: what happens if `names=...` provided do not match `self.names`?

src/cogent3/core/new_alignment.py
```diff
@@ -198,10 +196,16 @@ def get_seq_bytes(
 
     def get_seq_view(self, seqid: str) -> SeqDataView:
         seq_len = len(self._data[seqid])
-        return SeqDataView(seq=self, seqid=seqid, seq_len=seq_len)
+        return SeqDataView(seqs=self, seqid=seqid, seq_len=seq_len)
+
+    def subset(self, names: Union[str, typing.Sequence[str]]) -> SeqsData:
```
fredjaya, line 201: Should creating a new `SeqsData` happen without getting a `SeqDataView` first? i.e. when should `subset` be used instead of `__getitem__`?

src/cogent3/core/new_alignment.py
```diff
@@ -225,13 +229,15 @@ def _(self, index: int) -> Union[SeqDataView, new_seq.Sequence]:
...
 
 
 @prep_for_seqs_data.register
-def _(data: SeqsData, moltype: new_moltype.MolType) -> dict[str, numpy.ndarray]:
+def _(
+    data: SeqsData, moltype: new_moltype.MolType, label_to_name=None
+) -> dict[str, numpy.ndarray]:
     if moltype.is_compatible_alphabet(data.alphabet):
         return data
     else:
```
fredjaya, line 243: not needed if `data` is already returned?

fredjaya (commented):
Some comments for `new_alignment`. Should `SeqDataView` be renamed to `SeqsDataView` to match `SeqsData`?

GavinHuttley:
> Some comments for `new_alignment`. Should `SeqDataView` be renamed to `SeqsDataView` to match `SeqsData`?

I think keep it as the singular form `SeqDataView` because it refers to a view on a single sequence. By using the plural in `SeqsData`, we are communicating that this is a collection of sequences.

src/cogent3/core/new_alignment.py
KatherineCaley: If you want to subset your data (to a subset of more than one), I don't think it makes sense to go via individual views

src/cogent3/core/new_alignment.py
KatherineCaley: `<dict>.get` should handle if the keys are not present. But if the resulting data is an empty dict then we should probs raise an error

src/cogent3/core/new_alignment.py
```diff
@@ -138,11 +142,73 @@ def copy(self, sliced: bool = False):
...
 
 
-class SeqsData:
+class SeqsDataABC(ABC):
+    """Abstract base class for respresenting the collection of sequences underlying
+    a SequenceCollection
+    """
+
+    alphabet: new_alpha.CharAlphabet
+    make_seq: Callable = None
```
GavinHuttley, line 151: add a comment about refactoring the type hint for callable. (At some point we should be specific in the type hints for the input argument types about and the returned value type.)

src/cogent3/core/new_alignment.py
```diff
@@ -224,35 +305,28 @@ def _(self, index: int) -> Union[SeqDataView, new_seq.Sequence]:
...
-) -> dict[str, numpy.ndarray]:
-    raise NotImplementedError(f"SeqsData can not be constructed for type {type(data)}")
-
-
-@prep_for_seqs_data.register
-def _(data: SeqsData, moltype: new_moltype.MolType) -> dict[str, numpy.ndarray]:
-    if moltype.is_compatible_alphabet(data.alphabet):
+@singledispatch
+def make_seqsdata_dict(
+    data, label_to_name: Callable = None
```
GavinHuttley, line 310: as above, add a comment about refactoring the type hint for callable

src/cogent3/core/new_alignment.py
```diff
@@ -286,9 +361,19 @@ def make_unaligned_seqs(
 
     if len(data) == 0:
         raise ValueError("data must be at least one sequence.")
-    seqs_data = prep_for_seqs_data(data, moltype)
 
-    return SequenceCollection(seqs_data=seqs_data, moltype=moltype, info=info)
+    seqs_data = make_seqsdata_dict(data, label_to_name=label_to_name)
+
+    if isinstance(seqs_data, dict):
```
GavinHuttley, line 367: add a comment about design to revisit `make_unaligned_seqs()`. (Specifically I would do a single dispatch on `make_unaligned_seqs()` such that the version which takes a `SeqsData` instance does the checking on the alphabet consistency with the molecular type and then creates the sequence collection. All other types are all bundled together in one version that uses `make_seqsdata_dict()` and creates the `SeqsData` instance, and then calls `make_unaligned_seqs()` again.)

GavinHuttley (changes requested):
Add a basic test for negate_condition to the test_misc.py file.

src/cogent3/core/new_alignment.py
KatherineCaley: Definitely agree to revisit, is evolving a bit randomly as functionality is added to SequenceCollection

GavinHuttley (approved):
Awesome! Keep cranking.

PR #1890, opened 2024-06-06
SequenceCollection refactor - general and annotation db methods

KatherineCaley:
A few tests marked as xfail - we should probably decide whether these are behaviours we want to support on the new sequence collection

### DIFF
--- a/src/cogent3/core/alignment.py
+++ b/src/cogent3/core/alignment.py
@@ -484,17 +484,17 @@ class _SequenceCollectionBase:
         self._annotation_db = None
 
     @property
-    def named_seqs(self):
+    def named_seqs(self):  # ported
         if self._named_seqs is None:
             self._named_seqs = _make_named_seqs(self.names, self._seqs)
         return self._named_seqs
 
     @property
-    def annotation_db(self):
+    def annotation_db(self):  # ported
         return self._annotation_db
 
     @annotation_db.setter
-    def annotation_db(self, value):
+    def annotation_db(self, value):  # ported
         if value == self._annotation_db:
             return
 
@@ -503,13 +503,13 @@ class _SequenceCollectionBase:
         for seq in self.seqs:
             seq.replace_annotation_db(value, check=False)
 
-    def __str__(self):
+    def __str__(self):  # ported
         """Returns self in FASTA-format, respecting name order."""
         from cogent3.format.alignment import FORMATTERS
 
         return FORMATTERS["fasta"](self.to_dict())
 
-    def _set_additional_attributes(self, curr_seqs):
+    def _set_additional_attributes(self, curr_seqs):  # will not port
         """Sets additional attributes based on current seqs: class-specific."""
         self.seq_data = curr_seqs
         self._seqs = curr_seqs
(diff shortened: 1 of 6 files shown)


src/cogent3/core/alignment.py
```diff
@@ -913,11 +915,11 @@ def to_nexus(self, seq_type, wrap=50):
         return nexus_from_alignment(self, seq_type, wrap=wrap)
 
     @property
-    def num_seqs(self):
+    def num_seqs(self):  # ported
         """Returns the number of sequences in the alignment."""
         return len(self.named_seqs)
 
-    def __add__(self, other):
+    def __add__(self, other):  # will not port
```
KatherineCaley, line 922: should we port? would have to think about what this means for SeqsData

src/cogent3/core/alignment.py
```diff
@@ -940,7 +942,7 @@ def __add__(self, other):
         new.annotation_db = None
         return new
 
-    def add_seqs(self, other, before_name=None, after_name=None):
+    def add_seqs(self, other, before_name=None, after_name=None):  # will not port
```
KatherineCaley, line 945: should we port? would have to think about what this means for SeqsData

src/cogent3/core/alignment.py
```diff
@@ -1049,13 +1051,13 @@ def write(self, filename=None, format=None, **kwargs):
...
 
-    def __len__(self):
+    def __len__(self):  # ported
         """len of SequenceCollection returns length of longest sequence."""
         return self.seq_len
 
     def get_translation(
         self, gc=None, incomplete_ok=False, include_stop=False, trim_stop=True, **kwargs
-    ):
+    ):  # will not port
```
KatherineCaley, line 1060: actually will port

src/cogent3/core/alignment.py
GavinHuttley: on SeqsData this operation should be supported, but it returns a new SeqsData instance. So that means adding `add_seqs()` to the ABC

src/cogent3/core/new_alignment.py
```diff
@@ -395,6 +332,7 @@ def __init__(
         self.info = info
         self._repr_policy = dict(num_seqs=10, num_pos=60, ref_name="longest", wrap=60)
         self._annotation_db = annotation_db or DEFAULT_ANNOTATION_DB()
+        self.seq_len = max(self.seqs.seq_lengths().values())
```
GavinHuttley, line 335: delete this attribute

src/cogent3/core/new_alignment.py
```diff
@@ -535,7 +472,412 @@ def take_seqs_if(
...
+        if self.annotation_db.compatible(seq_db, symmetric=False):
+            # our db contains the tables in other, so we update in place
+            self.annotation_db.update(annot_db=seq_db, seqids=self.names)
+        else:
+            # we use the union method to define a new one
+            # the setter handles propagation of the new instance to bound
+            # sequences
+            self.annotation_db = self.annotation_db.union(seq_db)
+
+    def annotate_from_gff(
```
GavinHuttley, line 554: delete this method, users must use constructor argument or assignment

GavinHuttley (changes requested):
you rock! just a few tweaks

GavinHuttley (approved):
bonza

PR #1894, opened 2024-06-08
Seq collections refactor

KatherineCaley:
- add `SeqsData.to_alphabet`
- add `SequenceCollection.to_moltype`
- add `SequenceCollection.dotplot`

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -455,6 +455,9 @@ class SequenceCollection:
         # todo: kath, note gavins comment that this method could operate on only the names
         # list and not the seqs_data object
 
+        if isinstance(names, str):
+            names = [names]
+
         if negate:
             names = [name for name in self.names if name not in names]
         else:
@@ -496,6 +499,13 @@ class SequenceCollection:
     ):
         """Returns new collection containing seqs where f(seq) is True.
 
+        Parameters
+        ----------
+        f
+            function that takes a sequence object and returns True or False
+        negate
+            select all sequences EXCEPT those where f(seq) is True
+
         Notes
         -----
         The seqs in the new collection are the same objects as the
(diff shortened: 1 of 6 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -517,6 +564,39 @@ def degap(self, **kwargs):
...
+        ----------
+        moltype
+            name of the new moltype, e.g, 'dna', 'rna'.
+
+        Notes
+        -----
+        Cannot convert from nucleic acids to proteins. Use get_translation() for that.
+
+        """
+        mtype = new_moltype.get_moltype(moltype)
```
GavinHuttley, line 580: suggest inserting the following statement
```python
if mtype is self.moltype:
    return self  # nothing to be done
```

src/cogent3/core/new_alignment.py
```diff
@@ -718,6 +798,113 @@ def to_phylip(self):
...
+            name = list(numpy.random.choice(names, size=1))[0]
+            name1 = name1 or name
+            name2 = name2 or name
+
+        if not {name1, name2} <= set(self.names):
+            msg = f"{name1}, {name2} missing"
+            raise ValueError(msg)
+
+        seq1 = self.get_seq(seqname=name1, copy_annotations=True)
+        seq2 = self.get_seq(seqname=name2, copy_annotations=True)
```
GavinHuttley, line 867: we only copy annotations if it's an operation where the user is likely to modify them. In this case, since it's just for display don't copy annotations.

src/cogent3/core/new_alignment.py
```diff
@@ -1244,7 +1431,7 @@ def make_unaligned_seqs(
     # create a db from sequences if they're annotated
     seqs_anno_db = merged_db_collection(data)
     # if db from seqs and provided db, merge
-    # todo: kath, should we default to not merging unless specified?
+    # todo: kath, should we default to provided db if both? or merge?
```
GavinHuttley, line 1434: default to provided db, add a comment to describe this in the docstring

tests/test_core/test_alignment.py
```diff
@@ -390,7 +390,7 @@ def test_get_identical_sets(self):  # ported
         got = frozenset(frozenset(s) for s in got)
         self.assertEqual(got, expect)
 
-    def test_is_ragged(self):  # ported
+    def test_is_ragged(self):  # will not port
```
GavinHuttley, line 393: why not porting this? (message me in Zulip)

GavinHuttley (changes requested):
one change for efficiency and some comments to be added

tests/test_core/test_alignment.py
KatherineCaley: messaged

PR #1895, opened 2024-06-10
DEV: Porting sequence tests (trim_stop_codon) and new module methods

fredjaya:
Porting `sequence` tests, mainly related to trim_stop_codons, and changes to `new_moltype` and `new_genetic_code` to get these working (+ tests). Adding review comments for things that need extra attention.


src/cogent3/core/new_genetic_code.py
```diff
@@ -127,9 +127,6 @@ def __post_init__(self, ncbi_code_sequence: str, ncbi_start_codon_map: str):
     def __hash__(self):
         return hash(self.name)
 
-    def __eq__(self, other):
```
fredjaya, line 130: Is the duplicate `__eq__` defined on L162 the correct implementation?

tests/test_core/test_new_moltype.py
```diff
@@ -192,3 +192,20 @@ def test_get_degenerate_positions(data_type, moltype):
...
+    (
+        new_moltype.ASCII,
+        new_moltype.DNA,
+        new_moltype.RNA,
+        new_moltype.PROTEIN,
+        new_moltype.PROTEIN_WITH_STOP,
+    ),
+)
+def test_gaps(moltype):
+    # TODO: fred, bytes
```
fredjaya: I don't quite understand how to test for `new_moltype.BYTES` e.g. `MolType(range(0, 255)))`

tests/test_core/test_new_sequence.py
```diff
@@ -12,9 +12,6 @@ def test_moltype_make_seq(name):
     assert str(seq) == raw
 
 
-@pytest.mark.xfail(
```
fredjaya, line 15: Fixed with recent `new_moltype` changes.

src/cogent3/core/new_sequence.py
```diff
@@ -1700,10 +1702,12 @@ def trim_stop_codon(
         If sequence contains gap characters, the result preserves the sequence
         length by adding gap characters at the end.
         """
+        from cogent3.core import new_genetic_code
```
fredjaya, line 1705: Is this ok? Problems with circular imports

tests/test_core/test_new_sequence.py
```diff
@@ -21,7 +29,158 @@ def test_moltype_make_bytes_seq():
...
+    "gc,seq", ((1, "TCCAGG"), (2, "TCCAAA"), (1, "CCTGA"), (2, "CCAGG"))
+)
+def test_has_terminal_stop_false(gc, seq):
+    gc = new_genetic_code.get_code(gc)
+    seq = new_moltype.DNA.make_seq(seq=seq)
+    assert not seq.has_terminal_stop(gc=gc)
+
+
+@pytest.mark.xfail(
+    reason="UnboundLocalError: cannot access local variable 'seq' where it is not associated with a value"
```
fredjaya, line 50: looking into for next PR

tests/test_core/test_new_sequence.py
```diff
@@ -21,7 +29,158 @@ def test_moltype_make_bytes_seq():
...
+    "gc,seq", ((1, "TCCAGG"), (2, "TCCAAA"), (1, "CCTGA"), (2, "CCAGG"))
+)
+def test_trim_terminal_stop_false(gc, seq):
+    gc = new_genetic_code.get_code(gc)
+    seq = new_moltype.DNA.make_seq(seq=seq)
+    assert str(seq.trim_stop_codon(gc=gc)) == str(seq)
+
+
+@pytest.mark.xfail(
+    reason="cogent3.core.alphabet.AlphabetError: None length not divisible by 3"
```
fredjaya, line 97: looking into for next PR

fredjaya:
looking into codacy issues

tests/test_core/test_new_sequence.py
fredjaya: bad test picked up by codacy - fixing

src/cogent3/core/new_genetic_code.py
GavinHuttley: I think this one is correct. Please delete the other -- good catch!

src/cogent3/core/new_moltype.py
```diff
@@ -424,6 +425,10 @@ def degen_gapped_alphabet(self):
         """monomers + gap + ambiguous characters"""
         return self._degen_gapped
 
+    @property
+    def gaps(self):
+        return frozenset([self.gap, self.missing])
```
GavinHuttley, line 430: only add these attributes to the returned value if they are not None. (Some moltypes do not specify them.)

src/cogent3/core/new_sequence.py
GavinHuttley: This is fine, leaving it as import module name (not from module name import)

tests/test_core/test_new_genetic_code.py
```diff
@@ -75,3 +81,16 @@ def test_sizeframes():
     }
     got = set(code.sixframes(seq))
     assert got == expect
+
+
+@pytest.mark.parametrize("code_id", range(1, 5))
+def test_compare_codes_true(code_id):
+    gc = new_genetic_code.get_code(code_id)
+    assert gc == gc
```
GavinHuttley, line 89: something will always equal itself, so that must be an error. What are you trying to test? Perhaps you mean `gc.ID == code_id`?

tests/test_core/test_new_moltype.py
```diff
@@ -192,3 +192,20 @@ def test_get_degenerate_positions(data_type, moltype):
...
+    "moltype",
+    (
+        new_moltype.ASCII,
+        new_moltype.DNA,
+        new_moltype.RNA,
+        new_moltype.PROTEIN,
+        new_moltype.PROTEIN_WITH_STOP,
+    ),
+)
+def test_gaps(moltype):
```
GavinHuttley, line 207: BYTES will fail this since it defines gap and missing as None, so implement the test that BYTES.gaps returns None or perhaps an empty set?

GavinHuttley (changes requested):
nice work! minor tweaks

tests/test_core/test_new_moltype.py
GavinHuttley: BYTES does defines both gap and missing as None. But looking at the original implementation, it just uses the defaults. So I think we should modify the current BYTES definition to reflect that of the original.

src/cogent3/core/new_genetic_code.py
fredjaya: Deleting the other throws:
```
.../test_new_sequence.py: 
    from cogent3.core import (
.../new_genetic_code.py:407:
    _CODES[code] = code
.../new_genetic_code.py:132: in __eq__
    return self.name == other.name
E   AttributeError: 'str' object has no attribute 'name'
```

tests/test_core/test_new_moltype.py
fredjaya: Thanks - will add to next PR

tests/test_core/test_new_moltype.py
fredjaya: Thanks, will work on next PR

fredjaya:
Can reproduce the following error by running
```bash
$ cd Cogent3/
$ pytest # py v3.12
# throws error below
```

but no error when running the specific test module:
```bash
$ pytest tests/test_core/test_new_sequence
# all passing
```

```
=================================== FAILURES ===================================
____________________ test_has_terminal_stop_true[1-TCCTGA] _____________________

gc = Standard
======================================================
aa               IUPAC code    codons                 ...         
STOP             *             TGA,TAA,TAG            
------------------------------------------------------
seq = DnaSequence(TCCTGA)

    @pytest.mark.parametrize("gc,seq", ((1, "TCCTGA"), (1, "ACGTAA---"), (2, "TCCAGG")))
    def test_has_terminal_stop_true(gc, seq):
        gc = new_genetic_code.get_code(gc)
        seq = new_moltype.DNA.make_seq(seq=seq)
>       assert seq.has_terminal_stop(gc=gc)

test_core/test_new_sequence.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../src/cogent3/core/new_sequence.py:1673: in has_terminal_stop
    _, s = self.parse_out_gaps()
../src/cogent3/core/new_sequence.py:[138](https://github.com/cogent3/cogent3/actions/runs/9475104778/job/26105868403?pr=1895#step:5:139)0: in parse_out_gaps
    gap = re.compile(f"[{re.escape(self.moltype.gap)}]+")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

pattern = None

    def escape(pattern):
        """
        Escape special characters in a string.
        """
        if isinstance(pattern, str):
            return pattern.translate(_special_chars_map)
        else:
>           pattern = str(pattern, 'latin1')
E           TypeError: decoding to str: need a bytes-like object, NoneType found

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/re.py:277: TypeError

---------- coverage: platform darwin, python 3.9.13-final-0 ----------
```

src/cogent3/core/new_genetic_code.py
GavinHuttley: then change code to
```python
return self.name == getattr(other, "name", None)
```

GavinHuttley:
The statement
```python
gap = re.compile(f"[{re.escape(self.moltype.gap)}]+")
```
is clearly the culprit and that's because `self.moltype.gap` has somehow become None. As the moltypes are only created once, it is possible some other piece of code is assigning None to the `gap` attribute.
The one way to be sure this can't happen is to modify the MolType dataclass such that `gap` becomes `_gap` and `_missing` becomes `_missing`. Then `gap` and `missing` become `dataclasses.InitVar` only and the assignments are done in the `__post_init__` method. You'd then need to implement two properties which only have getter methods. So here's a snippet along these lines
```python
@dataclasses.dataclass
class MolType:
    name: str
    monomers: dataclasses.InitVar[StrORBytes]
    make_seq: dataclasses.InitVar[typing.Type]
    gap: dataclasses.InitVar[str] = dataclasses.field(default=IUPAC_gap)
    missing: dataclasses.InitVar[str] = dataclasses.field(default=IUPAC_missing)
    # put the following under the "private attributes..." block
    _gap: OptStr =  dataclasses.field(init=False)
    _missing: OptStr =  dataclasses.field(init=False) 

    def __post_init__(self, ..., gap, missing):
        self._gap = gap
        self._missing = missing

    @property
    def gap(self):
        return self._gap

    @property
    def missing(self):
        return self._missing
```
Any code that then seeks to assign a value to `gap` will raise an exception and can be fixed (but we leave the above mod in place)

fredjaya:
The regex error was indeed due to `self.moltype.gap` being assigned `None` in the test. I have updated `test_new_moltype.test_gaps_none()` to avoid that. Let me know if I should continue with the `MolType` mods for this PR or next

tests/test_core/test_new_sequence.py
```diff
@@ -114,8 +114,13 @@ def test_index_a_seq(cast):
 def test_index_a_seq_float_fail(cast):
     seq = new_moltype.DNA.make_seq(seq="TCCAG")
     index = cast(1)
+
+    # avoid codacy "statement seems to have no effect"
+    def idx_seq():
```
GavinHuttley, line 119: When codacy raises an issue that we know is invalid, we can identify the error code and add a comment on the line to force codacy to ignore that line.
Instead, change to the following
```python
with pytest.raises(TypeError):
    seq[index]  # pylint: disable=W0104
```

GavinHuttley (changes requested):
address codacy issue more directly

GavinHuttley:
do the MolType mods in the next PR

GavinHuttley (approved):
Nice one -- keep cranking

PR #1896, opened 2024-06-12
Seq collections refactor

KatherineCaley:


src/cogent3/core/new_moltype.py
```diff
@@ -529,30 +529,143 @@ def _(self, seq: numpy.ndarray) -> bool:
...
     @functools.singledispatchmethod
-    def get_degenerate_positions(self, seq, include_gap=True) -> numpy.ndarray:
-        """returns a boolean array indicating degenerate positions"""
+    def get_degenerate_positions(self, seq, include_gap=True) -> list[int]:
+        """returns indices matching degenerate characters"""
         raise TypeError(f"{type(seq)} not supported")
 
     @get_degenerate_positions.register
-    def _(self, seq: numpy.ndarray, include_gap=True) -> numpy.ndarray:
+    def _(self, seq: numpy.ndarray, include_gap=True) -> list[int]:
```
GavinHuttley, line 537: type hint the bool arguments

src/cogent3/core/new_moltype.py
```diff
@@ -529,30 +529,143 @@ def _(self, seq: numpy.ndarray) -> bool:
...
+    def degap(self, seq) -> StrORBytesORArray:
+        """removes all gap and missing characters from a sequence"""
+        # refactor: design
+        # previous implementation also removed missing characters, as does this
+        # implementation. But should we revisit this?
+        raise TypeError(f"{type(seq)} not supported")
+
+    @degap.register
+    def _(self, seq: str) -> str:
+        trans = dict([(i, None) for i in map(ord, self.gaps)])
```
GavinHuttley, line 567: add a comment to refactor, we should be caching translation callables rather than doing this on every method call. We should make bytes the primary method and an instance of the convert_alphabet which deletes the gap character.

src/cogent3/core/new_moltype.py
```diff
@@ -529,30 +529,143 @@ def _(self, seq: numpy.ndarray) -> bool:
...
+            the motif being queried.
+
+        """
+        ambigs_missing = {
+            self.missing,
+            *frozenset(self.ambiguities.keys()),
+        }
+        return query_motif in ambigs_missing
+
+    def resolve_ambiguity(
```
GavinHuttley, line 600: rather complex method, your comments below are good so we will follow up later

tests/test_core/test_new_moltype.py
```diff
@@ -167,28 +167,115 @@ def test_gap_index_constant(moltype):
...
+    assert got == expect
+
+
+def test_resolve_ambiguity_nucs():
+    got = new_moltype.DNA.resolve_ambiguity("AT?", allow_gap=False)
+    assert len(got) == 4
+    assert len(got[0]) == 3
+
+
+def test_resolve_ambiguity_codons():
```
GavinHuttley, line 204: need to implement a codon alphabet class, please remind me when I get back. Add a comment here to use new_genetic_code when that job is done.

tests/test_core/test_new_moltype.py
```diff
@@ -167,28 +167,115 @@ def test_gap_index_constant(moltype):
...
+    assert len(new_moltype.DNA.resolve_ambiguity("AT?", alphabet=codon_alpha)) == 4
+    assert len(new_moltype.DNA.resolve_ambiguity("???", alphabet=codon_alpha)) == 61
+
+    with pytest.raises(new_alphabet.AlphabetError):
+        new_moltype.DNA.resolve_ambiguity("at-")
+    with pytest.raises(new_alphabet.AlphabetError):
+        new_moltype.DNA.resolve_ambiguity("---", alphabet=codon_alpha)
+
+
+def test_is_ambiguity():
```
GavinHuttley, line 229: pytest.mark.parametrize and do two tests for the True and False cases

GavinHuttley (changes requested):
few additional comments only

src/cogent3/core/new_moltype.py
KatherineCaley: @GavinHuttley, I will add a comment about caching. The reason I didn't send all dispatches through bytes is that I thought it would be best to return the same type as the original sequence.

KatherineCaley: have added a comment about making bytes the primary method too

src/cogent3/core/new_alignment.py
```diff
@@ -350,10 +363,11 @@ class SequenceCollection:
     def __init__(
         self,
         *,
-        seqs_data: SeqsData,
+        seqs_data: SeqsDataABC,
         moltype: new_moltype.MolType,
         names: list[str] = None,
         info: Union[dict, InfoClass] = None,
+        source: str = None,
```
GavinHuttley, line 370: add comment about dealing with optional args type hints

src/cogent3/core/new_alignment.py
```diff
@@ -556,17 +586,21 @@ def to_dict(self, as_array: bool = False) -> dict[str, Union[str, numpy.ndarray]
...
+        The returned collection will not retain an annotation_db if present.
         """
         # refactor: array - chars > gap char
-        seqs = {name: self.seqs[name].degap() for name in self.names}
-        return make_unaligned_seqs(seqs, moltype=self.moltype, info=self.info, **kwargs)
+        seqs = {
+            name: self.moltype.degap(self.seqs.get_seq_array(seqid=name))
+            for name in self.names
+        }
+        return make_unaligned_seqs(
```
GavinHuttley, line 601: add a comment to refactor. Within a class, methods that return self should use `self.__class__(...)`, using that function raises the prospect of circular imports.

src/cogent3/core/new_alignment.py
```diff
@@ -912,6 +1007,378 @@ def dotplot(
...
+        assert not self.is_ragged(), "all sequences must have same length"
+        from cogent3.parse import cisbp, jaspar
+
+        assert pssm or path, "Must specify a PSSM or a path"
+        assert not (pssm and path), "Can only specify one of pssm, path"
+
+        if isinstance(names, str):
+            names = [names]
+
+        if path:
```
GavinHuttley, line 1053: add a comment to redesign. This method is handling possible cases which would be better delegated to a new function called `load_pssm()` whch should be somewhere else than here.

src/cogent3/core/new_alignment.py
```diff
@@ -912,6 +1007,378 @@ def dotplot(
...
+
+        return numpy.array(result)
+
+    def get_ambiguous_positions(self):
+        """Returns dict of seq:{position:char} for ambiguous chars.
+
+        Used in likelihood calculations.
+        """
+        result = {}
+        for name in self.names:
```
GavinHuttley, line 1089: add a comment to refactor for performance

src/cogent3/core/new_alignment.py
```diff
@@ -912,6 +1007,378 @@ def dotplot(
...
+            valid input to cogent3.get_code(), a genetic code object, number
+            or name, defaults to standard code
+        strict
+            If True, raises an exception if a seq length not divisible by 3
+        """
+        if not self.has_terminal_stop(gc=gc, strict=strict):
+            return self
+
+        new_seqs = {s.name: s.trim_stop_codon(gc=gc, strict=strict) for s in self.seqs}
+        result = make_unaligned_seqs(new_seqs, moltype=self.moltype, info=self.info)
```
GavinHuttley, line 1111: another use of make_unaligned_seqs(). Make a comment about refactor to directly call `SequenceCollection` class.

src/cogent3/core/new_moltype.py
```diff
@@ -532,32 +532,141 @@ def _(self, seq: bytes) -> bool:
 
     @is_gapped.register
     def _(self, seq: numpy.ndarray) -> bool:
-        return (seq == self.degen_gapped_alphabet.gap_index).any()
+        gaps_indices = [self.degen_gapped_alphabet.index(gap) for gap in self.gaps]
+        return numpy.isin(seq, gaps_indices).any()
```
GavinHuttley, line 536: nice!

src/cogent3/core/new_moltype.py
```diff
@@ -532,32 +532,141 @@ def _(self, seq: bytes) -> bool:
...
     @functools.singledispatchmethod
-    def get_degenerate_positions(self, seq, include_gap=True) -> numpy.ndarray:
-        """returns a boolean array indicating degenerate positions"""
+    def get_degenerate_positions(self, seq, include_gap: bool = True) -> list[int]:
+        """returns indices matching degenerate characters"""
         raise TypeError(f"{type(seq)} not supported")
 
     @get_degenerate_positions.register
-    def _(self, seq: numpy.ndarray, include_gap=True) -> numpy.ndarray:
+    def _(self, seq: numpy.ndarray, include_gap: bool = True) -> list[int]:
```
GavinHuttley, line 544: you can remove the singledispatch decoration. Juts convert to using this method with a small tweak
```python
seq = self.degen_gapped_alphabet.to_indices(seq)
...
```
The alphabet.to_indices(seq) method handles all the different types!

src/cogent3/core/new_moltype.py
```diff
@@ -532,32 +532,141 @@ def _(self, seq: bytes) -> bool:
...
+            self.missing: set([*self.alphabet, self.gap]),
+        }
+        for m in self.alphabet:
+            ambiguities[m] = frozenset(m)
+        if alphabet is None:
+            word_alpha = self.alphabet.get_kmer_alphabet(
+                k=len(ambig_motif), include_gap=allow_gap
+            )
+            # refactor: design
+            # with_gap_motif() is not implemented on KmerAlphabet, check whether
```
GavinHuttley, line 639: KmerAlphabet is definitely incomplete...

GavinHuttley (changes requested):
mainly suggesting adding some comments, but one code simplication -- yay!

src/cogent3/core/new_alignment.py
KatherineCaley: ah yes I was wondering about this

GavinHuttley (approved):
awesome!

PR #1907, opened 2024-06-21
TST: xfailing sequence tests

fredjaya:
Three different xfails here, let me know if they should be left xfailed or addressed!


GavinHuttley (approved):
awesome!

PR #1909, opened 2024-06-21
Amendments to improve genetic code translate

GavinHuttley:


KatherineCaley (approved):

src/cogent3/core/new_genetic_code.py
```diff
@@ -209,12 +233,12 @@ def translate(
             dna = dna[:-diff]
 
         # convert to indices and then bytes
-        seq = self.codons.to_indices(dna).tobytes()
+        seq = self.codons.to_indices(dna)
```
KatherineCaley, line 236: If the input is already `bytes` is there a reason to convert to indices and then bytes?

GavinHuttley: Yes. Indices are integers, bytes is the "raw" underlying data which is required by the `bytes.translate` method.

GavinHuttley:
The codacy issue here will be fixed when the `KmerAlphabet` is fully implemented

src/cogent3/core/new_genetic_code.py
GavinHuttley: Yes, easiest with an example
```python
# example user input
seq = b"TCAG" 
# we require that seq correspond to integers 0, 1, 2, 3 but
seq[0] == 84
```
so we use `to_indices()` to produce that conversion.

KatherineCaley: `dna` could be str, bytes, or an array. If it's bytes, does it need to be converted to indices and then back to bytes?

PR #1927, opened 2024-06-27
TST: port seqview tests

fredjaya:
no new xfails, main changes so tests work with the new implementation are:
- adding alphabets arg to SeqView
- replacing value with str_value

nearly there...


tests/test_core/test_sequence.py
```diff
@@ -1275,504 +1275,12 @@ def test_get_translation_trim_stop():  # ported
     assert str(aa) == "I*L"
 
 
-@pytest.mark.parametrize("start", (None, 0, 1, 10, -1, -10))
```
KatherineCaley, line 1278: I think we should be leaving the old tests in place

GavinHuttley (changes requested):
leave old tests in place, just comment as # ported

fredjaya:
Will do. Should all the ones deleted up until now be re-added?

fredjaya:
Ported a few more tests and marked those in place. Let me know if I should re-add all the previous ones deleted before this PR

GavinHuttley (changes requested):
All tests of code that remains in the original sequence, moltype, alignment and genetic_code modules, should also remain.

GavinHuttley (approved):
awesome!

PR #1929, opened 2024-06-27
ENH: deserialisation of `SequenceCollection` and `SeqsData`, fixes #1924

### LINKED ISSUE #1924: ENH: deserialise function for SequenceCollection
- [x] `SequenceCollection`
- [x] `SeqsData`
- [x] Sequence
- [x] DNA/RNA/Protein sequence

KatherineCaley:
Includes 
- serialisation and deserialisation of `Sequence` and `SequenceCollection` classes 
- primitive serialisation of `Seq(Data)View` classes and no deserialisation

addresses #1924

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -193,14 +193,16 @@ class SeqDataView(new_seq.SeqViewABC, new_seq.SliceRecordABC):
 
         data = {"type": get_object_provenance(self), "version": __version__}
         data["init_args"] = self._get_init_kwargs()
-        data["init_args"].update({
-            "step": self.step,
-            "start": self.start,
-            "stop": self.stop,
-            "seq": self.seq.to_rich_dict(),
-            "offset": int(self.parent_start),
-            "seq_len": self.seq_len
-        })
+        data["init_args"].update(
+            {
+                "step": self.step,
+                "start": self.start,
+                "stop": self.stop,
+                "seq": self.seq.to_rich_dict(),
+                "offset": int(self.parent_start),
+                "seq_len": self.seq_len,
+            }
+        )
 
         return data
 
@@ -781,7 +783,7 @@ class SequenceCollection:
         moltype = data["init_args"].pop("moltype")
         moltype = new_moltype.get_moltype(moltype)
 
-        return cls(seqs_data=seqs_data, **data["init_args"], moltype = moltype)
+        return cls(seqs_data=seqs_data, **data["init_args"], moltype=moltype)
 
     def to_json(self):
         """returns json formatted string"""
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -189,20 +189,31 @@ class SeqDataView(new_seq.SeqViewABC, new_seq.SliceRecordABC):
         return self
 
     def to_rich_dict(self) -> dict[str, str | dict[str, str]]:
-        """returns a json serialisable dict"""
+        """returns a json serialisable dict.
+
+        Notes
+        -----
+        This method will slice the underlying sequence to the start and stop values
+
+        Warning
+        -------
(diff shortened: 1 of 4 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -187,9 +188,34 @@ def copy(self, sliced: bool = False):
...
+    def to_rich_dict(self) -> dict[str, str | dict[str, str]]:
+        """returns a json serialisable dict.
+
+        Notes
+        -----
+        This method will slice the underlying sequence to the start and stop values
+
+        Warning
+        -------
+        This method is not intended to provide deserialisation of this object,
```
GavinHuttley, line 200: change "deserialisation" to "serialisation" and "it is intended to usage by an eclosing class".

Also, if this is the `SeqDataView` class then it cannot work for deserialisation since that class requires it's `seq` object to be a `SeqsDataABC`.

src/cogent3/core/new_sequence.py
```diff
@@ -171,7 +197,7 @@ def to_rich_dict(self, exclude_annotations=False):
         )
         if hasattr(self, "annotation_offset"):
             offset = int(self._seq.parent_start)
```
GavinHuttley, line 199: can you just confirm that `parent_start` is on the SliceRecordABC

src/cogent3/core/new_sequence.py
```diff
@@ -182,6 +208,12 @@ def to_rich_dict(self, exclude_annotations=False):
 
         return data
 
+    @classmethod
+    def from_rich_dict(cls, data):
+        moltype, seq = _moltype_seq_from_rich_dict(data)
+
+        return cls(moltype, seq, **data)
```
GavinHuttley, line 215: use keyword arguments for future proffing

src/cogent3/core/new_sequence.py
```diff
@@ -1579,25 +1611,57 @@ def parent_coordinates(self) -> tuple[str, int, int, int]:
...
+
+
 class ProteinSequence(Sequence):
     """Holds the standard Protein sequence."""
 
     # constructed by PROTEIN moltype
-    ...
+
+    @classmethod
+    def from_rich_dict(cls, data):
```
GavinHuttley, line 1625: I wonder if you can get away with not defining the `from_rich_dict()` method on all the subclasses, since they inherit that. Registering the deserialiser for each subclass may be sufficient!

tests/test_core/test_new_alignment.py
```diff
@@ -362,6 +363,19 @@ def test_seqs_data_to_alphabet_invalid():
...
+def test_seqs_data_round_trip(reverse, alpha):
+    seqs_data = new_aln.SeqsData(
+        data={"seq1": "ACGG", "seq2": "CGCA", "seq3": "CCG-"}, alphabet=alpha
+    )
+    seqs_data = seqs_data.reverse_seqs() if reverse else seqs_data
+
+    rd = seqs_data.to_rich_dict()
+    got = deserialise_object(rd)
+    assert isinstance(got, new_aln.SeqsData)
+    assert got.to_rich_dict() == seqs_data.to_rich_dict()
```
GavinHuttley, line 376: this is just exercising the reverse method. I think you should check the value for at least one sequence is correct against expectation when you use `str()` on `get_seq_view()`.

GavinHuttley (changes requested):
Few tweaks, but looking good!

src/cogent3/core/new_sequence.py
KatherineCaley: yep

src/cogent3/core/new_sequence.py
KatherineCaley: can indeed get away with it!

tests/test_core/test_new_alignment.py
KatherineCaley: good point!

Issue #1897, opened 2024-06-14
some app.typing union types not working when importing annotations from future


GavinHuttley:
The SeqsCollectionType, UnalignedSeqsType don't get resolved by typing.get_constraint_names() when `from __future__ import annotations` is invoked.

We should detect this setting and either change algorithm for resolving types or raise an exception.

Here's a demo.
```python
# from __future__ import annotations

from cogent3.app.composable import define_app
from cogent3.app.typing import SeqsCollectionType
from cogent3.core.alignment import ArrayAlignment

@define_app
def make_coll(data: dict[str, str]) -> SeqsCollectionType:
    return ArrayAlignment(data)

@define_app
def array_align_to_fasta(aln: ArrayAlignment) -> str:
    return aln.to_fasta()

# the following works with the from future line commented out, but fails with TypeError if it's active
app = make_coll() + array_align_to_fasta()
```

rmcar17:
So, the problem with this of course is that the `from __future__ import annotations` stringifys the type hints. This can be resolved through using `typing.get_type_hints` rather than `inspect` in `composable.py`'s `_get_raw_hints` function (though might take some time for me to properly verify).

However, should this example actually proceed without error? For instance, it is currently allowable for `make_coll` to return something of type `SequenceCollection` or `Alignment` in addition to `ArrayAlignment` but `array_align_to_fasta` only accepts something of type `ArrayAlignment`.

GavinHuttley:
Yes because
```python
AlignedSeqsType = TypeVar("AlignedSeqsType", "Alignment", "ArrayAlignment") # this
UnalignedSeqsType = TypeVar("UnalignedSeqsType", bound="SequenceCollection")
SeqsCollectionType = Union[AlignedSeqsType, UnalignedSeqsType] # is part of this
```

GavinHuttley:
Perhaps the thing to do for now is to raise detect this case and raise an exception about not currently future proof? (pun intended)

rmcar17:
> Yes because
> 
> ```python
> AlignedSeqsType = TypeVar("AlignedSeqsType", "Alignment", "ArrayAlignment") # this
> UnalignedSeqsType = TypeVar("UnalignedSeqsType", bound="SequenceCollection")
> SeqsCollectionType = Union[AlignedSeqsType, UnalignedSeqsType] # is part of this
> ```

Just to follow up on this, shouldn't this simpler case fail?

```python
from cogent3.app.composable import define_app

@define_app
[...]

PR #1931, opened 2024-06-28
DEV: raise not implemented error on top level stringified type hint

rmcar17:
Not a fix to #1897, though raises an error when it detects a top-level stringified type hint for either the input or output of an app.

By top-level, I mean it doesn't check nested type hints - i.e. patterns like `typing.Union["int","str"]`. To do so would break the current behaviour of `cogent3.app.typing.get_constraint_names`. It would be better to properly resolve the issue instead.

It will detect the problem-causing `from __future__ import annotations` pattern and raise a `NotImplementedError` for now.

### DIFF
--- a/src/cogent3/app/composable.py
+++ b/src/cogent3/app/composable.py
@@ -153,6 +153,21 @@ def _get_raw_hints(main_func, min_params):
     if depth > 2:
         raise TypeError(msg.format("return_type", return_type, depth))
 
+    if isinstance(first_param_type, str):
+        msg = (
+            "Apps do not yet support string type hints "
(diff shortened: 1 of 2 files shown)


GavinHuttley (changes requested):
That looks good, just add a test. Note that you can create a function, like
```python
def foo(a: "type") -> int:
    return int(a)
```
and trigger the error during execution of the app decorator as
```python
def test_bad_wrap():
    with pytest.raises(NotImplementedError):
        define_app(foo)
```

rmcar17:
> That looks good, just add a test. 
Done!

GavinHuttley (approved):
awesome, thanks!

Issue #1932, opened 2024-07-01
MolType methods to port


GavinHuttley:
- [x] strip_degenerate: use `new_alphabet.convert_alphabet(delete=<all the degenerate characters>)` ?
- [x] disambiguate
- [x] strip_bad: converts to ints and deletes things beyond allowed?
- [x] strip_bad_and_gaps: as above
- [x] count_gaps
- [ ] gap_indices: look at algorithm relationship to `Sequence.parse_out_gaps()`, they should share code
- [ ] gap_vector: where is the method used?
- [ ] gap_maps: do not port, this overlaps with IndelMap
- [x] count_gaps
- [x] count_degenerate
- [x] possibilities: but think of a better name!
- [x] mw
- [ ] can_match: check definition of match, look at test cases, do they seem biologically interesting?
- [ ] can_mismatch: as above
- [x] can_pair
- [x] can_mispair
- [x] must_pair

KatherineCaley:
todo: 
- [ ] strip_degenerate: use `new_alphabet.convert_alphabet(delete=<all the degenerate characters>)` ?
- [ ] disambiguate
- [ ] strip_bad: converts to ints and deletes things beyond allowed?
- [ ] strip_bad_and_gaps: as above
- [ ] count_gaps
- [ ] count_gaps
- [ ] count_degenerate
- [ ] possibilities: but think of a better name!
- [ ] mw
- [ ] can_pair
- [ ] can_mispair
- [ ] must_pair

KatherineCaley:
@GavinHuttley possible other name for `possibilities` is `count_variants`

PR #1935, opened 2024-07-02
ENH: add strip methods to moltype

### LINKED ISSUE #1921: Sequence and SeqView classes should not be used to replace characters
Delete the `_seq_filter()` method from sequence classes in `new_sequence.py`. No longer being used.

Also delete `new_sequence.Sequence.replace()` method.  Users will need to do this explicitly themselves.

There will be side-effects in that some functions (e.g. `moltype._convert_to_rna()` and `_convert_to_dna()`) employ the replace method. We can just produce revised functions that cast incoming data to string and using the replace method on that.

KatherineCaley:
- add `strip_degenerate`, `strip_bad`, `strip_bad_and_gaps`, `disambiguate` to `Moltype` -- addresses #1932
- add `count_variants`, `count_gaps`, `count_degenerate` to `Moltype` -- addresses #1938
- remove replace from sequences -- fixes #1921 
- port Sequence tests

### DIFF
--- a/src/cogent3/core/new_sequence.py
+++ b/src/cogent3/core/new_sequence.py
@@ -35,7 +35,7 @@ from cogent3.core.annotation_db import (
 )
 from cogent3.core.info import Info as InfoClass
 from cogent3.core.location import FeatureMap, IndelMap, LostSpan
-from cogent3.format.fasta import alignment_to_fasta
+from cogent3.format.fasta import seqs_to_fasta
 from cogent3.maths.stats.contingency import CategoryCounts
 from cogent3.maths.stats.number import CategoryCounter
 from cogent3.util.deserialise import register_deserialiser
@@ -152,7 +152,7 @@ class Sequence:
                 result = self.moltype.complement(result)
         return result
 
-    def to_fasta(self, make_seqlabel=None, block_size=60):
+    def to_fasta(self, make_seqlabel=None, block_size=60) -> str:
         """Return string of self in FASTA format, no trailing newline
 
         Parameters
@@ -171,9 +171,11 @@ class Sequence:
             label = self.label
         elif hasattr(self, "name") and self.name:
             label = self.name
-        return alignment_to_fasta({label: str(self)}, block_size=block_size)
+        return seqs_to_fasta({label: str(self)}, block_size=block_size)
 
-    def to_rich_dict(self, exclude_annotations=True):
+    def to_rich_dict(
+        self, exclude_annotations: bool = True
+    ) -> dict[str, str | dict[str, str]]:
         """returns {'name': name, 'seq': sequence, 'moltype': moltype.label}
 
         Notes
@@ -209,12 +211,12 @@ class Sequence:
         return data
 
     @classmethod
-    def from_rich_dict(cls, data):
+    def from_rich_dict(cls, data: dict):
         moltype, seq = _moltype_seq_from_rich_dict(data)
 
         return cls(moltype=moltype, seq=seq, **data)
 
-    def to_json(self):
+    def to_json(self) -> str:
         """returns a json formatted string"""
         return json.dumps(self.to_rich_dict())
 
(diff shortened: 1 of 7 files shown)


src/cogent3/core/new_sequence.py
```diff
@@ -327,84 +329,80 @@ def strip_degenerate(self):
         """Removes degenerate bases by stripping them out of the sequence."""
         return self.__class__(
             moltype=self.moltype,
-            seq=self.moltype.strip_degenerate(self),
+            seq=self.moltype.strip_degenerate(str(self)),
```
GavinHuttley, line 332: comment about refactor, perhaps `numpy.array(self)` more efficient?

src/cogent3/core/new_sequence.py
```diff
@@ -327,84 +329,80 @@ def strip_degenerate(self):
...
             info=self.info,
         )
 
     def strip_bad(self):
         """Removes any symbols not in the alphabet."""
+        # refactor: design
+        # previous implementation did NOT strip X from DNA even though its not in the alphabet
```
GavinHuttley, line 339: no, we don't want to keep that behaviour

src/cogent3/core/new_sequence.py
```diff
@@ -327,84 +329,80 @@ def strip_degenerate(self):
...
-            moltype=self.moltype, seq=self.moltype.strip_bad(self), info=self.info
+            moltype=self.moltype, seq=self.moltype.strip_bad(str(self)), info=self.info
         )
 
     def strip_bad_and_gaps(self):
         """Removes any symbols not in the alphabet, and any gaps."""
         return self.__class__(
             moltype=self.moltype,
-            seq=self.moltype.strip_bad_and_gaps(self),
+            seq=self.moltype.strip_bad_and_gaps(str(self)),
```
GavinHuttley, line 349: add comment on same issue as above: `str(self)` versus `numpy.array(self)`

src/cogent3/core/new_sequence.py
```diff
@@ -327,84 +329,80 @@ def strip_degenerate(self):
...
         )
 
-    def is_gapped(self):
+    def is_gapped(self) -> bool:
         """Returns True if sequence contains gaps."""
-        return self.moltype.is_gapped(self)
+        return self.moltype.is_gapped(str(self))
 
-    def is_gap(self, char=None):
+    def is_gap(self, char: str = None) -> bool:
```
GavinHuttley, line 357: checking how often this method used, I suspect we can just delete this method and associated tests

src/cogent3/core/new_sequence.py
```diff
@@ -1113,34 +1117,51 @@ def add_feature(
...
+        else:
+            # assume converting between ASCII/BYTES and nucleic acid
+            seq = str(self)
+
+        if not moltype.most_degen_alphabet().is_valid(seq):
+            raise ValueError(
+                f"Changing from old moltype={self.moltype.label} to new "
+                f"moltype={moltype.label} is not valid for this data"
+            )
+        sv = SeqView(seq=seq, alphabet=moltype.most_degen_alphabet())
```
GavinHuttley, line 1160: add a comment to the docstring that if this method is applied to a sequence that is part of a sequence collection, the result is no longer part of that collection.

src/cogent3/core/new_sequence.py
```diff
@@ -1699,7 +1711,7 @@ def complement(self):
         will return list of keys.
         """
         return self.__class__(
-            moltype=self.moltype, seq=self.moltype.complement(self), info=self.info
+            moltype=self.moltype, seq=self.moltype.complement(str(self)), info=self.info
```
GavinHuttley, line 1714: everywhere you are using `str(self)` add a comment about checking using either `array(self)`.

GavinHuttley (changes requested):
massive amount of work!! mainly trivial comments to be added. One thing to note is that everywhere you use `str(self)` on some objects, `numpy.array` is probably available too. Se we should add a comment about checking performance of those.

src/cogent3/core/new_sequence.py
KatherineCaley: the underlying methods uses `bytes` so i've changed it to that

PR #1942, opened 2024-07-05
TST: Port Sequence unittest classes

fredjaya:
- 3 new xfails
- 3 xpasses removed
- @KatherineCaley could you check if any old ArraySeq-related tests in the following classes should be ported? If not, I think it's the last of the Sequence tests!
    - DnaSequenceTests
    - CodonSequenceTests
    - DnaSequenceGapTests
    - SequenceIntegrationTests
    - ModelSequenceSequenceTests

### DIFF
--- a/tests/test_core/test_new_sequence.py
+++ b/tests/test_core/test_new_sequence.py
@@ -189,7 +189,7 @@ def test_add_bad():
     mt = new_moltype.DNA
     seq1 = mt.make_seq(seq="AAA")
     with pytest.raises(new_alphabet.AlphabetError):
-        f"{seq1}s8d3j%31 s-']"
+        _ = f"{seq1}s8d3j%31 s-']"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


tests/test_core/test_new_sequence.py
```diff
@@ -174,6 +175,51 @@ def test_sequence_strip_degenerate(seq, expect):
...
+    assert got == "AAACCC"
+    assert got.moltype == mt
+
+
+@pytest.mark.xfail(reason="No error when adding sequences")
+def test_add_bad():
+    mt = new_moltype.DNA
+    seq1 = mt.make_seq(seq="AAA")
+    with pytest.raises(new_alphabet.AlphabetError):
+        _ = f"{seq1}s8d3j%31 s-']"
```
KatherineCaley, line 192: This should be changed to actually adding (with `+`) the sequence to the bad string. Including a variable in an f-string exercises whether the variable can be formatted into a string - and since we define a `__str__` method for Sequences we do not expect this to fail. 

To ensure add fails for bad sequences we need to add a validity check in the `__add__` method. 

```py 

if not self.moltype.most_degen_alphabet().is_valid(str(other)):
    raise new_alphabet.AlphabetError(f"Invalid sequence characters in other for moltype={self.moltype.label}")
```

KatherineCaley (changes requested):
I think you're right to leave the array seq tests, the rest looks good - have suggested changing one test and method

👾

KatherineCaley (approved):

PR #1947, opened 2024-07-15
MAINT: refactoring sequence parsing

GavinHuttley:
Sequence parsing is mediated by classes in parse/sequence.py which assumes all parsers
are line-based. These preclude more efficient algorithms. This patch replaces those with
a single class that is simpler and wraps the current parsers.

The patch also includes a fasta parser that is quicker than the line-based current ones.

Note: the new parser is not yet a drop-in replacement.


src/cogent3/parse/fasta.py
```diff
@@ -376,5 +381,82 @@ def GroupFastaParser(
...
+    data, converter: OptConverter = None
+) -> typing.Iterable[typing.Tuple[str, OutTypes]]:
+    """generator returning sequence labels and sequences converted bytes from a fasta file
+
+    Parameters
+    ----------
+    path
+        location of the fasta file
+    converter
+        a callable that uses converts sequence characters into nominated bytes,
```
KatherineCaley, line 413: "that uses converts sequence" ?

PR #1950, opened 2024-07-16
DEV: preps and documentation of new_type 

KatherineCaley:
DEV:
Misc changes to find middle ground between old and new API, i.e, 
- fixing occurrences of `make_seq` which do not use keyword args 
- fixing occurrences of creating seqs without a moltype 
- using `get_moltype` instead of directly importing a moltype in order to allow code to respect setting `new_type=True` with an environment variable 

DOC
- documentation of the existence of new API


GavinHuttley (approved):
Huge amount of detailed work -- thank you @KatherineCaley !!

PR #1961, opened 2024-07-23
ENH: begin Alignment refactor

KatherineCaley:
In this PR: 
- re-add `AlignedData` class and helper code- thanks @fredjaya!
- add `Aligned` class which wraps an `AlignedDataView`
- add `Alignment` and tests of its basic usage

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -2642,8 +2642,8 @@ class AlignedSeqsData(AlignedSeqsDataABC):
         self._seqs = {}
         for k, v in seqs.items():
             seq = self._alphabet.to_indices(v)
-            seq.flags.writeable = False
             self._seqs[k] = seq
+            seq.flags.writeable = False
         self._gaps = {}
         for k, v in gaps.items():
             self._gaps[k] = v
@@ -2792,6 +2792,8 @@ class AlignedSeqsData(AlignedSeqsDataABC):
         if len(seq) == self.align_len:
             return seq[start:stop]
 
+        # refactor: design
+        # direct numpy array approach would be faster
         gaps = self._gaps[seqid]
         indel_map = IndelMap(
             gap_pos=gaps[:, 0], cum_gap_lengths=gaps[:, 1], parent_length=self.align_len
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -2650,9 +2650,8 @@ class AlignedData:
         seq = self.seqs[seqid]
         gaps = self.gaps[seqid]
 
-        # If there are no gaps, gaps not be an instance of IndelMap
-        # and we can return sliced sequence
-        if not isinstance(gaps, IndelMap):
+        # if there's no gaps, just slice the sequence
+        if len(seq) == self.align_len:
             return seq[start:stop]
 
         new_map = gaps[start:stop] if start is not None or stop is not None else gaps
@@ -2673,9 +2672,8 @@ class AlignedData:
         seq = self.seqs[seqid]
         gaps = self.gaps[seqid]
 
-        # If there are no gaps, gaps not be an instance of IndelMap
-        # and we can return sliced sequence
-        if not isinstance(gaps, IndelMap):
+        # if there's no gaps, just slice the sequence
+        if len(seq) == self.align_len:
             return seq[start:stop]
 
         return self.moltype.most_degen_alphabet().gapped_by_map(seq, gaps)[start:stop]
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -6,6 +6,7 @@ import typing
 import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
+from dataclasses import InitVar, dataclass, field
 from functools import singledispatch, singledispatchmethod
 from pathlib import Path
 from typing import Any, Callable, Iterator, Mapping, Optional, Union
@@ -21,6 +22,7 @@ from cogent3.core.annotation_db import (
     SupportsFeatures,
 )
 from cogent3.core.info import Info as InfoClass
+from cogent3.core.location import IndelMap
 from cogent3.core.profile import (
     PSSM,
     MotifCountsArray,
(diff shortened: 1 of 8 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+        if span.step and span.step < 0:
+            raise NotImplementedError("Cannot reverse an Aligned")
+
+        return self.__class__(
+            data=self.data[span.start : span.stop], moltype=self.moltype
+        )
+
+
+@dataclass
+class AlignedData:
```
GavinHuttley, line 2547: Change the name to `AlignedSeqsData` and inherit from `SeqsDataABC`. (We also need to define an `AlignedSeqsDataABC`, which has only methods unique to this class.)

The `AlignedSeqsData` class should mirror the organisation of the `SeqsData` class. It should not be a `dataclass` (only because the use of slots with data classes is not supported at the minimal Python version that Cogent3 currently supports.), but a handcrafted class which has slots defined that are the same as those for `SeqsData`, with the addition of slots for gaps and anything else alignment specific

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+        # todo: kath
+        ...
+
+
+class AlignedDataView(new_sequence.SeqViewABC, new_sequence.SliceRecordABC):
+    # refactor: docstring
+    def __init__(
+        self,
+        *,
+        seq: AlignedData,
```
GavinHuttley, line 2767: I really think that we need to change this argument from seq to parent for all the view classes.

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+        self._offset = offset
+        self._seqid = seqid
+
+    @property
+    def seqid(self) -> str:
+        return self._seqid
+
+    @property
+    def seq_len(self) -> int:
+        return self._seq_len
```
GavinHuttley, line 2798: Are we distinguishing between aligned length and ungapped sequence length here?

KatherineCaley:
I will add tests shortly to address the coverage drop. The codacy failure is due to codacy not tracking the alignment-refactor branch - I have addressed this so future commits will be checked.

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+        self,
+        *,
+        seqid: str,
+        start: OptInt = None,
+        stop: OptInt = None,
+    ) -> str:
+        """Return sequence corresponding to seqid as a string.
+        start/stop are in alignment coordinates. Excludes gaps."""
+
+        return self.alphabet.from_indices(
```
GavinHuttley, line 2691: How is this method different to `get _gapped_seq_str`? It seems to me that those coordinates are alignment coordinates but the returned object is the underlying ungapped sequence. In that case you need the `IndelMap` object for conversion of alignment coordinates into sequence coordinates.

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+    def to_alphabet(self):
+        # todo: kath
+        ...
+
+    def to_rich_dict(self):
+        # todo: kath
+        ...
+
+
+class AlignedDataView(new_sequence.SeqViewABC, new_sequence.SliceRecordABC):
```
GavinHuttley, line 2762: Add the dunder methods for strings, bytes and arrays to this view class. (I note here that these dunder methods should be made abstract methods on `new_sequence. SeqViewABC`).

src/cogent3/core/new_sequence.py
```diff
@@ -1277,36 +1276,39 @@ def with_masked_annotations(
         return new
 
     def gapped_by_map_segment_iter(
-        self, segment_map, allow_gaps=True, recode_gaps=False
+        self, map, allow_gaps=True, recode_gaps=False
```
GavinHuttley, line 1279: If this is on a sequence, shouldn't it be returning an instance of self?

GavinHuttley (changes requested):
Awesome work!
Some interesting challenges ahead! I think we will wind up in working on this pull request for a little bit just to establish some core elements before we proceed too far.

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+        """Returns Sequence object, including gaps."""
+        # refactor: design
+        # should we mirror the above property and call this gapped_seq
+        # and make it a property?
+        return self.moltype.make_seq(
+            seq=self.data.gapped_str_value, name=self.data.seqid
+        )
+
+    def __str__(self) -> str:
+        return self.data.gapped_str_value
```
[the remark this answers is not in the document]
KatherineCaley: @GavinHuttley Can do, however would have thought it was inefficient to create a Sequence instance just to cast to str, array, or bytes.

src/cogent3/core/new_sequence.py
KatherineCaley: it iters the segments as str

src/cogent3/core/new_alphabet.py
```diff
@@ -7,11 +7,13 @@
 import numba
 import numpy
 
+from cogent3.core.location import IndelMap
```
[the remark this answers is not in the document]
KatherineCaley: okay! removed

src/cogent3/core/new_alignment.py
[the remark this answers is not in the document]
KatherineCaley: done

src/cogent3/core/new_alignment.py
KatherineCaley: @GavinHuttley `get_gapped_seq_str` includes gaps and `get_seq_str` does not. `get_seq_str` uses the `IndelMap` for coordinate conversion - (see `AlignedData.get_seq_array`)

src/cogent3/core/new_alignment.py
KatherineCaley: I've renamed to `parent_len` - its the length of the aligned

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+
+    def __array__(self) -> numpy.ndarray:
+        return self.data.gapped_array_value
+
+    def __bytes__(self) -> bytes:
+        return self.data.gapped_bytes_value
+
+    def __iter__(self):
```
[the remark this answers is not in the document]
KatherineCaley: where to handle reverse complement is a good question!

src/cogent3/core/new_alignment.py
```diff
@@ -2374,3 +2379,559 @@ def _(
...
+        )
+
+
+@dataclass
+class AlignedData:
+    # refactor: docstring
+    seqs: Optional[dict[str, StrORBytesORArray]]
+    gaps: Optional[dict[str, numpy.ndarray]]
+    moltype: MolTypes
+    make_seq: Optional[MakeSeqCallable] = None
```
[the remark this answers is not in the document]
KatherineCaley: It's not being used to make an `Aligned` instance, instead, being used by the `Aligned` instance to make a `Sequence`.

src/cogent3/core/new_sequence.py
KatherineCaley: the method below`gapped_by_map` return an instance of self

src/cogent3/core/new_alignment.py
```diff
@@ -2614,28 +2701,53 @@ def __getitem__(self, index: Union[str, int]) -> Aligned:
...
+
+    @make_seq.setter
+    def make_seq(self, make_seq: MakeSeqCallable) -> None:
+        self._make_seq = make_seq
+
+    @property
+    def align_len(self) -> int:
+        return self._align_len
+
+    def seq_lengths(self) -> dict[str, int]:
```
GavinHuttley, line 2739: it should return the length of the sequences, not the alignment length. Look at the method on `AlignmentI.get_seq_lengths()`

src/cogent3/core/new_alignment.py
```diff
@@ -2647,16 +2759,23 @@ def get_seq_array(
         """Return sequence corresponding to seqid as an array of indices.
         start/stop are in alignment coordinates. Excludes gaps.
         """
-        seq = self.seqs[seqid]
-        gaps = self.gaps[seqid]
-
+        seq = self._seqs[seqid]
```
GavinHuttley, line 2762: add "and the alignment coordinates are converted to sequence coordinates" in the docstring

src/cogent3/core/new_alignment.py
```diff
@@ -2666,17 +2785,31 @@ def get_gapped_seq_array(
...
-        seq = self.seqs[seqid]
-        gaps = self.gaps[seqid]
+        seq = self._seqs[seqid]
 
         # if there's no gaps, just slice the sequence
         if len(seq) == self.align_len:
             return seq[start:stop]
 
-        return self.moltype.most_degen_alphabet().gapped_by_map(seq, gaps)[start:stop]
+        gaps = self._gaps[seqid]
```
GavinHuttley, line 2797: add a comment "refactor: design". The iter over spans will be slower than a more direct numpy based approach, but it works!

src/cogent3/core/new_alignment.py
```diff
@@ -2732,25 +2865,31 @@ def get_gapped_seq_bytes(
 
     def add_seqs(
         self, seqs: dict[str, StrORArray], force_unique_keys=True
-    ) -> AlignedData:
-        """Returns a new AlignedData object with added sequences."""
+    ) -> AlignedSeqsData:
+        """Returns a new AlignedSeqsData object with added sequences."""
```
GavinHuttley, line 2869: assert that the new aligned sequences have the same `align_len` as self. Let the user concatenate things if they want to.

src/cogent3/core/new_alignment.py
```diff
@@ -2532,46 +2554,111 @@ def _(self, span: int):
...
         if span.step and span.step < 0:
             raise NotImplementedError("Cannot reverse an Aligned")
 
-        return self.__class__(
-            data=self.data[span.start : span.stop], moltype=self.moltype
-        )
+        return self.__class__(data=self.data[span.start : span.stop])
+
+
+class AlignedSeqsDataABC(ABC):
```
GavinHuttley, line 2565: this should inherit from `SeqsDataABC`

src/cogent3/core/new_alignment.py
```diff
@@ -2532,46 +2554,111 @@ def _(self, span: int):
...
+        *,
+        seqid: str,
+        start: OptInt = None,
+        stop: OptInt = None,
+    ) -> bytes: ...
 
 
-@dataclass
-class AlignedData:
+class AlignedSeqsData(SeqsDataABC, AlignedSeqsDataABC):
```
GavinHuttley, line 2608: only inherits from the new `AlignedSeqsDataABC`

src/cogent3/core/new_alignment.py
```diff
@@ -2870,19 +3030,28 @@ def __init__(
         super().__init__(**kwargs)
 
     @property
-    def seqs(self) -> AlignedData:
+    def seqs(self) -> AlignedSeqsData:
```
GavinHuttley, line 3033: return type should be the ABC

src/cogent3/core/new_alignment.py
```diff
@@ -2920,18 +3089,22 @@ def make_aligned_seqs(
...
+    alphabet = moltype.most_degen_alphabet()
+    aligned_data = AlignedSeqsData.from_aligned_seqs(
+        data=data, alphabet=alphabet, make_seq=moltype.make_seq
+    )
     return make_aligned_seqs(aligned_data, moltype=moltype, info=info)
 
 
 @make_aligned_seqs.register
-def _(data: AlignedData, *, moltype: str, info: dict = None) -> Alignment:
+def _(data: AlignedSeqsData, *, moltype: str, info: dict = None) -> Alignment:
```
GavinHuttley, line 3102: change type to `AlignedSeqsDataABC` and elsewhere

src/cogent3/core/new_sequence.py
```diff
@@ -2459,69 +2455,69 @@ class SeqView(SeqViewABC, SliceRecordABC):
...
-        "_seq_len",
+        "_parent_len",
     )
 
     def __init__(
         self,
         *,
-        seq: str,
+        parent: str,
         alphabet: new_alphabet.AlphabetABC,
```
GavinHuttley, line 2478: using the ABC for type hints is a good pattern

GavinHuttley (changes requested):
Respect! A huge effort. A few more mods.

GavinHuttley (approved):
looks nice. Need to walk through the tests more deliberately, but I suspect moving onto other methods will expose issue!!

Go ahead and merge.

KatherineCaley:
> looks nice. Need to walk through the tests more deliberately, but I suspect moving onto other methods will expose issue!!
> 
> Go ahead and merge.

I'll merge now and focus on improving the test coverage in the next PR which should hopefully tease out issues

PR #1970, opened 2024-07-31
ENH: select_translatable frame argument

GavinHuttley:
[NEW] users can now specify that all sequences in a collection are in
    the nominated frame.


GavinHuttley:
Note that the codacy warning is about the scriv markdown, it can be safely ignored since those are reformatted on release

KatherineCaley (approved):
looks good

Issue #1971, opened 2024-07-31
REL get github actions for doing a release


GavinHuttley:
see https://docs.pypi.org/trusted-publishers/using-a-publisher/

GavinHuttley:
fixed in e85eb9a

PR #1974, opened 2024-08-07
DEV: add release workflow

rmcar17:
Partial resolution of #1971. Adds a workflow which on manual trigger publishes to PyPI.

Remaining tasks:
- create a release environment
- add github actions as a trusted publisher for `cogent3` through pypi

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,93 @@
+name: Release
+
+on: [workflow_dispatch]
+
+jobs:
+  test:
+    name: "Test on Python ${{ matrix.python-version }} (${{ matrix.os }})"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 0
+
+      - uses: "actions/setup-python@v5"
+        with:
+            python-version: "${{ matrix.python-version }}"
+
+      - name: "Installs for ${{ matrix.python-version }}"
+        run: |
+          pip install --upgrade pip
+          pip install nox
+
+      - name: "Run nox for Python ${{ matrix.python-version }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      
+      - name: Install build dependency
```
GavinHuttley, line 44: change to flit instead of build

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+        run: |
+          pip install build
```
GavinHuttley, line 46: pip install flit

.github/workflows/release.yml
```diff
@@ -0,0 +1,94 @@
...
+          pip install --upgrade pip
+
+      - name: Build sdist and wheel
+        run: python -m build --wheel --sdist
```
GavinHuttley, line 50: change run command to `python -m flit build` (this will bundle everything tracked by git) and automatically builds both sdist and wheels.

GavinHuttley (changes requested):
awesome stuff! change to using flit since I know that automatically grabs everything into the tarball and wheel that's tracked by git

.github/workflows/release.yml
rmcar17: So what build does is it reads the `[build-system]` information in `pyproject.toml` and then uses flit to build the project. It means if you decide to change the build system for whatever reason the action will still work.

.github/workflows/release.yml
rmcar17: See other commment

.github/workflows/release.yml
rmcar17: `build` reads the `pyproject.toml`, detects flit as the backend and uses flit to build the wheel and sdist

.github/workflows/release.yml
GavinHuttley: good points

PR #2003, opened 2024-08-27
ENH: support slicing of Alignment/Aligned

KatherineCaley:
In this PR: 

- implementing changes to the design of aligned/alignment as discussed on 30/07
  - stop tracking slice history directly in `AlignedDataView`, now via a slice record object 
  - generally trying to mirror the `SeqsData` design more closely in `AlignedSeqsData` - especially with the assignment of `make_seq` and `make_aligned` methods
- Slicing an `Alignment` propagates to `Aligned` instances!

### DIFF
--- a/src/cogent3/core/location.py
+++ b/src/cogent3/core/location.py
@@ -1206,10 +1206,9 @@ class IndelMap(MapABC):
         adj_gaps = []
         lengths = self.get_gap_lengths()
 
-        # work out the first gap.
+        # determine the beginning of the slice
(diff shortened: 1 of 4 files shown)


KatherineCaley:
fyi, codacy complaining about: 
- the complexity of the `to_html` function - fair, I can add a refactor comment 
- that self.make_aligned is not callable (it is, it is just dynamically assigned)

GavinHuttley (approved):
massive amount of work. I'm approving this without fully understanding, we will chat about issues in person

PR #2013, opened 2024-09-04
ENH: add support for reversing and strides when indexing IndelMaps and Alignments, fixes #1967

### LINKED ISSUE #1967: ENH: implement support for strides in IndelMap
The following currently is not supported
```python
import numpy
from cogent3.core.location import IndelMap

im = IndelMap(gap_pos=numpy.array([3, 8]), gap_lengths=numpy.array([2, 1]), parent_length=11)
im[::3]
```
We need to support, this means adjusting the gap positions and the cum_lengths.

KatherineCaley:
in this PR: 
- support for slicing IndelMaps with stride < 0,  fixes #1967 
- support for slicing Alignments with stride < 0 
- support for negative indexing of Alignments

### DIFF
--- a/src/cogent3/core/location.py
+++ b/src/cogent3/core/location.py
@@ -1018,6 +1018,17 @@ def _step_adjusted_length(start, end, adj, step):
 def _input_vals_pos_step(
     seqlen: int, start: OptInt, stop: OptInt, step: int
 ) -> tuple[int, int, int]:
+    """returns standardised start, stop, step values for positive step slicing. 
+    
+    Notes
+    -----
+    The input values for start and stop can be +ve, -ve or None. The returned
+    start and stop are strictly +ve and within the sequence length, and if not
+    provided default to 0 and seqlen respectively. 
+
+    If the slice results in a zero-length slice, the returned start, stop, step
+    will be 0, 0, 1 respectively.
+    """
     start = 0 if start is None else start
     if start > 0 and start >= seqlen:
         # start beyond seq is an empty slice
@@ -1045,9 +1056,23 @@ def _input_vals_pos_step(
 def _input_vals_neg_step(
     seqlen: int, start: OptInt, stop: OptInt, step: int
 ) -> tuple[int, int, int]:
+    """returns standardised start, stop, step values for negative step slicing.
+    
+    Notes
+    -----
+    The input values for start and stop can be positive, negative or None. The
+    returned start and stop are strictly negative and correspond to indices 
+    within the sequence length. If not provided, start defaults to -1 and stop  
+    to -seqlen-1.
+
+    If the slice results in a zero-length slice, the returned start, stop, step
+    will be 0, 0, 1 respectively.
+    """
     # Note how Python reverse slicing works
     # we need to make sure the start and stop are both
     # negative, for example "abcd"[-1:-5:-1] returns "dcba"
+    # returning the complete string in reverse order is only 
+    # possible with negative indexing
     if start is None or start >= seqlen:  # set default
         start = -1  # Done
     elif start >= 0:  # convert to -ve index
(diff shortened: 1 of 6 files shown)


tests/test_core/test_new_alignment.py
```diff
@@ -2938,14 +2938,49 @@ def test_alignment_repr():
...
+    expect = aligned_dict[seqid][start:stop:step].replace("-", "")
+    assert got == expect
+
+
+@pytest.mark.parametrize("start", range(6))
+@pytest.mark.parametrize("stop", range(6))
+@pytest.mark.parametrize("step", range(-3, 0))
+@pytest.mark.parametrize("seqid", ("seq1", "seq2", "seq3", "seq4"))
+def test_alignment_slice_neg_step_gapped(aligned_dict, start, stop, step, seqid):
+    """slicing an alignment should propogate the slice to aligned instances"""
```
GavinHuttley, line 2968: don't these tests always create an empty record? That is, the following is True
```python
assert "ACGGTG"[1:3:-3] == "
```
You would need to reverse your ranges in the fixtures I think

GavinHuttley (changes requested):
I'm not sure your tests for negative stride slicing are correct. For slicing with a negative stride to work, start > stop is required.

tests/test_core/test_new_alignment.py
KatherineCaley: I'm not sure that reversing the range of the fixture will change anything. Pytest will do the combination of all fixtures. You're right that about 2/3rds of the combinations will result in an empty slice - but I thought it was worth checking that in both cases that were valid slices and not valid slices, the result was the same.

KatherineCaley:
> I'm not sure your tests for negative stride slicing are correct. For slicing with a negative stride to work, start > stop is required.

I figured the gain from comprehensive testing was worth it (great for catching edge cases) outweighed the large number of tests resulting in empty slices. I am happy to change strategies to test a more curated set of input values if you'd prefer.

tests/test_core/test_new_alignment.py
GavinHuttley: Gotcha, I did not fully think this through. As all permutations of start / stop are being produced, sometimes start > stop.

GavinHuttley (changes requested):
Tiny changes.

This remark is for my(present and future)self: Docstrings use memory. I think commenting on code (i.e. `# some comment`) should be used instead of docstrings when the text is about implementation details or assumptions, Docstrings should communicate the intent of the function and control of that intent via arguments..

src/cogent3/core/location.py
```diff
@@ -1017,6 +1015,82 @@ def _step_adjusted_length(start, end, adj, step):
...
+
+    If the slice results in a zero-length slice, the returned start, stop, step
+    will be 0, 0, 1 respectively.
+    """
+    start = 0 if start is None else start
+    if start > 0 and start >= seqlen:
+        # start beyond seq is an empty slice
+        return 0, 0, 1
+
+    stop = seqlen if stop is None else stop
```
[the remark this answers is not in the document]
KatherineCaley: For stop, using `or` as you suggest will introduce a bug if the stop is 0 - it will evaluate to False.

src/cogent3/core/location.py
```diff
@@ -999,15 +999,15 @@ def _update_lengths(
     result_lengths[result_indices] += gap_lengths[other_indices]
 
 
-def _step_adjustment(gap_start, start, step):
+def _step_adjustment(gap_start: int, start: int, step: int):
```
GavinHuttley, line 1002: eek, missing return type!

GavinHuttley (changes requested):
final change I promise!

src/cogent3/core/location.py
KatherineCaley: oops!

KatherineCaley:
@GavinHuttley hopefully ready to go

GavinHuttley (approved):
awesome!!

PR #2024, opened 2024-09-18
ENH: IndelMap gets methods for identifying and removing common gaps

GavinHuttley:
[NEW] IndelMap.shared_gaps() takes either another IndelMap instance,
     or a numpy array which consists of start, end coordinates that
     define gap locations in alignment coordinates. It returns a
     numpy array with start, end coordinates for shared gaps. This
     method relies on two new functions, coords_intersect() and
     span_and_span() the latter defining the intersection for two
     spans.
     The result of this method can be applied to multiple IndelMap
     instances from an alignment so that all common gaps can be
     identified and then removed for each case.

[NEW] IndelMap.minus_gaps() takes either another IndelMap instance,
     or a numpy array which consists of start, end coordinates that
     define gap locations in alignment coordinates. It returns a
     new IndelMap instance after removing shared gaps. It uses
     a new coords_minus_coords() function.

### DIFF
--- a/src/cogent3/core/location.py
+++ b/src/cogent3/core/location.py
@@ -1682,7 +1682,7 @@ def coords_minus_coords(
     Parameters
     ----------
     coords1, coords2
-        series of coordinates defining segments
+        coordinates defining segments as [(gap start, gap end)..]
 
     Notes
     ------
@@ -1770,7 +1770,7 @@ def coords_intersect(
     Parameters
     ----------
     coords1, coords2
-        series of coordinates defining segments
+        coordinates defining segments as [(gap start, gap end)..]
 
     Notes
     ------
--- a/src/cogent3/core/location.py
+++ b/src/cogent3/core/location.py
@@ -1592,7 +1592,7 @@ class IndelMap(MapABC):
         return FeatureMap(spans=spans, parent_length=self.parent_length)
 
     @functools.singledispatchmethod
-    def shared_gaps(self, other) -> numpy.ndarray:
+    def shared_gaps(self, other: Union["IndelMap", numpy.ndarray]) -> numpy.ndarray:
         """returns a numpy array of the shared [(gap start, gap end), ...]
 
         Notes
@@ -1630,7 +1630,7 @@ class IndelMap(MapABC):
         return numpy.array(result, dtype=_DEFAULT_GAP_DTYPE)
 
     @functools.singledispatchmethod
-    def minus_gaps(self, other_gaps):
+    def minus_gaps(self, other_gaps: Union["IndelMap", numpy.ndarray]):
         """returns new map with gaps in other_gaps removed from self
 
         Parameters
--- a/src/cogent3/core/location.py
+++ b/src/cogent3/core/location.py
@@ -6,7 +6,6 @@ import json
 from abc import ABC, abstractmethod
 from bisect import bisect_left, bisect_right
 from functools import total_ordering
-from itertools import chain
 from typing import Any, Iterator, Optional, Sequence, Union
 
 import numpy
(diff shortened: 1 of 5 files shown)


src/cogent3/core/location.py
```diff
@@ -1596,6 +1591,202 @@ def make_seq_feature_map(self, align_feature_map: "FeatureMap") -> "FeatureMap":
 
         return FeatureMap(spans=spans, parent_length=self.parent_length)
 
+    @functools.singledispatchmethod
+    def shared_gaps(self, other) -> numpy.ndarray:
```
KatherineCaley, line 1595: to be consistent with above, type hint `, other: "IndelMap"`

src/cogent3/core/location.py
```diff
@@ -1596,6 +1591,202 @@ def make_seq_feature_map(self, align_feature_map: "FeatureMap") -> "FeatureMap":
...
+            raise AssertionError(
+                f"other_gaps {other_gaps!r} from a different alignment?"
+            )
+        self_gaps = self.get_gap_align_coordinates()
+
+        result = coords_intersect(self_gaps, other_gaps)
+        return numpy.array(result, dtype=_DEFAULT_GAP_DTYPE)
+
+    @functools.singledispatchmethod
+    def minus_gaps(self, other_gaps):
```
KatherineCaley, line 1633: `def minus_gaps(self, other_gaps: Union["IndelMap", numpy.ndarray ): -> "IndelMap"`

src/cogent3/core/location.py
```diff
@@ -1596,6 +1591,202 @@ def make_seq_feature_map(self, align_feature_map: "FeatureMap") -> "FeatureMap":
...
+        """
+        if len(self) != len(other_gaps):
+            raise AssertionError(
+                f"{len(other_gaps)=} != {len(self)=}, from a different alignment?"
+            )
+
+        return self.minus_gaps(other_gaps.get_gap_align_coordinates())
+
+    @minus_gaps.register
+    def _(self, other_gaps: numpy.ndarray):
```
KatherineCaley, line 1653: `-> "IndelMap"`

src/cogent3/core/location.py
```diff
@@ -1596,6 +1591,202 @@ def make_seq_feature_map(self, align_feature_map: "FeatureMap") -> "FeatureMap":
...
+
+def coords_minus_coords(
+    coords1: numpy.ndarray, coords2: numpy.ndarray
+) -> numpy.ndarray:
+    """returns the coords1 minus any overlaps with coords2
+
+    Parameters
+    ----------
+    coords1, coords2
+        series of coordinates defining segments
```
KatherineCaley, line 1685: can you make it clearer that coords will be in the form [(gap start, gap end)..]

src/cogent3/core/location.py
```diff
@@ -1596,6 +1591,202 @@ def make_seq_feature_map(self, align_feature_map: "FeatureMap") -> "FeatureMap":
...
+    Notes
+    ------
+    Assumes both coordinate series are sorted.
+    """
+    dtype = getattr(coords1, "dtype", _DEFAULT_GAP_DTYPE)
+    unique_segments = []
+    for a1, a2 in coords1:
+        total_intersect = None
+        for b1, b2 in coords2:
+            if b2 < a1:
```
KatherineCaley, line 1696: not sure why here is strictly less than, where below it is less than or equal to

src/cogent3/core/location.py
GavinHuttley: There is a limitation with python3.9, it raises:
```
NameError: name 'IndelMap' is not defined
```

src/cogent3/core/location.py
GavinHuttley: can't for reason above

src/cogent3/core/location.py
GavinHuttley: a1 and b1 are "starts", a2 and b2 are "ends. This condition is the end of segment *b* is before the start of segment *a*. Because a2 is not inclusive, the condition below is saying segment *a* and *b* do not intersect

PR #2029, opened 2024-09-30
ENH: better construction of dict arrays and distance matrices

GavinHuttley:


src/cogent3/core/new_sequence.py
```diff
@@ -2596,42 +2636,48 @@ def copy(self, sliced: bool = False):
 
 
 @singledispatch
-def _coerce_to_seqview(data, seqid, alphabet) -> SeqViewABC:
+def _coerce_to_seqview(data, seqid, alphabet, offset) -> SeqViewABC:
```
KatherineCaley, line 2639: in the old sequence the new argument is called `annotation_offset`, otherwise the changes are the same

KatherineCaley (approved):

src/cogent3/__init__.py
```diff
@@ -323,7 +332,7 @@ def load_seq(
     name, seq = data[0]
     name = label_to_name(name) if label_to_name else name
 
-    result = make_seq(seq, name, moltype=moltype, new_type=new_type)
+    result = make_seq(seq, name, moltype=moltype, new_type=new_type, **kw)
```
KatherineCaley, line 335: should you be passing `annotation_offset` to `make_seq` here?

src/cogent3/core/new_sequence.py
GavinHuttley: Given the old module is destined for removal, I'm going to let that inconsistency fade gracefully into oblivion!

src/cogent3/__init__.py
GavinHuttley: Good catch!

PR #2036, opened 2024-10-08
ENH: continued Alignment refactor 

KatherineCaley:
Major design change in this PR:
- DEV: `Seq[Data]View` no longer inherits from `SliceRecord`
- ENH: added intermediate class `_IndexableSeqs` which handles the construction of `Aligned` instances
- DEV: generally following the paradigm that all methods that are from `SeqsDataABC` should work in sequence coordinates and all methods unique to `AlignedSeqsDataABC` should work in aligned coordinates

Note: the drop in coverage is from the addition of feature querying methods for which the tests are still xfail


GavinHuttley (approved):
well that was easy!

GavinHuttley:
Um, hold off. Seems GitHub didn't show me all the changed files. Looking at the rest now.

src/cogent3/core/new_alignment.py
```diff
@@ -100,48 +106,25 @@ class SeqDataView(new_sequence.SeqViewABC, new_sequence.SliceRecordABC):
...
 
-    __slots__ = ("parent", "start", "stop", "step", "_offset", "_seqid", "_parent_len")
+    __slots__ = ("parent", "alphabet", "_seqid", "_parent_len", "_slice_record")
 
     def __init__(
         self,
         *,
         parent: SeqsData,
-        seqid: str,
+        alphabet: new_alphabet.CharAlphabet,
```
KatherineCaley, line 115: Note to self: I dont think `SeqsDataView` need alphabet, just because `SeqView` has it... its not being used

src/cogent3/core/new_alignment.py
```diff
@@ -3096,174 +3315,881 @@ def gapped_bytes_value(self) -> bytes:
...
+        seqs = ", ".join(seqs)
+
+        return f"{len(self.names)} x {len(self)} {self.moltype.label} alignment: {seqs}"
+
+    def __len__(self):
+        return len(self._seqs_data)
+
+    def _make_aligned(self, seqid: str) -> Aligned:
+        # we set the slice record on the AlignedDataView
+        data = self._seqs_data.get_view(seqid)
```
GavinHuttley, line 3496: add a comment to redesign. We will add slice_record as an argument to the `get_view()` method.

src/cogent3/core/new_alignment.py
```diff
@@ -3096,174 +3315,881 @@ def gapped_bytes_value(self) -> bytes:
...
+        data.slice_record = self._slice_record
+        return Aligned(data=data, moltype=self.moltype)
+
+    @property
+    def seqs(self) -> AlignedSeqsDataABC:
```
GavinHuttley, line 3501: Add a comment about redesign. We will make `_seqs` a private attribute and the indexable seqs class will be assigned to it in the constructor. The property will then just return the `_seqs` attribute.

src/cogent3/core/new_alignment.py
```diff
@@ -3096,174 +3315,881 @@ def gapped_bytes_value(self) -> bytes:
...
+        arr = self.array_value
         if dtype:
             arr = arr.astype(dtype)
         return arr
 
     def __bytes__(self) -> bytes:
-        return self.gapped_bytes_value
+        return self.bytes_value
+
+    def __getitem__(self, segment) -> new_sequence.SeqViewABC:
```
GavinHuttley, line 3329: Shouldn't the return type should be the aligned seq view abstract base class?

src/cogent3/core/new_alignment.py
```diff
@@ -3096,174 +3315,881 @@ def gapped_bytes_value(self) -> bytes:
...
+    def get_gap_array(self, include_ambiguity: bool = True) -> numpy.ndarray:
+        """returns bool array with gap state True, False otherwise
+
+        Parameters
+        ----------
+        include_ambiguity
+            if True, ambiguity characters that include the gap state are
+            included
+        """
+        result = numpy.full((self.num_seqs, len(self)), False, dtype=bool)
```
GavinHuttley, line 3640: Add a comment about redesign. The method for returning a gap array should be on the IndelMap class.

GavinHuttley (changes requested):
just a few comments

src/cogent3/core/new_alignment.py
KatherineCaley: indeed it should!

src/cogent3/core/new_alignment.py
KatherineCaley: This would only work for the case where `include_ambiguity` is False (it defaults to True), otherwise `IndelMap` would need to know about the seq and alphabet

GavinHuttley (approved):
neeext!

PR #2037, opened 2024-10-09
API: _IndexableSeqs class API changes

GavinHuttley:
[NEW] allow specifying the method name to be invoked in the parent

[NEW] _IndexableSeqs supports len() and repr()

[NEW]_IndexableSeqs suports indexing by int or string

[CHANGED] SequenceCollection no longer assigns to SeqsData.make_seq

[CHANGED] SequenceCollection gets a _post_init method invoked at the end
    of the constructor. Overridden in Alignment so behaviour is customised.
    The method assigns the _IndexableSeqs instance to a new ._seqs attribute.
    The .seqs property is now returns ._seqs

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -662,13 +662,30 @@ class SequenceCollection:
         self._seqs_data = seqs_data
         self.moltype = moltype
         self.names = names
-        self._seqs_data.make_seq = self.moltype.make_seq
         if not isinstance(info, InfoClass):
             info = InfoClass(info) if info else InfoClass()
         self.info = info
         self.source = source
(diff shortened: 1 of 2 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -662,13 +662,30 @@ def __init__(
...
+    def _post_init(self):
+        # override in subclasses
+        self._seqs = _IndexableSeqs(self, make_seq=self._make_seq)
+
+    def _make_seq(self, name: str) -> new_sequence.Sequence:
+        sv = self._seqs_data.get_view(name)
+        return self.moltype.make_seq(seq=sv, name=name)
+
+    @property
+    def seqs(self) -> AlignedSeqsDataABC:
```
KatherineCaley, line 683: return type should be `SeqsDataABC`

KatherineCaley (approved):

PR #2039, opened 2024-10-10
API: remove strand arg, modify offset handling for Alignments

KatherineCaley:
API: remove strand as arg to seq collection constructor 
 - the way I had added tracking of the strand was not practical, considering that for a sequence, its strand is the sign of its step.

DEV: change how offset is tracked
- ensure that when a sequence is created from an alignment with an offset, this is propagated to the sequence instance - this is only necessary when the abs(step) == 1, otherwise we don't need it for annotation handling 

@GavinHuttley keen for your feedback about the direction of this PR because strand/offset are both important for feature querying

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -3382,7 +3382,7 @@ class _IndexableSeqs:
 
     def __repr__(self) -> str:
         one_seq = self[self.parent.names[0]]
-        return f"({one_seq!r}, + {self.parent.num_seqs -1} seqs)"
+        return f"({one_seq!r}, + {self.parent.num_seqs-1} seqs)"
 
     def __len__(self) -> int:
         return self.parent.num_seqs
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -393,7 +393,7 @@ class SeqsData(SeqsDataABC):
         return self.get_seq_str(seqid=seqid, start=start, stop=stop).encode("utf8")
 
     def get_view(self, seqid: str) -> SeqDataView:
-        seq_len = self.seq_lengths()[seqid]
+        seq_len = len(self._data[seqid])
         offset = self._offset.get(seqid, 0)
         slice_record = new_sequence.SliceRecord(
             step=-1 if self.is_reversed else 1, parent_len=seq_len, offset=offset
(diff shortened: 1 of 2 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -424,7 +393,7 @@ def get_seq_bytes(
         return self.get_seq_str(seqid=seqid, start=start, stop=stop).encode("utf8")
 
     def get_view(self, seqid: str) -> SeqDataView:
-        seq_len = len(self._data[seqid])
+        seq_len = self.seq_lengths()[seqid]
```
GavinHuttley, line 396: I would just directly access the attribute rather than the method. Also, I think we may be tying the API too closely to our implementation. How about we have a method `get_seq_length(name)` which users can get access to. This means the implementation does not have to create a dictionary and return it. Or worse yet, create a dictionary (which is mutable) and the user modifies it and gets unexplained behaviour 😳

Just add that as a design note on the method, don't implement yet

src/cogent3/core/new_alignment.py
```diff
@@ -2382,7 +2333,7 @@ def make_unaligned_seqs(
     info: dict = None,
     source: OptPathType = None,
     annotation_db: SupportsFeatures = None,
-    strand: dict[str, int] = None,
+    offset: dict[str, int] = None,
```
GavinHuttley, line 2336: as this is for many sequences (potentially) make a comment about making the argument and attribute name plural

add a comment about documenting the argument too

src/cogent3/core/new_alignment.py
```diff
@@ -3632,6 +3566,8 @@ def majority_consensus(self) -> new_sequence.Sequence:
             pos = CategoryCounter(pos)
             states.append(pos.mode)
 
+        # refactor: design
+        # should this return an Aligned object?
```
GavinHuttley, line 3570: i think it's ok for it to be a Sequence since it's primarily for summarising alignment column contents and it can do so as a Sequence just fine.

GavinHuttley (changes requested):
I don't know how properties like offsets etc.. will be used yet until we try to implement the feature handling. Just add the comments I've suggested and we will then start back on implementing that.

GavinHuttley (approved):
Bonza 💪!

PR #2050, opened 2024-10-20
ENH: add function to split the internal name field in newick into name and support value, fixes #2045

### LINKED ISSUE #2045: ENH: extend support for bootstrap values from newick strings
`cogent3` stores support values (e.g. from a bootstrap) as `<phylonode>.params["support"]` entry. This information is also stored in some versions of newick strings. For example, IQTree uses `((a,b)##:0.1,..` and `((a,b)node_name/##:0.1,..` where `##` is an integer to record support measures. The current cogent3 parser treats both those values as name. We need to add code that correctly transforms these names into the name and support values.

To address this issue, write a function `split_name_and_support(field: str | None) -> tuple[str | None, float | None)` in `cogent3.core.tree`.

It should cope with them being either a name and no support, support and no name, or both. Specific cases are:

- field is `None` or an empty string, returns `None, None`
- field is `"edge.98/24"` returns `"edge.98", 24.0`
- field is `"edge.98"` returns `"edge.98", None`
- field is `"24"` returns `None, 24.0`

YapengLang:
#2045

## Summary by Sourcery

Add a new function to split internal node name fields in Newick format into separate name and support values, and implement corresponding tests to ensure correct functionality.

New Features:
- Introduce a function `split_name_and_support` to handle Newick format internal node name fields that contain both a name and a support value.

Tests:
- Add parameterized tests for the `split_name_and_support` function to verify its behavior with various input cases, including valid and invalid formats.

### DIFF
--- a/src/cogent3/core/tree.py
+++ b/src/cogent3/core/tree.py
@@ -2206,7 +2206,7 @@ def split_name_and_support(name_field: str | None) -> tuple[str | None, float | 
     # handle case where mutiple '/' in the name field
     elif len(support) > 1:
         raise ValueError(
-            f"Support value at node: {name!r} should be int or float not {"/".join(support)!r}."
+            f"Support value at node: {name!r} should be int or float not {'/'.join(support)!r}."
         )
     else:
         support_value = None
--- a/src/cogent3/core/tree.py
+++ b/src/cogent3/core/tree.py
@@ -27,6 +27,7 @@ Definition of relevant terms or abbreviations:
 
 from __future__ import annotations
 
+import contextlib
 import json
 import numbers
 import re
@@ -1696,6 +1697,7 @@ class PhyloNode(TreeNode):
             params["length"] = length
         kwargs["params"] = params
         super(PhyloNode, self).__init__(*args, **kwargs)
+        # split name into name/support here
 
     def _set_length(self, value):
         if not hasattr(self, "params"):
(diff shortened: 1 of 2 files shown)


src/cogent3/core/tree.py
```diff
@@ -2179,6 +2179,39 @@ def set_max_tip_tip_distance(self):
...
+        name = parts[0]
+        # the support value should be either a float or an int
+        try:
+            support = float(parts[1])
+        except ValueError as e:
+            raise ValueError(
+                f"Support value at node: {name} should be int or float."
+            ) from e
+        return name, support
+    elif len(parts) == 1:
```
GavinHuttley, line 2203: This code can be made much simpler with the following.
```python
name, *support = name_field.split("/")
name = name or None
support = support or None
try:
    support = float(support)
except (ValueError, TypeError) as e:
    raise ValueError(
                f"Support value at node: {name!r} should be int or float not {support!r}."
            ) from e
```

that covers all your cases I believe

tests/test_core/test_tree.py
```diff
@@ -2332,3 +2333,27 @@ def test_load_tree_bad_encoding():
...
+        ("edge.98/24", ("edge.98", 24.0)),
+        ("edge.98", ("edge.98", None)),
+        ("24", (None,24.0)),
+    ),
+)
+def test_split_name_and_support(name, expected):
+    assert split_name_and_support(name) == expected
+
+
+def test_split_name_and_support_invalid_support():
```
GavinHuttley, line 2352: I would merge this and the following test using a `parametrize`, e.g.
```python
@pytest.mark.parametrize("invalid",("edge.98/invalid", "edge.98/23/invalid"))
def test_split_name_and_support_invalid_support(invalid):
    with pytest.raises(ValueError):
        split_name_and_support(invalid)
```

tests/test_core/test_tree.py
```diff
@@ -2332,3 +2333,27 @@ def test_load_tree_bad_encoding():
...
+    with pytest.raises(ValueError):
+        split_name_and_support("edge.98/invalid")
+
+
+def test_split_name_and_support_invalid_name():
+    with pytest.raises(ValueError):
+        split_name_and_support("edge.98/24/invalid")
```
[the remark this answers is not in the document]
GavinHuttley: @YapengLang add this check to the end of your revised test

src/cogent3/core/tree.py
```diff
@@ -2221,6 +2254,8 @@ def edge_from_edge(self, edge, children, params=None):
 
     def create_edge(self, children, name, params, name_loaded=True):
         """Callback for newick parser"""
+        # TODO: split name and support by forward slash if present
```
GavinHuttley, line 2257: Delete this comment. We will use the function in the next PR

GavinHuttley (changes requested):
Looking good @YapengLang ! But we can make the code simpler

GavinHuttley (commented):
Awesome, thanks @YapengLang !

PR #2054, opened 2024-10-22
ENH: extract support measures on tree nodes from newick strings

GavinHuttley:
[NEW] handled automatically while parsing, uses the new `split_name_and_support()`
    function.

[NEW] added scriv changelog snippet

[CHANGED] modernised length property and no longer inherit from
    object

## Summary by Sourcery

Enhance the tree parsing functionality to extract and store node support measures from Newick strings. Modernize the TreeNode class by removing the object inheritance and updating the length property. Add tests for the new parsing functionality and update the documentation with a changelog entry.

New Features:
- Add support for parsing node support measures from Newick strings in tree objects.

Enhancements:
- Modernize the TreeNode class by removing inheritance from object and updating the length property to use Python's property decorator.

Documentation:
- Add a new changelog entry describing the enhancement of parsing node support from Newick format.

Tests:
- Add tests for the new functionality of parsing node support measures from Newick strings.

### DIFF
--- a/changelog.d/20241023_080011_Gavin.Huttley.md
+++ b/changelog.d/20241023_080011_Gavin.Huttley.md
@@ -1,0 +1,41 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+### Contributors
+
+- @YapengLang added the key function for parsing support from newick format
+  to tree object 💪
+
+
+### ENH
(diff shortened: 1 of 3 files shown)


GavinHuttley:
Noting here that the sourcery bot is incorrect. We do not call `split_name_and_support()` for tip nodes. Also, the contact issue is about the scriv Markdown, so ignoring it.

PR #2062, opened 2024-10-25
ENH: update handling of names for `SequenceCollection`

KatherineCaley:
updating the names of sequences in an alignment or sequence collection updates the name map between collection and seqsdata, the latter remaining unchanged

## Summary by Sourcery

Enhance the SequenceCollection class by implementing a name map to manage sequence names, ensuring that renaming sequences updates the name map without altering the underlying seqs_data. Update tests to reflect these changes.

Enhancements:
- Update the handling of sequence names in SequenceCollection by introducing a name map to maintain a mapping between collection names and seqs_data names.

Tests:
- Remove tests related to setting sequence names directly and add tests to ensure sequence names can be renamed multiple times without altering the underlying seqs_data.

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -374,25 +374,6 @@ class SeqsData(SeqsDataABC):
         else:
             raise ValueError(f"provided {names=} not found in collection")
 
-    def rename_seqs(self, renamer: Callable[[str], str]) -> SeqsData:
-        # todo: kath, create a map between original seqid and current seqname for feature querying?
-        renamed_data = {}
-        renamed_offset = {}
-
-        for name in self._data:
-            new_name = renamer(name)
-            renamed_data[new_name] = self._data[name]
-            if name in self._offset:
-                renamed_offset[new_name] = self._offset[name]
-
-        return self.__class__(
-            data=renamed_data,
-            alphabet=self.alphabet,
-            offset=renamed_offset,
-            reversed=self._reversed,
-            check=False,
-        )
-
     def add_seqs(
         self,
         seqs: dict[str, StrORBytesORArray],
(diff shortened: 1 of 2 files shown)


tests/test_core/test_new_alignment.py
```diff
@@ -1982,6 +1957,20 @@ def test_sequence_collection_rename_seqs():
...
+    assert set(new._seqs_data.names) == set(data)
+
+
+def test_sequence_collection_subsequent_rename():
+    """sequences can be renamed multiple times"""
+    data = {"seq1": "ACGTACGTA", "seq2": "ACCGAA---", "seq3": "ACGTACGTT"}
+    seqs = new_alignment.make_unaligned_seqs(data, moltype="dna")
+    new = seqs.rename_seqs(lambda x: x.upper())
+    new_again = new.rename_seqs(lambda x: x[:1])
+    expect = {"S"}
```
GavinHuttley, line 1970: this makes me realise the rename_seqs function should check that the resulting names are unique, otherwise a user goes from 3 seqs down to 1.

please add a test that fails with ValueError if the number of names in the resulting collection is not equal to the number in the original (see my suggestion above).

for this test, change the second lambda to
```python
new_again = new.rename_seqs(lambda x: f"{x[0]}{x[-1]}")
```

src/cogent3/core/new_alignment.py
```diff
@@ -719,30 +687,29 @@ def add_seqs(
...
         )
 
     def rename_seqs(self, renamer: Callable[[str], str]):
         """Returns new collection with renamed sequences."""
-        result = self.__class__(
-            seqs_data=self._seqs_data.rename_seqs(renamer),
+        new_name_map = {
+            renamer(name): old_name for name, old_name in self._name_map.items()
+        }
+        return self.__class__(
```
GavinHuttley, line 705: add the following condition
```python
if len(new_name_map) != len(self._name_map):
    raise ValueError(f"non-unique names produced by {renamer=}")
```

GavinHuttley (changes requested):
good PR! I did notice that we need to make sure a renamer generates unique names.

GavinHuttley (approved):
Bonza! Bonza!

PR #2065, opened 2024-10-26
ENH: ensure all nodes have names when using `make_tree`, and convert tip names to strings

YapengLang:
#2045

## Summary by Sourcery

Enhance the make_tree function to ensure all nodes have names and convert tip names to strings. Add tests to verify these enhancements.

Enhancements:
- Ensure all nodes have names when using the make_tree function.

Tests:
- Add test to verify that tip names are converted to strings when creating a tree from integer tip names.
- Add test to ensure all node names are uniquely assigned and no node name is an empty string.

### DIFF
--- a/src/cogent3/__init__.py
+++ b/src/cogent3/__init__.py
@@ -31,10 +31,7 @@ from cogent3.core.moltype import (
     get_moltype,
 )
 from cogent3.core.tree import PhyloNode, TreeBuilder, TreeError, TreeNode
-from cogent3.evolve.fast_distance import (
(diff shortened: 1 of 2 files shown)


GavinHuttley (changes requested):
If we have a tree with 1 million tips, naming all internal nodes could be expensive! It's better to provide an optional argument like `name_nodes: bool` which controls whether that method is called. To match the current behaviour, that should be `False` by default. Add tests that check both False and True cases explicitly.

PR #2079, opened 2024-10-29
ENH: add visualisation capabilities to new Alignment class

KatherineCaley:
includes other methods (filtered, trim_stop_codons,  omit_gap_pos, etc) required by the drawing methods

## Summary by Sourcery

Enhance the new Alignment class with drawing capabilities and additional sequence handling methods. Implement tests to verify the new functionalities, including drawing, filtering, and trimming sequences.

New Features:
- Introduce drawing capabilities to the new Alignment class, allowing for the creation of visual representations such as sequence logos, coevolution plots, and information plots.

Enhancements:
- Add methods to the Alignment class for filtering and trimming sequences based on specific criteria, such as gap fractions and stop codons.
- Implement new methods for handling sequence positions, including take_positions and take_positions_if, to facilitate selective sequence analysis.

Tests:
- Add new tests for the drawing capabilities of the Alignment class, including tests for dotplot functionality and coevolution drawables.
- Introduce tests for the new filtering and trimming methods, ensuring correct behavior when handling gaps and stop codons.

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -3498,7 +3498,6 @@ class Alignment(SequenceCollection):
             else new_sequence.SliceRecord(parent_len=self._seqs_data.align_len)
         )
         self._array_seqs = None
-        self._array_positions = None
 
     def _post_init(self):
         self._seqs = _IndexableSeqs(self, make_seq=self._make_aligned)
@@ -3580,28 +3579,21 @@ class Alignment(SequenceCollection):
         """Returns a numpy array of sequences, axis 0 is seqs in order
         corresponding to names"""
         if self._array_seqs is None:
-            start, stop, step = (
-                self._slice_record.start,
-                self._slice_record.stop,
-                self._slice_record.step,
+            # create the dest array dim
+            arr_seqs = numpy.empty(
+                (self.num_seqs, len(self)), dtype=self._seqs_data.alphabet.dtype
             )
-            seq_arrays = [
-                self._seqs_data.get_gapped_seq_array(
-                    seqid=name, start=start, stop=stop, step=step
-                )
-                for name in self._name_map.values()
-            ]
-            self._array_seqs = numpy.stack(seq_arrays, dtype=numpy.uint8)
+            for i, seq in enumerate(self.seqs):
+                arr_seqs[i, :] = numpy.array(seq)
+            arr_seqs.flags.writeable = False  # make sure data is immutable
+            self._array_seqs = arr_seqs
         return self._array_seqs
 
     @property
     def array_positions(self) -> numpy.ndarray:
         """Returns a numpy array of positions, axis 0 is alignment positions
         columns in order corresponding to names."""
-        if self._array_positions is None:
-            array_seqs = self.array_seqs
-            self._array_positions = array_seqs.T
-        return self._array_positions
+        return self.array_seqs.T
 
     def get_seq(
         self, seqname: str, copy_annotations: bool = False
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -4186,7 +4186,6 @@ class Alignment(SequenceCollection):
 
         return self.filtered(gaps_ok, motif_length=motif_length)
 
-
     def trim_stop_codons(self, gc: Any = None, strict: bool = False, **kwargs):
         # refactor: array
         if not self.has_terminal_stop(gc=gc, strict=strict):
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -3427,7 +3427,7 @@ class Alignment(SequenceCollection):
             moltype=self.moltype,
             info=self.info,
         )
-        if abs(slice.step) > 1:
+        if abs(new_slice.step or 1) > 1:
             return new
 
         # for simple slices, we retain the annotation database
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -904,6 +904,10 @@ class SequenceCollection:
         -----
         The returned collection will not retain an annotation_db if present.
         """
+        # refactor: design
+        # todo: this needs to be implemented for Alignment - need to think about the design.
+        # in the old implementation, this method removed all gaps and then returned a
+        # SequenceCollection (even for alignemnts).
         seqs = {
             name: self.moltype.degap(self._seqs_data.get_seq_array(seqid=name))
             for name in self._name_map.values()
@@ -4293,7 +4297,7 @@ class Alignment(SequenceCollection):
             result.annotation_db = self.annotation_db
 
         return result
-    
+
     def _get_seq_features(
         self,
         *,
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -60,6 +60,10 @@ StrORArray = Union[str, numpy.ndarray]
 StrORBytesORArray = Union[str, bytes, numpy.ndarray]
 MolTypes = Union[str, new_moltype.MolType]
 
+# small number: 1-EPS is almost 1, and is used for things like the
+# default number of gaps to allow in a column.
+EPS = 1e-6
+
 
 def assign_sequential_names(num_seqs: int, base_name: str = "seq", start_at: int = 0):
     """Returns list of sequential, unique names, e.g., ['seq_0' ... 'seq_n'] for
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -704,6 +704,10 @@ class SequenceCollection:
         # want to modify the underlying data, instead we create a new collection
         # with a subset of the names, recorded in the name_map dict.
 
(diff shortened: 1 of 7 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -4111,6 +4358,361 @@ def get_features(
...
+            whether to include gap counts, shown on right y-axis
+        """
+        from cogent3.draw.drawable import AnnotatedDrawable, Drawable
+
+        window = window or numpy.sqrt(len(self))
+        window = int(window)
+        y = self.entropy_per_pos()
+        nan_indices = numpy.isnan(y)
+        max_entropy = y[nan_indices == False].max()
+        y = max_entropy - y  # convert to information
```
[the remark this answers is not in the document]
GavinHuttley: that's a valid comment. Line 4629 will fail with a ValueError from numpy.max() on an empty matrix. I think
```python
if nan_indices.sum() == y.shape[0]: # assuming 1D array
    y = 0.0 # or numpy.nan, depends on context
```

src/cogent3/core/new_alignment.py
```diff
@@ -77,6 +81,95 @@ def assign_sequential_names(num_seqs: int, base_name: str = "seq", start_at: int
     return [f"{base_name}_{i}" for i in range(start_at, start_at + num_seqs)]
 
 
+class GapsOk:
```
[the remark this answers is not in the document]
GavinHuttley: We will ignore this because we intend to rewrite the two approaches to using numpy array operations and possibly converting to `numba.jit` functions. @KatherineCaley add a comment about redesign to this class

GavinHuttley (changes requested):
for the visualisation cases, can you trial the docs gallery code and see if it reproduces the published versions.

src/cogent3/core/new_alignment.py
```diff
@@ -3974,6 +4106,122 @@ def sliding_windows(
...
+            exclude columns with any gaps, 1 to include all columns.
+        motif_length
+            set's the "column" width, e.g. setting to 3 corresponds to codons.
+            A motif that includes a gap at any position is included in the
+            counting.
+        """
+
+        gaps = list(self.moltype.gaps)
+
+        gaps_ok = GapsOk(
```
GavinHuttley, line 4188: this GapsOk class may not be required for positions. This algorithm will be much faster with the proposed `iter_array_pos()` method. Add comment # redesign: use array methods

src/cogent3/core/new_alignment.py
```diff
@@ -4111,6 +4359,361 @@ def get_features(
...
+        # we order by tracks
+        top = 0
+        space = 0.25
+        annotes = []
+        for feature_type in drawables:
+            new_bottom = top + space
+            for i, annott in enumerate(drawables[feature_type]):
+                annott.shift(y=new_bottom - annott.bottom)
+                if i > 0:
+                    # todo modify the api on annott, we should not be using
```
GavinHuttley, line 4419: can you put the design: prefix on this comment som we can find them

src/cogent3/core/new_alignment.py
```diff
@@ -4111,6 +4359,361 @@ def get_features(
...
+        par_kw
+            dict of values for configuring parallel execution.
+
+        Returns
+        -------
+        DictArray of results with lower-triangular values. Upper triangular
+        elements and estimates that could not be computed for numerical reasons
+        are set as nan
+        """
+        from cogent3.draw.drawable import AnnotatedDrawable, Drawable
```
GavinHuttley, line 4519: add a comment here about redesign. These graphical representations of matrices should be separate functions in the drawing submodule somewhere

src/cogent3/core/new_alignment.py
```diff
@@ -4111,6 +4359,361 @@ def get_features(
...
+
+        window = window or numpy.sqrt(len(self))
+        window = int(window)
+        y = self.entropy_per_pos()
+        nan_indices = numpy.isnan(y)
+        max_entropy = y[nan_indices == False].max()
+        y = max_entropy - y  # convert to information
+        # now make all nan's 0
+        y[nan_indices] = 0
```
GavinHuttley, line 4633: add a comment on redesign about moving these drawing code into drawing submodule

GavinHuttley (changes requested):
we need to add some abstract methods `iter_alignment_pos()` to the underlying `AlignedSeqsDataABC` which returns a numpy array.

Mostly comments about adding comments ... how meta!

src/cogent3/core/new_alignment.py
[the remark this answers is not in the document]
GavinHuttley: my bad, change `y=0.0` to `y.fill(0.0)`

src/cogent3/core/new_alignment.py
```diff
@@ -3974,6 +4106,122 @@ def sliding_windows(
...
+                f"aligned length not divisible by motif_length={motif_length}"
+            )
+        gv = []
+        kept = False
+        seqs = [
+            self.get_gapped_seq(n).get_in_motif_size(motif_length, **kwargs)
+            for n in self.names
+        ]
+
+        positions = list(zip(*seqs))
```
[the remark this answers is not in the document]
KatherineCaley: I've added an `array_positions` property which matches the API from `ArrayAlignment`

src/cogent3/core/new_alignment.py
```diff
@@ -3012,6 +3124,37 @@ def get_gapped_seq_bytes(
...
+        array_seqs = self.get_array_seqs(names)
+
+        # if motif_length is 1, return the transposed array
+        if motif_length == 1:
+            return array_seqs.T
+
+        # For motif_length > 1, create an array with the specified motif length
+        num_positions = array_seqs.shape[1]
+        num_motifs = num_positions // motif_length
+        motif_array = numpy.zeros(
```
GavinHuttley, line 3149: add comment about whether we can use numpy.reshape for this

src/cogent3/core/new_alignment.py
```diff
@@ -3563,28 +3735,51 @@ def get_position_indices(
...
-                return not f(x)
+        return [i for i, col in enumerate(self.positions) if new_f(col)]
 
-        else:
-            new_f = f
+    def take_positions(self, cols: list, negate: bool = False):
+        """Returns new Alignment containing only specified positions.
 
-        result = [i for i, col in enumerate(self.positions) if new_f(col)]
+        Parameters
```
GavinHuttley, line 3755: design comment about using array_pos

src/cogent3/core/new_alignment.py
```diff
@@ -3449,8 +3601,14 @@ def __repr__(self):
         return f"{len(self.names)} x {len(self)} {self.moltype.label} alignment: {seqs}"
 
     def __len__(self):
+        # refactor: design
+        # should the len of a sliced alignment be updated? i.e.,
+        # return len(self._slice_record)
```
GavinHuttley, line 3606: this is the correct line, any code that fails as a result of this change is wrong

GavinHuttley (changes requested):
couple of small tweaks / comments

src/cogent3/core/new_alignment.py
```diff
@@ -2681,6 +2782,17 @@ def get_gapped_seq_bytes(
         step: OptInt = None,
     ) -> bytes: ...
 
+    @abstractmethod
+    def get_array_seqs(self, names: list[str]) -> numpy.ndarray:
```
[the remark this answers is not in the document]
KatherineCaley: @GavinHuttley I don't remember this exact discussion? delete the abstract methods but keep the implementation on AlignedSeqsData?

src/cogent3/core/new_alignment.py
KatherineCaley: will push through in the next PR

src/cogent3/core/new_alignment.py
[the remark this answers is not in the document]
GavinHuttley: delete both the abstract methods and the method on AlignedSeqsData. Because a slice and strand information are only available to the Alignment and Aligned instances, those classes have to be the place for building `array_seqs` and `array_pos`.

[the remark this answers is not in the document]
KatherineCaley: gotcha. done

src/cogent3/core/new_alignment.py
```diff
@@ -3262,6 +3265,8 @@ def map(self) -> IndelMap:
                 numpy.array([], dtype=int),
             )
 
+        # refactor: design
+        # should the returned map be sliced with the slice_record?
```
GavinHuttley, line 3269: my reaction to this comment is yes. But we need to revisit where the responsibilities lie for doing slicing and make sure we're consistent across the API for all the classes. I would think that the AlignedDataView handles slicing of all the variants of a sequence (str, bytes, array by gapped, ungapped) and thus for the gapped arrays must also handle slicing of the indel map too. I suggest looking at every method that delivers an attribute relied on by Aligned. Each of those should employ the same strategy. If not, let's discuss!

GavinHuttley (approved):
I think time to revisit consistency in public attributes of AlignedDataView. See comments on comment.

PR #2091, opened 2024-11-04
ENH: add no_degenerates, sample, get_translation to new Alignment

KatherineCaley:
## Summary by Sourcery

Add new methods 'no_degenerates', 'sample', and 'get_translation' to the Alignment class, enhancing its functionality for sequence manipulation and analysis. Refactor existing tests to improve coverage and add new tests for the introduced methods.

New Features:
- Introduce the 'no_degenerates' method to remove degenerate characters from alignments, with options for motif length and gap allowance.
- Add the 'sample' method to allow random sampling of alignment positions, supporting both with and without replacement options.
- Implement the 'get_translation' method to translate nucleotide sequences into protein sequences, with options for handling stop codons.

Enhancements:
- Refactor test cases to use parameterized tests for better coverage and flexibility in testing different sequence collection methods.

Tests:
- Add new tests for the 'sample' method to verify functionality with and without replacement, and for different motif sizes.
- Include tests for the 'no_degenerates' method to ensure correct exclusion of columns with degenerate characters.

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -2499,7 +2499,7 @@ def _(
 
 @singledispatch
 def seq_to_gap_coords(
-    seq: StrORBytesORArray,
+    seq: typing.union[StrORBytesORArray, new_sequence.Sequence],
     *,
     alphabet: new_alphabet.AlphabetABC,
 ) -> tuple[numpy.ndarray, numpy.ndarray]:
@@ -4511,7 +4511,7 @@ class Alignment(SequenceCollection):
                 include_stop=include_stop,
                 trim_stop=trim_stop,
             )
-            translated[seqname] = numpy.array(pep)
+            translated[seqname] = pep
         kwargs["moltype"] = pep.moltype
         seqs_data = self._seqs_data.from_seqs(
             data=translated, alphabet=pep.moltype.most_degen_alphabet()
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -4426,6 +4426,13 @@ class Alignment(SequenceCollection):
         defaults generates a standard bootstrap resampling of the alignment.
         """
         # refactor: type hint for randint, permutation
+        # refactor: array
+        #  Given array_pos property, it will be efficient to generate random
+        # indices and use numpy.take() using that. In the case of motif_length
+        # != 1, the total number of positions is just len(self) // motif_length.
+        # Having produced that, those indices can be scaled back up, or the
+        # numpy array reshaped.
+
         population_size = len(self) // motif_length
         if not n:
             n = population_size
@@ -4484,6 +4491,7 @@ class Alignment(SequenceCollection):
 
         return result
 
+    @extend_docstring_from(SequenceCollection.get_translation)
     def get_translation(
         self,
         gc: int = None,
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -4320,6 +4320,38 @@ class Alignment(SequenceCollection):
 
         return self.gapped_by_map(keep, info=self.info)
 
+    def no_degenerates(self, motif_length: int = 1, allow_gap: bool = False):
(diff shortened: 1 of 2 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -4360,6 +4392,67 @@ def has_terminal_stop(self, gc: Any = None, strict: bool = False) -> bool:
...
+
+        Notes
+        -----
+        By default (resampling all positions without replacement), generates
+        a permutation of the positions of the alignment.
+
+        Setting with_replacement to True and otherwise leaving parameters as
+        defaults generates a standard bootstrap resampling of the alignment.
+        """
+        # refactor: type hint for randint, permutation
```
GavinHuttley, line 4428: add refactor array here too. Given `array_pos` property, it will be efficient to generate random indices and use `numpy.take()` using that. In the case of motif_length != 1, the total number of positions is just `len(self) // motif_length`. Having produced that, those indices can be scaled back up, or the numpy array reshaped...

src/cogent3/core/new_alignment.py
```diff
@@ -4320,6 +4320,38 @@ def filtered(
...
+        ----------
+        motif_length
+            sequences are segmented into units of this size and the segments are
+            excluded if they contain degenerate characters.
+        allow_gap
+            whether gaps are allowed or whether they are treated as a degenerate
+            character (latter is default, as most evolutionary modelling treats
+            gaps as N).
+        """
+        # refactor: array
```
GavinHuttley, line 4336: note, the alphabets are designed such that the canonical characters occur before the first gap, then the first ambiguity etc.. So, an array approach is

```python
# make sure valid moltype, then
cutoff = len(self.moltype)
indices = (self.array_pos < cutoff).all(axis=1)
# for when motif_length != 1, but already handled non-modulo sequence
num_motif = len(self) // motif_length
motif_valid = indices.reshape(num_motif, motif_length).any(axis=1).flatten()
selected = self.array_pos[selected]
```
or something like that

tests/test_core/test_new_alignment.py
```diff
@@ -3848,6 +3876,41 @@ def test_get_feature():
...
+            "seq2": "AACCDDEEFFGGHHIIKKLLMMNNPP",
+        },
+        moltype="protein",
+    )
+    # ensure length correct
+    sample = alignment.sample(n=10, motif_length=2)
+    assert len(sample), 20
+    # test columns alignment preserved
+    seqs = list(sample.to_dict().values())
+    assert seqs[0], seqs[1]
```
GavinHuttley, line 3908: should that be `seqs[0] == seqs[1]`?

tests/test_core/test_new_alignment.py
```diff
@@ -3848,6 +3876,41 @@ def test_get_feature():
...
+    # ensure each char occurs twice as sampling dinucs without replacement
+    for char in seqs[0]:
+        assert seqs[0].count(char) == 2
+
```
GavinHuttley, line 3912: add assertion that the adjacent characters are identical
```python
assert all(len(set(seq[0][i: i+2])) == 1 for i in range(0, len(s[0], 2)))
```

GavinHuttley (changes requested):
test suggestions and some more comments

tests/test_core/test_new_alignment.py
KatherineCaley: good catch! issue from copy pasta from unittesting

src/cogent3/core/new_alignment.py
KatherineCaley: @GavinHuttley this makes me think that we should have a from_array method on `AlignedSeqsData`, where we provide the array of sequences and a list of names defining the order of the sequences in the array. What do you think

```py

@classmethod
def from_names_and_array(names: list, data: numpy.ndarray) -> AlignedSeqsData:
    ...

```

GavinHuttley: agreed

GavinHuttley (approved):
bonza bonza!

PR #2097, opened 2024-11-05
BUG: moltype.to_degenerates now checks if moltype has degen alpha

KatherineCaley:
## Summary by Sourcery

Fix moltype.to_degenerates to ensure it checks for a degenerate alphabet before processing, and add a test for HTML generation from text moltype alignments.

Bug Fixes:
- Fix moltype.to_degenerates to check if moltype has a degenerate alphabet before proceeding.

Tests:
- Add a test for producing HTML from text moltype alignments.

### DIFF
--- a/src/cogent3/core/new_moltype.py
+++ b/src/cogent3/core/new_moltype.py
@@ -660,13 +660,9 @@ class MolType:
                 f"{seq[:4]!r} not valid for moltype {self.name!r}"
             )
 
-        # what index is the first degenerate character
-        for index, val in enumerate(self.most_degen_alphabet()):
-            if val in self.ambiguities:
-                break
-        else:
-            return False
-        return (seq >= index).any()
+        first_degen = self.most_degen_alphabet().gap_index + 1
+
+        return (seq >= first_degen).any()
 
     @functools.singledispatchmethod
     def is_gapped(self, seq, validate: bool = True) -> bool:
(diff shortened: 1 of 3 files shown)


src/cogent3/core/new_moltype.py
```diff
@@ -647,18 +647,21 @@ def _(self, seq: str, validate: bool = True) -> bool:
...
+        if self.degen_alphabet is None:
+            return False
+
         if validate and not self.is_valid(seq):
             raise new_alphabet.AlphabetError(
                 f"{seq[:4]!r} not valid for moltype {self.name!r}"
             )
 
-        for index, val in enumerate(self.degen_gapped_alphabet):
+        # what index is the first degenerate character
```
GavinHuttley, line 663:
```python
# since you've already excluded non-degen moltypes
first_degen = self.most_degen_alphabet().gap_index + 1
```

GavinHuttley (approved):
but I have answered a question you asked in a comment, so you can address that and then merge

src/cogent3/core/new_moltype.py
KatherineCaley: @GavinHuttley That's not a question. That is a comment (that already existed in the function) which I moved below my changes. 

The code below the comment finds the first degenerate character - which I can update to the snippet you provided above if you'd prefer

GavinHuttley: my approach more succinct, try that and let me know if it fails!

KatherineCaley: it works!

PR #2106, opened 2024-11-12
ENH: add feature handling to new style alignments

KatherineCaley:


src/cogent3/core/alignment.py
```diff
@@ -4952,7 +4960,7 @@ def add_from_ref_aln(self, ref_aln, before_name=None, after_name=None):
 
         return self.add_seqs(temp_aln, before_name, after_name)
 
-    def replace_seqs(self, seqs, aa_to_codon=True):
+    def replace_seqs(self, seqs, aa_to_codon=True):  # will not port
```
KatherineCaley, line 4963: will indeed port but with a name change to `apply_scaled_gaps`

src/cogent3/core/new_moltype.py
```diff
@@ -716,9 +716,12 @@ def get_degenerate_positions(
                 f"{seq[:4]!r} not valid for moltype {self.name!r}"
             )
 
+        if self.degen_alphabet is None:
+            return []
+
         alpha = self.most_degen_alphabet()
         seq = alpha.to_indices(seq)
-        for index, val in enumerate(self.degen_gapped_alphabet):
+        for index, val in enumerate(alpha):
```
GavinHuttley, line 724: Same issue as we've discussed before. You can assume that the ambiguity indices occur after the `len(moltype)+1` -- plus one for the gap index.

GavinHuttley (changes requested):
massive effort!!! 🚀💥💪

Some suggestions, comments etc..

GavinHuttley (approved):
bonza bonza!

PR #2107, opened 2024-11-13
Performance optimisations

GavinHuttley:
## Summary by Sourcery

Optimize sequence data handling for improved performance and add tests for data type conversions.

Enhancements:
- Improve performance by optimizing the handling of sequence data types in the SeqView class, allowing for more efficient conversion between string, bytes, and array representations.

Tests:
- Add tests to verify the correct conversion of sequence data types to and from indices, ensuring compatibility with different input formats such as strings, bytes, and numpy arrays.

### DIFF
--- a/src/cogent3/core/new_alphabet.py
+++ b/src/cogent3/core/new_alphabet.py
@@ -107,7 +107,7 @@ class AlphabetABC(ABC):
     def with_gap_motif(self): ...
 
     @abstractmethod
-    def to_indices(self, seq: str) -> numpy.ndarray: ...
+    def to_indices(self, seq: StrORBytesORArray) -> numpy.ndarray: ...
 
     @abstractmethod
     def from_indices(self, seq: numpy.ndarray) -> str: ...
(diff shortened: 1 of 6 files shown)


src/cogent3/core/new_sequence.py
```diff
@@ -2640,15 +2640,24 @@ def _get_init_kwargs(self):
             "slice_record": self.slice_record,
         }
 
+    # DESIGN NOTE
+    # alphabet.to_indices() and alphabet.from_indices() cope with most primitive
+    # types that parent may be
     @property
```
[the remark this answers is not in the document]
GavinHuttley: extra method call adds unnecessary overhead

src/cogent3/core/new_alphabet.py
```diff
@@ -771,17 +779,17 @@ def to_index(self, seq) -> numpy.ndarray:
         raise TypeError(f"{type(seq)} not supported")
 
     @to_index.register
-    def _(self, seq: str) -> numpy.ndarray:
+    def _(self, seq: str) -> int:
```
KatherineCaley, line 782: is seq is a single character? if yes should we change the arg name to char

src/cogent3/core/new_sequence.py
```diff
@@ -2597,7 +2597,7 @@ class SeqView(SeqViewABC):
     def __init__(
         self,
         *,
-        parent: typing.Union[str, "SeqsDataABC"],
+        parent: StrORBytesORArray,
```
KatherineCaley, line 2600: note that `SeqDataView` inherits from `SeqView`, and parent is `SeqsDataABC` in that case.

src/cogent3/core/new_alphabet.py
GavinHuttley: On CharAlphabet yes, but on KmerAlphabet and CodonAlphabet it will be > 1 character.

src/cogent3/core/new_sequence.py
GavinHuttley: I don't think a base class should declare as valid the types for all potential subclasses. Better to redefine `__init__` in the subclass to make it explicit.

PR #2108, opened 2024-11-13
More performance optimisations

GavinHuttley:
## Summary by Sourcery

Enhancements:
- Optimize performance by applying Numba JIT compilation to several functions, including _gap_spans, seq_to_align_index, align_to_seq_index, decompose_gapped_seq, and compose_gapped_seq.

### DIFF
--- a/src/cogent3/core/location.py
+++ b/src/cogent3/core/location.py
@@ -8,6 +8,7 @@ from bisect import bisect_left, bisect_right
 from functools import total_ordering
 from typing import Any, Iterator, Optional, Sequence, Union
 
+import numba
 import numpy
 from numpy.typing import NDArray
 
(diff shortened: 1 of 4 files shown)


src/cogent3/core/location.py
```diff
@@ -1077,6 +1079,101 @@ def _input_vals_neg_step(
     return (0, 0, 1) if start < stop else (start, stop, step)
 
 
+@numba.jit
+def seq_to_align_index(
+    gap_pos: IntArrayTypes,
+    cum_lengths: IntArrayTypes,
+    parent_length: int,
```
KatherineCaley, line 1086: just a comment that in general we should document that the parent is the seq, sometimes I get confused which length parent_length refers to (alignment or seq)

src/cogent3/core/location.py
```diff
@@ -1077,6 +1079,101 @@ def _input_vals_neg_step(
...
+    if seq_index < 0:
+        seq_index += parent_length
+
+    if seq_index < 0:
+        raise IndexError(f"{seq_index} negative seq_index beyond limit ")
+
+    if not num_gaps or seq_index < gap_pos[0]:
+        return int(seq_index)
+
+    # if stop_index, check if the seq_index corresponds to a gap position
```
KatherineCaley, line 1113: very minor, but what is the `stop_index` referred to in this comment?

KatherineCaley (approved):

src/cogent3/core/location.py
GavinHuttley: it should be slice_stop, good catch!

PR #2111, opened 2024-11-14
API: add SliceRecordABC.plus_start and plus_stop

GavinHuttley:
[NEW] these return start and stop in plus strand coordinates,
     irrespective of whether reversed or not. This is most critical
     in SeqDataView where previous code was passing just parent_start
     and parent_stop. The latter use the offset, which should be
     employed only for feature querying. Adding comments to that effect
     to the docstrings.

## Summary by Sourcery

Add plus_start and plus_stop properties to SliceRecordABC to provide consistent plus strand coordinates, refactor SeqDataView to utilize these properties, and update documentation and tests accordingly.

New Features:
- Introduce plus_start and plus_stop properties in SliceRecordABC to return start and stop in plus strand coordinates, regardless of strand orientation.

Enhancements:
- Refactor SeqDataView to use plus_start and plus_stop for sequence retrieval, ensuring consistent handling of strand orientation.

Documentation:
- Add notes to docstrings clarifying the intended use of parent_start and parent_stop properties, emphasizing they should not be used for slicing on the parent object.

Tests:
- Add tests for the new plus_start, plus_stop, and plus_step properties to ensure they return correct values for both forward and reverse strand orientations.

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -180,6 +180,11 @@ class SeqDataView(new_sequence.SeqView):
     @property
     def str_value(self) -> str:
         """returns the sequence as a string"""
+        return self.alphabet.from_indices(self.array_value)
+
+    @property
+    def array_value(self) -> numpy.ndarray:
+        """returns the sequence as a numpy array"""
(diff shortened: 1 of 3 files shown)


KatherineCaley (approved):
I think this is a necessary change

PR #2112, opened 2024-11-15
TST: porting testing to new alignment module

KatherineCaley:
## Summary by Sourcery

Port tests to the new alignment module and introduce additional tests for alignment functionalities, including casting to DNA/RNA and variable position identification.

Tests:
- Port tests to the new alignment module, including tests for sequence collection initialization, counts per sequence, and alignment slicing.
- Add new tests for alignment casting to DNA and RNA, ensuring info attributes are preserved.
- Introduce tests for variable positions and gap filtering in alignments.

### DIFF
--- a/src/cogent3/core/new_alignment.py
+++ b/src/cogent3/core/new_alignment.py
@@ -4190,7 +4190,13 @@ class Alignment(SequenceCollection):
 
         Parameters
         ----------
-
+        motif_length
+            number of elements per character.
+        include_ambiguity
+            if True, motifs containing ambiguous characters from the seq moltype
+            are included. No expansion of those is attempted.
+        allow_gap
+            if True, motifs containing a gap character are included.
         warn
             warns if motif_length > 1 and alignment trimmed to produce
             motif columns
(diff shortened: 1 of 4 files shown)


src/cogent3/core/new_alignment.py
```diff
@@ -4468,6 +4474,29 @@ def count_gaps_per_seq(
...
+        """Return a list of variable position indexes.
+
+        Parameters
+        ----------
+        include_gap_motif
+            if False, sequences with a gap motif in a
+            column are ignored.
+
+        """
+        indices = []
```
GavinHuttley, line 4487: implement the following as a separate function outside the class
```python

@numba.jit
def _var_pos(arr, below):   # pragma: no cover
    m, n = arr.shape
    result = np.zeros(n, dtype=np.bool_)
    
    for col in range(n):
        last = -1
        for row in range(m):
            value = arr[row, col]
            if value < below:
                if last == -1:
                    last = value
                elif value != last:
                    result[col] = True
                    break
    
    return result
```
that will return a bool list much faster. Then return the indices for where True occurs.

GavinHuttley (changes requested):
for computing values in columns, start using the numpy functions. For computing counts, for example, use `numpy.unique(arr, return_counts=True)`. More difficult for motif_length>1, there add a comment about writing a numba decorated function (I have a draft of that)

KatherineCaley:
> for computing values in columns, start using the numpy functions. 

@GavinHuttley  I found this was slower than using `set()` because it orders the resulting array

PR #2130, opened 2024-11-22
DEV: views and seqs_data do not support serialisation

KatherineCaley:
## Summary by Sourcery

Remove support for SeqView serialization and simplify the serialization process by directly using sequence strings. Update tests to reflect the removal of SeqView serialization.

Enhancements:
- Simplify the serialization process by removing the use of SeqView in the to_rich_dict method.

Tests:
- Remove tests related to SeqView serialization, as the feature is no longer supported.

### DIFF
--- a/tests/test_core/test_new_sequence.py
+++ b/tests/test_core/test_new_sequence.py
@@ -2103,7 +2103,6 @@ def test_seqview_copy(sliced, rev, integer_seq):
     copied = sv.copy(sliced=sliced)
 
     assert copied.str_value == raw_data[slice_start:slice_end]
-    assert copied.is_reversed == integer_seq.is_reversed
     assert (
         sliced and copied.parent is not sv.parent or copied.parent is integer_seq.parent
     )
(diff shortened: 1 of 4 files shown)


tests/test_core/test_new_sequence.py
```diff
@@ -2513,6 +2470,8 @@ def test_copied_parent_coordinates(sliced, rev, start_stop):
     # matches original
     assert copied.parent_coordinates() == seq.parent_coordinates()
     # and expected -- the coordinate name always reflects the underlying sequence
+    # the strand will always be 1 for a copied sequence because the reversal is
```
GavinHuttley, line 2473: I don't understand this comment.

GavinHuttley (changes requested):
just one question!

tests/test_core/test_new_sequence.py
KatherineCaley: half baked thought that slipped through

PR #2131, opened 2024-11-24
DEV: process token in tree node classes to handle incorrect identification of support or name when creating a node

### LINKED ISSUE #2116: BUG: newick trees with the same support values are incorrectly assigned suffixes
**Describe the bug**
When a supported newick tree provided, the parser will incorrectly add suffix to nodes with the same support value.

**To Reproduce**
```
>>> import cogent3

>>> tree = cogent3.make_tree("((a,b)100, (c,d)100);", name_nodes=True)
>>> print(tree.to_rich_dict())

```
it gives:

"{'newick': '((a,b)node1,(c,d)node2)', 'edge_attributes': {'a': {'length': None}, 'b': {'length': None}, 
'node1': {'length': None, 'support': **_100.0_**}, 'c': {'length': None}, 'd': {'length': None}, 
'node2': {'length': None, 'support': **_100.2_**}, 
'root': {'length': None}}, 'type': 'cogent3.core.tree.PhyloNode', 'version': '2024.7.19a9'}"

YapengLang:
In this PR:
- Added the static method `parse_token` to both TreeNode and PhyloNode.
- Applied `parse_token` in `TreeBuilder.create_edge()` to distinguish between name and support before creating a node.
- Modified `test_phylonode_support` to cover cases where the same support value occurs.

- Deprecated `name_node` parameter in `make_tree` and its tests, as nodes with only support are now always assigned names.

closes #2116

## Summary by Sourcery

Add a static method `parse_token` to TreeNode and PhyloNode classes to improve token parsing and fix node creation issues. Apply this method in `TreeBuilder.create_edge()` to correctly distinguish between name and support. Update tests to cover new scenarios and deprecate the `name_node` parameter in `make_tree`.

New Features:
- Introduce a static method `parse_token` in both TreeNode and PhyloNode classes to handle token parsing.

Bug Fixes:
- Fix incorrect identification of support or name when creating a node by applying `parse_token` in `TreeBuilder.create_edge()`.

Enhancements:
- Modify `test_phylonode_support` to cover cases where the same support value occurs.

Chores:
- Deprecate the `name_node` parameter in `make_tree` as nodes with only support are now always assigned names.


GavinHuttley (approved):
add deprecated_arg call to `name_nodes` argument as discussed

src/cogent3/__init__.py
```diff
@@ -735,11 +732,15 @@ def load_table(
...
 
+@c3warn.deprecated_args(
+    version="2024.12", reason="argument has no effect", discontinued="name_nodes"
+)
 def make_tree(
     treestring=None,
     tip_names=None,
     format=None,
     underscore_unmunge=False,
+    name_nodes=False,
```
GavinHuttley: delete the argument and remove it from the docstring, make sure it's not used in any tests too

GavinHuttley (changes requested):
the only thing you need is to add the decorator

YapengLang:
Hi @GavinHuttley , finished. please take a look.

PR #2176, opened 2024-12-17
STY: apply ruff --unsafe-fixes across code base

GavinHuttley:


GavinHuttley:
ignoring the codeql as those tests are slated for deletion

PR #2224, opened 2025-03-02
DOC: added docstring for new_moltype.MolType.get_degenerate_positions, fixes #2218

### LINKED ISSUE #2218: DOC: add docstring for new_moltype.MolType.get_degenerate_positions
check the tests of this method to see what the arguments mean.

Be sure and use the correct docstring style - we adhere to numpydoc style, but we do not type hint parameters.

The easiest way to do this is to install `autoDocstring - Python Docstring Generator` extension on VS Code and specify numpydoc as the style.

je634:
Using autoDocstring - Python Docstring Generator extension to generate Docstring for the definition of **get_degenerate_positions** in new_moltype.py.


GavinHuttley:
@je634 I realise this has been closed, but am commenting here to provide tips about how to address issues such as this one.
When an issue nominates a single method / function, only introduce changes to that method or function. Automated tools that modify more than that will cause merge conflicts with others and, more critically, there output is unlikely to be useful.
Read the issue, look at what the method is meant to be doing and address that only.

PR #2229, opened 2025-03-05
DOC: added docstring for new_moltype.MolType.get_degenerate_positions, fixes #2218

### LINKED ISSUE #2218: DOC: add docstring for new_moltype.MolType.get_degenerate_positions

je634:
## Summary by Sourcery

Documentation:
- Adds a docstring to the `get_degenerate_positions` method, including parameter descriptions, return type, and potential exceptions.

### DIFF
--- a/src/cogent3/core/new_moltype.py
+++ b/src/cogent3/core/new_moltype.py
@@ -858,26 +858,16 @@ class MolType:
         include_gap: bool = True,
         validate: bool = True,
     ) -> list[int]: 
-        """_summary_
-
+        """Return List of position indexs of degenerate characters in the sequence.
+    
         Parameters
         ----------
         seq : StrORBytesORArray
-            _description_
+            the sequence to be used for getting degenerate positions
         include_gap : bool, optional
-            _description_, by default True
+            if False, add 1 to the index
         validate : bool, optional
-            _description_, by default True
-
-        Returns
-        -------
-        list[int]
-            _description_
-
-        Raises
-        ------
-        new_alphabet.AlphabetError
-            _description_
+            if True, checks the sequence is validated
         """  
         if validate and not self.is_valid(seq):
             msg = f"{seq[:4]!r} not valid for moltype {self.name!r}"
--- a/src/cogent3/core/new_moltype.py
+++ b/src/cogent3/core/new_moltype.py
@@ -857,18 +857,18 @@ class MolType:
         seq: StrORBytesORArray,
         include_gap: bool = True,
         validate: bool = True,
-    ) -> list[int]: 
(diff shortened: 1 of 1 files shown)


GavinHuttley (changes requested):
Hi @je634 ,
Thanks for the effort. Unfortunately, this PR only includes an automated docstring output. Do you think this would help you as a user to understand what this method does?
Take a look at how this method is tested to see how it's being used and amend the docstring to reflect that.
I suggest you look at other examples in the same file, e.g. [is_ambiguity()](https://github.com/cogent3/cogent3/blob/7f98a2acf549d7ae1b0f743c41103783f3a45582/src/cogent3/core/new_moltype.py#L925).

je634:
Hi @GavinHuttley, 
I did some modifications on the docstring and have updated the PR. I am still unsure about what validate and included_gap does. I can tell that if the included_gap is true we do not change the value of the index, otherwise we will add 1 to the index. And for the validate bool variable, we will do some validation and printout a msg. Do you recommend me to go through other files to understand what exactly do they do?

GavinHuttley:
Hi @je634,
`validate` is used to check the sequence is valid for the alphabet, this is the first statement inside the method.
`include_gap` triggers a change in the minimal value not considered ambiguous. That is, if include_gap is False, then any state that is not "canonical" (A,C,G,T for DNA) is considered ambiguous which will include the gap state too.

je634:
Hi @GavinHuttley,
Sorry I just finished the last class and I noticed there is still a requested change. Could I get some information on what changes do I need to make? I clicked on "Codacy Static Code Analysis" under "1 pending check" and entered the page, which shows "2 new issues (0 max.) of at least minor severity." but nothing else. Could I get some instructions? Thank you!

GavinHuttley:
hi @je634 ,
You can ignore the codacy issue. Just focus on my comment regarding what the two arguments you enquired about do.

Once you have made your changes, you should also run the linter on your code as follows
```
$ cd to/repo/root
$ ruff check --fix-only . && ruff format .
```
then run the test suite
```
$ cd tests
$ pytest
```
then  committing your final changes and push

je634:
Hi @GavinHuttley, 
I believe I have corrected the appropriate docstring and run the linter yesterday. Could you please confirm if everything is in order?

GavinHuttley (approved):
Nice one @je634 !

PR #2231, opened 2025-03-05
ENH: improve performance of SequenceCollection.get_ambiguous_position…

### LINKED ISSUE #2214: ENH: improve performance of SequenceCollection.get_ambiguous_positions
This method should be changed to using the ability to cast each sequence to a numpy.array and that ambiguity states will have an index (integer) value > that that for the gap_index.

Specifically, moltype.most_degen_alphabet().gap_index and `numpy.array(self.seqs[name])`.

Make sure tests cover cases where a moltype that does not have degenerate characters is being tested too.

c-jonesy:
Fixes #2214.
Changed to using the ability to cast each sequence to a numpy.array

## Summary by Sourcery

Enhancements:
- Improves performance of SequenceCollection.get_ambiguous_positions by casting sequences to numpy arrays.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ Changelog = "https://github.com/cogent3/cogent3/blob/develop/changelog.md"
 
 [project.optional-dependencies]
 test = [
-    "ruff==0.9.7",
+    "ruff==0.9.9",
     "kaleido",
     "pandas",
     "plotly",


src/cogent3/core/alignment.py
GavinHuttley, line 1: revert the changes to the `alignment.py` file, it is being discontinued. Keep the changes to `new_alignment.py`

GavinHuttley (changes requested):
Whoops! Changed too many files. You should always run the test suite on your computer to check your change has not broken things. If you remove the code I commented on in this PR, I'll bet the test suite will pass on your computer.

So close!!

c-jonesy:
accidentally closed it oops

PR #2233, opened 2025-03-06
ENH: improve performance of SequenceCollection.get_ambiguous_positions

### LINKED ISSUE #2214: ENH: improve performance of SequenceCollection.get_ambiguous_positions

c-jonesy:
Fixes #2214
Changed to using the ability to cast each sequence to a numpy.array

## Summary by Sourcery

Enhancements:
- Improves performance of SequenceCollection.get_ambiguous_positions.

### DIFF
--- a/src/cogent3/core/alignment.py
+++ b/src/cogent3/core/alignment.py
@@ -1130,12 +1130,11 @@ class _SequenceCollectionBase:
         Used in likelihood calculations.
         """
         result = {}
-        alpha = self.moltype.most_degen_alphabet()
         for name in self.names:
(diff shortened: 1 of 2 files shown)


GavinHuttley (approved):
💥 awesome job @c-jonesy , thanks!

PR #2236, opened 2025-03-06
Deprecated tprob() and replaced with scipy.stats.t.sf()

Yezi-yangyang:
## Summary by Sourcery

Deprecates the tprob function, replacing it with a new tprob_new function that uses scipy.stats.t.sf() for improved statistical calculations.

Enhancements:
- Replaces the tprob function with tprob_new, utilising scipy.stats.t.sf() for enhanced statistical calculations.
- Adds a deprecated_callable decorator to the old tprob function, which will emit a warning when the old function is called, and directs users to the new function.
- Updates tests to use the new tprob_new function instead of the deprecated tprob function.
- Updates t_tailed_prob to use tprob_new instead of tprob.

### DIFF
--- a/src/cogent3/maths/stats/distribution.py
+++ b/src/cogent3/maths/stats/distribution.py
@@ -31,28 +31,20 @@ incbet = betai  # shouldn't have renamed it...
 
 # Probability integrals: low gives left-hand tail, high gives right-hand tail.
 
+import scipy.stats
 
+from cogent3.util import warning as c3warn
 def zprob(x):
     """Returns both tails of z distribution (-inf to -x, inf to x)."""
     return 2 * norm.sf(abs(x))
 
 
-import scipy.stats
-
-from cogent3.util import warning as c3warn
-
-
-def tprob_new(x, df):
-    """Returns both tails of t distribution (-infinity to -x, infinity to x)"""
-    return 2 * scipy.stats.t.sf(abs(x), df)
-
-
 @c3warn.deprecated_callable(
-    version="2025.3", reason="Use scipy.stats.t.sf() instead", new="tprob_new"
+    version="2025.6", reason="Use scipy.stats.t.sf() instead", is_discontinued=True
 )
-def tprob(x, df):  # pragma: no cover
+def tprob(x, df):  
     """deprecated, use new"""
-    return tprob_new(x, df)
+    return 2 * scipy.stats.t.sf(abs(x), df)
 
 
 def binomial_exact(successes, trials, prob):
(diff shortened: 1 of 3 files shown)


src/cogent3/maths/stats/distribution.py
```diff
@@ -37,9 +37,22 @@ def zprob(x):
     return 2 * norm.sf(abs(x))
 
 
-def tprob(x, df):
+import scipy.stats
```
GavinHuttley, line 40: imports needed to be at the top of the file. Using the linter `ruff` on the code will fix this.

src/cogent3/maths/stats/distribution.py
```diff
@@ -37,9 +37,22 @@ def zprob(x):
...
+
+from cogent3.util import warning as c3warn
+
+
+def tprob_new(x, df):
```
GavinHuttley, line 45: this function should be deleted

src/cogent3/maths/stats/distribution.py
```diff
@@ -37,9 +37,22 @@ def zprob(x):
...
     """Returns both tails of t distribution (-infinity to -x, infinity to x)"""
-    return 2 * t.sf(abs(x), df)
+    return 2 * scipy.stats.t.sf(abs(x), df)
+
+
+@c3warn.deprecated_callable(
+    version="2025.3", reason="Use scipy.stats.t.sf() instead", new="tprob_new"
```
GavinHuttley, line 51: the version for removal is incorrect, it should be 2025.6.

the decorator function should also have `is_discontinued=True`

src/cogent3/maths/stats/test.py
GavinHuttley, line 1: delete import and convert all cogent3 uses of tprob to using `scipy.stats.t.sf` instead

tests/test_maths/test_stats/test_distribution.py
GavinHuttley, line 1: delete import and test

GavinHuttley (changes requested):
Hi @Yezi-yangyang , thanks for the PR. Unfortunately, this PR does not address the stated issue correctly.
When a function is being marked for deprecation, it is intended it is just being marked to be removed. We only identify the replacement function so that users of cogent3's tprob know what they need to do in order to replace the function in their code. All cogent3 code that was using that function should be converted to use the nominated scipy replacement code instead.

GavinHuttley (changes requested):
please run the linter commands on your version of the repo and commit the changes it makes.

GavinHuttley:
hi @Yezi-yangyang ,
unfortunately, this PR doesn't address the issue correctly and does not achieve our standards. Thanks for your attempt!

PR #2339, opened 2025-06-03
ENH: Check all immediate children for presence of support

rmcar17:
When drawing a dendrogram, if the first child is a tip, then this won't contain IQ-TREE's ultrafast bootstrap support. This leads to support not being drawn on many trees.

Separately, the default threshold for drawing support is 1.0 - support values above this will not be drawn unless it's set to be higher. IQ-TREE uses 100 as the maximum support - should the default threshold be increased? 

Alternative as just discussed, have a `support_as_percentage` or proportion boolean value

## Summary by Sourcery

Enhancements:
- Check all immediate children for support parameters before disabling support display instead of only inspecting the first child


GavinHuttley (approved):
could be simplified to
```python
show_support = any("support" in ...)
```

PR #2412, opened 2025-07-31
ENH: `PhyloNode` and related files are now typed. Use `__slots__` on `PhyloNode`. Remove support for `TreeNode`.

rmcar17:
Marked as draft as backwards compatibility needs to be added.

## Summary by Sourcery

Refactor tree implementation to use a fully typed PhyloNode class, remove legacy TreeNode support, and improve performance and correctness across tree operations and drawing utilities

New Features:
- Introduce __slots__ on PhyloNode for memory optimization
- Add comprehensive type annotations (Self, ParamSpec, precise return types) across PhyloNode and related modules
- Standardize on PhyloNode as the sole tree node class, dropping TreeNode

Enhancements:
- Refactor PhyloNode methods with typed parameters and overloads for list interfaces, enforcing parent/child link cleanup
- Streamline tree I/O by removing XML support and focusing on Newick/JSON serialization
- Clarify tree API parameter names (e.g. tip_name_1, tip_name_2) and refine drawing utilities with consistent type annotations
- Update configuration files (pyproject.toml, ruff.toml) for dependency formatting, linting rules, and staging

Tests:
- Update tests to use PhyloNode exclusively and accommodate new signatures and behaviors

## Summary by Sourcery

Migrate tree implementation to a fully typed PhyloNode class, remove legacy TreeNode support, and unify all tree operations and drawing utilities around PhyloNode. Introduce __slots__ and thorough type annotations, simplify serialization to Newick/JSON, revamp drawing geometry classes, update tests to match new interfaces, and refresh project configuration for dependency and linting formats.

New Features:
- Add __slots__ to PhyloNode for memory and performance optimization
- Introduce comprehensive type annotations (Self, ParamSpec, precise return types) in PhyloNode, drawing modules, and related code

Enhancements:
- Standardize on PhyloNode as the sole tree node class, dropping TreeNode and deprecating its support
- Refactor parent/child link management, copy and traversal methods in PhyloNode with typed interfaces
- Simplify tree I/O by removing XML write support and focusing on Newick and JSON formats
- Overhaul Dendrogram and TreeGeometry classes with consistent typing and improved coordinate calculations
- Consolidate tests to use PhyloNode exclusively and adapt to revised signatures and behaviors

Build:
- Update pyproject.toml dependency formatting, add plotly-stubs
- Refresh ruff.toml lint rules and per-file ignore directives

CI:
- Adjust ruff.toml for new lint settings

Tests:
- Revise tests throughout to replace TreeNode with PhyloNode and reflect updated APIs


src/cogent3/core/tree.py
```diff
@@ -51,33 +57,25 @@
...
+if TYPE_CHECKING:
+    import os
     import pathlib
+    from collections.abc import Callable, Generator, Iterable, Iterator, Sequence
 
-    from cogent3.evolve.fast_distance import DistanceMatrix
-
+    from typing_extensions import Self
 
-def distance_from_r(m1, m2):
```
rmcar17, line 61: Should I add this back and discontinue it? Doubt it's being used.

src/cogent3/core/tree.py
```diff
@@ -189,129 +204,159 @@ def source(self, value: str | None) -> None:
...
-                nodes_stack.append([new_child, old_child, len(old_child.children)])
+                nodes_stack.append((new_child, old_child, len(old_child.children)))
             else:  # no unvisited children
                 nodes_stack.pop()
         return root
 
     __deepcopy__ = deepcopy = copy
 
-    def copy_topology(self, constructor=None):
+    def copy_topology(self, constructor: type[T] | None = None) -> T | Self:
```
rmcar17, line 372: Should I deprecate all uses of constructor in the class (here and elsewhere) since from now on we'll only have the one?

src/cogent3/core/tree.py
```diff
@@ -320,28 +365,31 @@ def copy_topology(self, constructor=None):
...
+            constructor = cast("type[T]", self.__class__)
         children = [c.copy_topology(constructor) for c in self.children]
-        return constructor(name=self.name[:], children=children)
+        return constructor(name=self.name, children=children)
 
     # support for basic tree operations -- finding objects and moving in the
     # tree
     @property
-    def parent(self) -> typing_extensions.Self | None:
+    def parent(self) -> Self | None:
```
rmcar17, line 388: A way to get the parent or throw an error could be useful. A slight annoyance is that doing node.is_root() somewhere doesn't automatically narrow the type of node.parent to Self (instead of Self | None) which causes type checkers to complain.

src/cogent3/core/tree.py
```diff
@@ -189,129 +204,159 @@ def source(self, value: str | None) -> None:
...
+    def _copy_node(cls, node: Self, memo: dict[int, Any] | None = None) -> Self:
+        result = cls(node.name)
+        efc = node._exclude_from_copy
+        for k in node.__slots__:
+            if k not in efc:
+                setattr(result, k, deepcopy(getattr(node, k), memo=memo))
+        return result
+
     # support for copy module
-    def copy(self, memo=None, _nil=None, constructor: None = None):
```
rmcar17, line 291: `constructor` was unused before

src/cogent3/core/tree.py
```diff
@@ -526,84 +560,56 @@ def iter_nontips(
...
+    def tip_children(self) -> list[Self]:
         """Returns direct children of self that are tips."""
         return [i for i in self.children if not i.children]
 
-    def non_tip_children(self):
+    def non_tip_children(self) -> list[Self]:
         """Returns direct children in self that have descendants."""
         return [i for i in self.children if i.children]
 
-    def child_groups(self):
```
rmcar17, line 545: Just entirely removed this function, unclear what its purpose is.

src/cogent3/core/tree.py
```diff
@@ -526,84 +560,56 @@ def iter_nontips(
...
+        if len(tips) != len(tip_names_set):
+            missing = tip_names_set - set(self.get_tip_names())
             msg = f"tipnames {missing} not present in self"
             raise ValueError(msg)
 
         # scrub tree
-        if "black" in self.params:
-            for n in self.preorder(include_self=True):
-                n.params.pop("black", None)
+        for n in self.preorder(include_self=True):
```
rmcar17, line 652: This fixes a bug in the old version where calling this multiple times would give weird results.

src/cogent3/core/tree.py
```diff
@@ -639,16 +645,19 @@ def separation(self, other):
...
+                curr: Self | None = self
                 while curr is not None and curr is not other:
                     count += 1
                     curr = curr.parent
                 return count
             count += 1
             other = other.parent
-        return None
+        msg = "Nodes do not belong to the same tree."
+        raise TreeError(msg)
```
rmcar17, line 697: Backwards incompatible but I think a good change. Throws an error instead of returning None when trying to work out the number of edges separating two nodes which don't appear in the same tree.

src/cogent3/core/tree.py
```diff
@@ -717,10 +733,10 @@ def make_tree_array(self, dec_list=None):
 
         # get a list of tips() name if one is not supplied
         if not dec_list:
-            dec_list = [dec.name for dec in self.tips()]
+            dec_list = self.get_tip_names()
             dec_list.sort()
         # make a blank array of the right dimensions to alter
-        result = zeros((len(node_list), len(dec_list)))
+        result = np.zeros((len(node_list), len(dec_list)), dtype=np.uint8)
```
rmcar17: Given the array only stores 0/1 I just chose the minimum sized int.

src/cogent3/core/tree.py
```diff
@@ -729,16 +745,16 @@ def make_tree_array(self, dec_list=None):
                     result[i, j] = 1
         return result, node_list
 
-    def remove_deleted(self, is_deleted) -> None:
-        """Removes all nodes where is_deleted tests true.
+    def remove_deleted(self, should_delete: Callable[[Self], bool]) -> None:
```
rmcar17, line 789: Renamed this arg

src/cogent3/core/tree.py
```diff
@@ -782,10 +798,12 @@ def prune(self, keep_root: bool = False) -> None:
                 child.parent = curr_parent
                 if not has_length:
                     continue
+
+                # Not sound to be adding all numbers like this. Why?!?!?!
                 for key, value in child.params.items():
                     node_val = node.params.get(key)
                     if is_number(value) or is_number(node_val):
-                        value = (value or 0.0) + (node_val or 0.0)
+                        value = (value or 0.0) + (node_val or 0.0)  # noqa: PLW2901
```
rmcar17: Potentially replace with callback. This has the potential to mess with parameters we might place on trees generated by `piqtree` for instance.

src/cogent3/core/tree.py
```diff
@@ -993,23 +1009,21 @@ def get_edge_names(
 
         return edge_names
 
-    def _get_neighbours_except(
-        self, parent: typing_extensions.Self | None = None
-    ) -> list[typing_extensions.Self]:
+    def get_neighbours_except(self, parent: Self | None = None) -> list[Self]:
```
rmcar17, line 1066: Made this public (I've written this method multiple times while performing searches on trees).

src/cogent3/core/tree.py
```diff
@@ -1286,18 +1277,22 @@ def get_node_names(
             only tips returned
         """
         if tips_only:
-            nodes = self.tips(include_self=True)
+            nodes = self.tips(include_self=include_self)
```
rmcar17, line 1339: Now propagates include_self when tips_only is true.

src/cogent3/core/tree.py
```diff
@@ -1286,18 +1277,22 @@ def get_node_names(
...
         else:
             nodes = list(self.preorder(include_self=include_self))
         return [node.name for node in nodes]
 
-    def get_tip_names(self, include_self: bool = False) -> list[str]:
+    def get_tip_names(self, include_self: bool = True) -> list[str]:
```
rmcar17, line 1344: Now defaults to True (if you're calling get_tip_names on a tip, likely you also want to include that tip name)

src/cogent3/core/tree.py
```diff
@@ -1968,13 +1929,13 @@ def root_at_midpoint(self) -> typing_extensions.Self:
         _adjust_lengths_from_root(tip_name=a, mid_point=mid_point, tree=new_tree)
         return new_tree
 
-    def get_distances(self, names: list[str] | None = None) -> DistanceMatrix:
+    def get_distances(self, names: Sequence[str] | None = None) -> DistanceMatrix:
```
rmcar17: Do we need this function given it duplicates the other one?

src/cogent3/core/tree.py
```diff
@@ -400,30 +442,28 @@ def postorder(
...
-    def get_root(self) -> typing_extensions.Self:
+    def get_root(self) -> Self:
         """Returns root of the tree self is in."""
         curr = self
         while curr._parent is not None:
             curr = curr._parent
         return curr
 
-    def rooted(self, edge_name: str) -> typing_extensions.Self:
+    def rooted(self, edge_name: str) -> Self:
```
rmcar17, line 479: If a tree is missing some branch lengths this might have weird behaviour.

src/cogent3/core/tree.py
rmcar17: Also this only runs when branch lengths exist so unclear for the original intention is.

src/cogent3/core/tree.py
```diff
@@ -1508,14 +1512,19 @@ def tip_to_tip_distances(
             return subtree.tip_to_tip_distances(
                 default_length=default_length,
             )
-        default_length = 0 if hasattr(self, "length") else 1
+        default_length = (
```
rmcar17, line 1584: This doesn't respect the original default_length - unclear why something like this was here originally

src/cogent3/core/tree.py
```diff
@@ -507,13 +543,11 @@ def iter_tips(self, include_self=False):
             else:
                 yield curr
 
-    def tips(self, include_self=False):
+    def tips(self, include_self: bool = False) -> list[Self]:
```
rmcar17, line 559: Should this default to True?

src/cogent3/draw/dendrogram.py
```diff
@@ -818,22 +864,23 @@ def _build_fig(self, **kwargs) -> None:
...
         if scale_shape:
-            self.layout.shapes = [*self.layout.get("shape", []), scale_shape]
-            self.layout.annotations += (scale_text,)
+            self.layout.shapes = [*self.layout.get("shape", []), scale_shape]  # type: ignore[reportUnknownMemberType]
+            self.layout.annotations += (scale_text,)  # type: ignore[reportUnknownMemberType]
         else:
-            self.layout.pop("shapes", None)
+            self.layout.pop("shapes", None)  # type: ignore[reportUnknownMemberType]
 
+        # TODO: look into TypedDict as all the type: ignores are getting a bit silly
```
rmcar17, line 881: Lots of type: ignores all around here. UnionDict doesn't play nicely with typing. Perhaps look into TypedDict?

src/cogent3/core/tree.py
GavinHuttley: Yes, always deprecate public methods. And add the   `# pragma: no cover` tag.

src/cogent3/core/tree.py
GavinHuttley: Agreed. And with the `deprecate_args` decorator, in a case like this where it's not being used, you can just delete it from the signature.

src/cogent3/core/tree.py
GavinHuttley: deprecate as before

src/cogent3/core/tree.py
GavinHuttley: deprecate and mark with no cov...

src/cogent3/core/tree.py
GavinHuttley: Agreed.

src/cogent3/core/tree.py
GavinHuttley: No. Pick one, deprecate the other.

src/cogent3/core/tree.py
GavinHuttley: To be discussed

src/cogent3/core/tree.py
GavinHuttley: I think it's intended to return tip descendants from self. If self is a tip, I'd expect that result to be an empty list. I think we need to discuss.

src/cogent3/draw/dendrogram.py
GavinHuttley: The problem is UnionDict is really being used as a "propertied" dict (keys are asccessible as properties) so it mimics the behaviour of plotly's Layout class. The reason we did that was to avoid a massive overhead of importing Plotly until truly needing it for display.
I agree this decision needs to be revisited, but it's not a priority.

ruff.toml
```diff
@@ -48,17 +48,20 @@ select = ["ALL"]
...
+    "N801",
+    "ICN001",
     "PLR0913",
     "FBT001",
     "FBT002",
     "PT011",
-    ]
+    "S311", # Allow use of random
+    "PLR2004", # Magic number usage where appropriate
+    "PLC0415" # Allow delayed imports
```
GavinHuttley, line 63: Yay!

ruff.toml
```diff
@@ -74,7 +77,7 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
     "ANN",
     "N802",
     "N803",
-    "S608",  # sql injection unlikely
+    "S608" # sql injection unlikely
```
GavinHuttley, line 80: double yay!

src/cogent3/draw/drawable.py
```diff
@@ -1,11 +1,18 @@
 import os
 import pathlib
+from typing import TYPE_CHECKING, Any, TypeVar, cast
 
 import numpy
 
 from cogent3.util.misc import extend_docstring_from
 from cogent3.util.union_dict import UnionDict
 
+if TYPE_CHECKING:
```
GavinHuttley: add   # pragma: no cover

src/cogent3/draw/drawable.py
```diff
@@ -156,83 +166,83 @@ def __init__(
...
-        kwargs["width"] = kwargs.get("width", fig.layout.width)
-        kwargs["height"] = kwargs.get("height", fig.layout.height)
+        kwargs["width"] = kwargs.get("width", fig.layout.width)  # type: ignore[reportUnknownMemberType]
+        kwargs["height"] = kwargs.get("height", fig.layout.height)  # type: ignore[reportUnknownMemberType]
 
         path = pathlib.Path(path).expanduser().absolute()
         write_image(fig, str(path), **kwargs)
 
-    def to_image(self, format="png", **kwargs):
+    def to_image(self, format: str = "png", **kwargs: Any) -> bytes:
```
GavinHuttley: Lets deprecate this argument and replace it with `format_name` to be consistent with the other cases I've already addressed.

src/cogent3/evolve/fast_distance.py
```diff
@@ -814,7 +818,7 @@ def to_rich_dict(self) -> dict:
         }
 
     def take_dists(
-        self, names: list[str] | str, negate: bool = False
+        self, names: Sequence[str] | str, negate: bool = False
```
GavinHuttley, line 821: `PySeqStr` and `Sequence[str]` are the same. I defined the former to distinguish the Python `Sequence` type from the `cogent3.core.sequence.Sequence` type

src/cogent3/parse/cogent3_json.py
```diff
@@ -1,21 +1,26 @@
...
 import json
+from collections.abc import Iterable
+from typing import TYPE_CHECKING, Any
 
 from cogent3.app import data_store
 from cogent3.util.deserialise import deserialise_object
 from cogent3.util.io import open_
 from cogent3.util.misc import get_object_provenance
 
+if TYPE_CHECKING:
```
GavinHuttley: on all TYPE_CHECKING blocks do `# pragma: no cover`

src/cogent3/parse/tree.py
```diff
@@ -8,16 +8,17 @@
...
 left to future work.
 
-Should possibly build a dict of {label:TreeNode} while parsing to make it
+Should possibly build a dict of {label:PhyloNode} while parsing to make it
 convenient to fill in additional data later, e.g. to fill in sequences from
-their numeric labels in Newick format. Alternatively, maybe TreeNode should
+their numeric labels in Newick format. Alternatively, maybe PhyloNode should
 get a buildIndex() method that performs the equivalent task.
 
 As of 12/27/03, should be capable of parsing the ClustalW .dnd files without
```
GavinHuttley: look at that date! I suggest we delete most, if not all, of this docstring.

src/cogent3/parse/tree_xml.py
GavinHuttley, line 1: I think we should just deprecate the xml parser / writer for trees.

src/cogent3/phylo/consensus.py
```diff
@@ -123,30 +132,33 @@ def weighted_rooted_majority_rule(weighted_trees, strict=False, attr="support"):
...
+            nodes[tip_name] = tree_build([], cast("str", tip_name), params)
         else:
             queue.append((len(clade), clade))
 
     while queue:
         queue.sort()
-        (size, clade) = queue.pop(0)
-        new_queue = []
-        for _size2, ancestor in queue:
+        (_, clade) = queue.pop(0)
```
GavinHuttley: assignment parentheses not needed

tests/test_parse/test_tree.py
```diff
@@ -130,32 +130,32 @@ class DndParserTests(TestCase):
...
         """DndParser should produce the correct tree when there are no names"""
         obs = DndParser(no_names)
-        exp = PhyloNode()
-        exp.append(PhyloNode())
-        exp.append(PhyloNode())
-        exp.children[0].append(PhyloNode())
-        exp.children[0].append(PhyloNode())
-        exp.children[1].append(PhyloNode())
-        exp.children[1].append(PhyloNode())
+        exp = PhyloNode("")
```
GavinHuttley, line 133: shouldn't we just set the default name as a empty string in `PhyloNode.__init__`?

src/cogent3/core/tree.py
```diff
@@ -29,20 +29,26 @@
...
+    Literal,
+    SupportsIndex,
+    TypeVar,
+    cast,
+    overload,
+)
 
-import typing_extensions
-from numpy import ceil, log, zeros
+import numpy as np
```
GavinHuttley: nooooo!

src/cogent3/core/tree.py
```diff
@@ -51,33 +57,25 @@
 from cogent3.util.deserialise import register_deserialiser
 from cogent3.util.io import atomic_write, get_format_suffixes, open_
 from cogent3.util.misc import get_object_provenance, is_number
+from cogent3.util.warning import deprecated_callable
 
-if typing_extensions.TYPE_CHECKING:
+if TYPE_CHECKING:
```
GavinHuttley: # pragma: no cover

src/cogent3/core/tree.py
```diff
@@ -118,28 +116,51 @@ class TreeNode:
...
+
     _exclude_from_copy = frozenset(["_parent", "children"])
 
     def __init__(
         self,
-        name: str | None = None,
-        children: list[TreeNode] | None = None,
-        parent: TreeNode | None = None,
-        params: dict[str, object] | None = None,
+        name: str,
```
GavinHuttley, line 146: perhaps make `name` a keyword argument with default value `""`

src/cogent3/core/tree.py
```diff
@@ -118,28 +116,51 @@ class TreeNode:
...
+        children: Iterable[Self | str] | None = None,
```
GavinHuttley, line 147: do we really allow children input to be an iterable of strings?

src/cogent3/core/tree.py
```diff
@@ -118,28 +116,51 @@ class TreeNode:
         name_loaded: ?
     """
 
+    __slots__ = ("_parent", "children", "name", "name_loaded", "params")
```
GavinHuttley: yay!

src/cogent3/core/tree.py
GavinHuttley: or make it a bool array?

src/cogent3/core/tree.py
GavinHuttley: The original intention is to handle branch lengths and for that it does make sense use addition. Otherwise, I think treatment of any other key/values in params should be modifiable by a callback function. If a user doesn't define one, just retain the value on the preserved node -- that approach is what I would expect for a bootstrap support measure.

src/cogent3/core/tree.py
rmcar17: Decided to discontinue and instead rely on .copy()

src/cogent3/core/tree.py
rmcar17: I can see both schools of thought, I suppose there is now a divergence between get_tip_names and tips

src/cogent3/core/tree.py
rmcar17: :(

rmcar17: fwiw np isn't used in the codebase anywhere for the number of processes and treating numpy as np is pretty standard across the python ecosystem, but have made the change to numpy

tests/test_parse/test_tree.py
rmcar17: Could do, however I feel like nodes not having names should be the exception rather than the rule and default args might encourage users to not specify names. In the tree contains any duplicate names then it breaks functions like `to_rich_dict`, `get_nodes_dict`, perhaps `subsets`, `child_parent_map`, and worse if duplicate tip names are allowed. An alternative approach might be to have a default of None and move the `TreeBuilder._unique_name` behaviour to `PhyloNode` and activate this on None.

src/cogent3/core/tree.py
rmcar17: See https://github.com/cogent3/cogent3/pull/2412#discussion_r2246748079 for further discussion

src/cogent3/core/tree.py
rmcar17: In the tests at least, definitely - not immediately sure what the parser does. `_to_self_child` automatically converts it to the same class, so children is always of type `list[PhyloNode]`

rmcar17:
Can't reply to the bool array suggestion but done that now

rmcar17:
Just updated the PR addressing a large batch of the feedback, and some more wide-reaching changes making length/support attributes. Added a new key called `length_and_support` to `PhyloNode.to_rich_dict` in case users want to support other meanings of length/support in the params dict (which would otherwise cause a conflict here).

rmcar17:
@GavinHuttley now passes all tests

src/cogent3/core/tree.py
```diff
@@ -2263,13 +2369,29 @@ def load_tree(
...
+) -> PhyloNode:
     """returns a cogent3 PhyloNode instance"""
     # we load tree using make_tree, then populate edge attributes
-    edge_attr = data["edge_attributes"]
+    edge_attr = cast("dict[str, dict[str, Any]]", data["edge_attributes"])
+    length_and_support: dict[str, dict[str, float | None]] | None = cast(
+        "dict[str, dict[str, float | None]] | None", data.get("length_and_support")
+    )
+
+    if length_and_support is None:
```
GavinHuttley, line 2381: As the rich dict method is used for json serialisation we need to support creating the new type instance from the old type data, which means this deserialise function should handle the two versions. In the old version, there is only the `"edge_attributes"` key. We can check the "version" string in `data`, and/or just do the following
```python
if "length_and_support" not in data:
    length_and_support = {}
    for edge_name, edge_params in edge_attr.items():
          length_and_support[edge_name] = {"length": edge_params.pop("length", None), "support": edge_params.pop("support", None)}
```
Then a test case can be based on
```python
{'newick': '(Galago,(Gorilla,Rhesus)edge.0,HowlerMon)root',
 'edge_attributes': {'Galago': {'length': 0.1417910624295473},
  'Gorilla': {'length': 0.0173247791826725},
  'Rhesus': {'length': 0.020457707345485687},
  'edge.0': {'length': 0.0073129236698150635},
  'HowlerMon': {'length': 0.04134744219481945},
  'root': {'length': None}},
 'type': 'cogent3.core.tree.PhyloNode',
 'version': '2025.7.10a6'}
```

src/cogent3/core/tree.py
```diff
@@ -2263,13 +2369,29 @@ def load_tree(
...
+        warnings.warn(
```
GavinHuttley, line 2382: then delete this warning

GavinHuttley (changes requested):
Yet another monumental effort!

The only remaining issue is that the deserialise function should handle both the old and the new types. I've added some code and some data for testing that you can use to address this as a comment.

rmcar17:
@GavinHuttley actually the deserialiser already handles the new and old json formats. It does this by checking for the existence of the `"length_and_support"` key and if it's not there then it extracts the length/support from the params dict. I included the warning there to suggest to people to migrate to the new format.

GavinHuttley:
@rmcar17 and of course you're right! I should have been a bit more careful in my reading. It doesn't look like the two versions are tested though?

rmcar17:
Will add test

rmcar17:
@GavinHuttley added a test checking whether loading from the old format works as expected!

GavinHuttley (approved):
Awesome job @rmcar17 !

PR #2424, opened 2025-08-14
Deprecate test.py functions for scipy,  numpy , and statsmodel

khiron:
test.py 47 functions marked up as deprecated_callable

Internal functions that use deprecated functions 
- tree.py 
function `distance_from_r` used deprecated function `correlation` - replaced with scipy.stats.pearsonr
- contingency.py 
function `G-fit` uses `cogent3.maths.stats.test.G_fit `function that is not deprecated

all tests are green

tests/test_test.py remains and still needs to be culled of tests that test deprecated functions


src/cogent3/core/tree.py
```diff
@@ -46,11 +46,11 @@
     overload,
 )
 
+from scipy.stats import pearsonr
```
GavinHuttley: If you put this import inside the function, which I note is also being deprecated, then we can just delete the entire function instead of having to find this import and delete it too.

src/cogent3/maths/stats/test.py
```diff
@@ -1652,7 +1864,7 @@ def mw_test(x, y):
     return U, p
 
 
-def mw_boot(x, y, num_reps=1000):
+def mw_boot(x, y, num_reps=1000):  # convert to using scipy mannwhitneyu
```
GavinHuttley: delete comment

GavinHuttley (changes requested):
Awesome job. Thorough as usual!

Tiny tweaks

src/cogent3/core/tree.py
khiron: nice.  done.

src/cogent3/maths/stats/test.py
khiron: ugh.  deleted.  sry.

PR #2425, opened 2025-08-15
MAINT: alphabet, genetic_code, moltype, and location are now fully typed

rmcar17:
Two words: good luck!


pyproject.toml
```diff
@@ -136,6 +136,11 @@ skip_fragments = "README.*"
 new_fragment_template = "file: changelog.d/templates/new.md.j2"
 entry_title_template = "file: changelog.d/templates/title.md.j2"
 
+[[tool.mypy.overrides]]
```
GavinHuttley, line 139: double square brackets?

src/cogent3/core/alphabet.py
```diff
@@ -118,55 +108,64 @@ def num_canonical(self) -> int: ...
...
 
-class MonomerAlphabetABC(ABC):
+class MonomerAlphabetABC(ABC, Generic[StrOrBytes]):
     @abstractmethod
     def get_kmer_alphabet(
         self,
         k: int,
         include_gap: bool = True,
-    ) -> "KmerAlphabet": ...
+    ) -> "KmerAlphabet[StrOrBytes] | CharAlphabet[StrOrBytes]": ...
```
GavinHuttley, line 168: CharAlphabet[StrOrBytes] should not be a valid return value.

src/cogent3/core/alphabet.py
```diff
@@ -175,43 +174,44 @@ def as_bytes(self) -> bytes: ...
     def convert_seq_array_to(
         self,
         *,
-        alphabet: typing_extensions.Self,
+        alphabet: "CharAlphabet[Any]",
```
GavinHuttley, line 177: Given the definition of CharAlphabet all specify StrOrBytes elements, change "CharAlphabet[Any]" to "CharAlphabet[StrOrBytes]"?

I note that this happens in multiple locations, so do a search and replace.

src/cogent3/core/alphabet.py
```diff
@@ -536,43 +542,45 @@ def to_json(self) -> str:
         return json.dumps(self.to_rich_dict())
 
     @classmethod
-    def from_rich_dict(cls, data: dict) -> typing_extensions.Self:
+    def from_rich_dict(cls, data: dict[str, Any]) -> "CharAlphabet[Any]":
```
GavinHuttley: "CharAlphabet [Any]" -> "CharAlphabet [StrOrBytes]"

src/cogent3/core/alphabet.py
```diff
@@ -633,8 +644,8 @@ def _get_kmer_alpha(
 
 @functools.cache
 def make_converter(
-    src_alpha: CharAlphabet,
-    dest_alpha: CharAlphabet,
+    src_alpha: CharAlphabet[Any],
```
GavinHuttley, line 647: another one

src/cogent3/core/alphabet.py
```diff
@@ -676,19 +693,16 @@ def make_converter(
         src_indices.append(src_alpha.to_indices("T"))
         dest_indices.append(dest_alpha.to_indices("U"))
 
-    src_indices = numpy.concatenate(src_indices)
-    dest_indices = numpy.concatenate(dest_indices)
-
     return convert_alphabet(
-        src_indices.tobytes(),
-        dest_indices.tobytes(),
+        numpy.concatenate(src_indices).tobytes(),
```
GavinHuttley, line 697: why is numpy.concatenate required here?

src/cogent3/core/alphabet.py
```diff
@@ -1185,7 +1222,7 @@ def to_json(self) -> str:
         return json.dumps(self.to_rich_dict())
 
     @classmethod
-    def from_rich_dict(cls, data: dict) -> "KmerAlphabet":
+    def from_rich_dict(cls, data: dict[str, Any]) -> "KmerAlphabet[Any]":
```
GavinHuttley: "KmerAlphabet[Any]" -> "KmerAlphabet[StrOrBytes]"

Do a search and replace.

src/cogent3/core/alphabet.py
```diff
@@ -1348,7 +1387,7 @@ def from_index(self, index: int) -> str:
             msg = f"invalid {index=}"
             raise ValueError(msg) from e
 
-    def from_indices(self, indices: NumpyIntArrayType) -> list[str]:
+    def from_indices(self, indices: NumpyIntArrayType) -> list[str]:  # type: ignore
```
GavinHuttley: should this be `ignore[override]` too?

src/cogent3/core/location.py
```diff
@@ -83,21 +105,26 @@ class SpanI:
...
-    __slots__ = []  # override in subclass
+    __slots__ = ("end", "reverse", "start")  # override in subclass
+
+    def __init__(self) -> None:
+        self.start: int
+        self.end: int
+        self.reverse: bool
 
-    def __contains__(self, other) -> bool:
+    def __contains__(self, other: "SpanI | int") -> bool:
```
GavinHuttley: need to add   `# pragma: no cover` to all these `NotImplementedError` methods

src/cogent3/core/location.py
```diff
@@ -1293,7 +1319,7 @@ def get_gap_lengths(self) -> IntArrayTypes:
         lengths[1:] = numpy.diff(lengths)
         return lengths
 
-    def nongap(self) -> Iterator[SpanTypes]:
+    def nongap(self) -> Iterable[SpanTypes]:
```
GavinHuttley, line 1322: merits of Generator return type versus Iterable?

src/cogent3/core/location.py
```diff
@@ -1346,6 +1372,9 @@ def spans(self) -> Iterator[SpanTypes]:
         if self.num_gaps and self.gap_pos[-1] < self.parent_length:
             yield Span(self.gap_pos[-1], self.parent_length)
 
+    def iter_spans(self) -> Iterator[SpanTypes]:
```
GavinHuttley, line 1375: merits of Generator return type versus Iterable?

src/cogent3/core/moltype.py
```diff
@@ -527,13 +551,13 @@ def __repr__(self) -> str:
...
 
-    def __eq__(self, other: typing_extensions.Self) -> bool:
+    def __eq__(self, other: object) -> bool:
         return id(self) == id(other)
 
     def __len__(self) -> int:
         return len(self._monomers)
 
-    def __iter__(self) -> typing.Iterator[str]:
+    def __iter__(self) -> Iterator[StrOrBytes]:
```
GavinHuttley: Generator return type?

src/cogent3/core/moltype.py
```diff
@@ -1328,10 +1376,10 @@ def can_mispair(self, first: str, second: str) -> bool:
 
     def get_css_style(
         self,
-        colors: dict[str, str] | None = None,
+        colors: Mapping[str, str] | None = None,
```
GavinHuttley: why Mapping rather than dict?

src/cogent3/core/moltype.py
```diff
@@ -1401,14 +1463,14 @@ def _most_degen_alphabet(mt: MolType) -> c3_alphabet.CharAlphabet:
     "cogent3.core.moltype.MolType",
     "cogent3.core.c3_moltype.MolType",
 )
-def deserialise_c3_moltype(data: dict[str, str]) -> MolType:
+def deserialise_c3_moltype(data: dict[str, str]) -> MolType[Any]:
```
GavinHuttley, line 1466: "MolType[Any]" -> "MolType[StrOrBytes]"

Search and replace to catch all cases.

pyproject.toml
rmcar17: Correct. see: https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml-file

src/cogent3/core/alphabet.py
rmcar17: It is a valid return type, in CharAlphabet the method starts:

```python
    def get_kmer_alphabet(
        self, k: int, include_gap: bool = True
    ) -> "KmerAlphabet[StrOrBytes] | CharAlphabet[StrOrBytes]":
        """returns kmer alphabet with words of size k

        Parameters
        ----------
        k
            word size
        include_gap
            if True, and self.gap_char, we set
            KmerAlphabet.gap_char = self.gap_char * k

        Notes
        -----
        If self.missing_char is present, it is included in the new alphabet as
        missing_char * k
        """
        # refactor: revisit whether the include_gap argument makes sense
        if k == 1:
            return self
```

So the behaviour should change if we want it to not return a CharAlphabet

src/cogent3/core/alphabet.py
rmcar17: Alright so this is already handled by the way Generics work in Python. The definition for the class for MonomerAlphabet is 
```python
StrOrBytes = TypeVar("StrOrBytes", str, bytes)

class MonomerAlphabetABC(ABC, Generic[StrOrBytes]):
    ...
```

This essentially means that `MonomerAlphabetABC` can be subscripted by either doing `MonomerAlphabetABC[str]` or `MonomerAlphabetABC[bytes]`. Then any internal methods which use `StrOrBytes` essentially haved the generic replaced by `str` or `bytes` in that context. `MonomerAlphabetABC[Any]` essentially means any of the defined constraints, it doesn't matter. 

Let's say an instance of `MonomerAlphabetABC` is representing a `str` alphabet (i.e. it's of type MonomerAlphabetABC[str]). It doesn't make sense to use StrOrBytes in this specific context because that would mean that the other alphabet also has to be a string alphabet. However, the implementation is agnostic as to whether the other alphabet has the same base type (see implementation of `make_converter`). Hence using `Any` here, (saying both `MonomerAlphabetABC[str]` and `MonomerAlphabetABC[bytes]` are fine) should be correct.

src/cogent3/core/alphabet.py
rmcar17: That one makes sense for `from_rich_dict`. I was thinking that the type of alphabet was unknown from the dict, particularly as a class method, but on reflection this lets you do stuff like:

```python
my_alphabet = CharAlphabet[str].from_rich_dict(my_dict)
```

And that binds `my_alphabet` to a `str` `my_alphabet`

src/cogent3/core/alphabet.py
rmcar17: Using `[Any]` makes sense here as the method doesn't need to have both `src_alpha` and `dest_alpha` both be strings or both be bytes alphabets. Having one be a string alphabet and the other being a bytes alphabet works.

src/cogent3/core/alphabet.py
rmcar17: The previous lines used to have `src_indices = numpy.concatenate(src_indices)` which didn't play nicely with typing as it changed the type of `src_indices`, similar for dest. Just simplified it by moving the concatenate step into the return as well.

src/cogent3/core/alphabet.py
rmcar17: Makes sense for `from_rich_dict`

src/cogent3/core/alphabet.py
rmcar17: Yep!

src/cogent3/core/location.py
rmcar17: The `_spans_from_locations` it relies on is not actually a generator, but instead it returns a `SeqSpanTypes`. It returns a tuple of spans. Iterator was incorrect in this context, made the minimal change to `Iterable` instead.

src/cogent3/core/location.py
rmcar17: This one is an Iterator. In this context, essentially no difference between `Iterator` and `Generator`, though a `Generator` can accept additional typing arguments for send and return values instead of just yield values.

src/cogent3/core/moltype.py
rmcar17: Sure

src/cogent3/core/moltype.py
rmcar17: Have changed it back, played around with Mapping while fiddling around with covariance/contravariance issues (inserting bytes keys into what is previously defined as a str dict in the bytes case) but gave up in the end and forgot to change it back.

src/cogent3/core/moltype.py
rmcar17: For just a function (not method in generic class), can't have a TypeVar appear on just the return side. Needs to appear on the parameter side as well so the TypeChecker has something to bind it to. As MolType was defined

```python
StrOrBytes = TypeVar("StrOrBytes", str, bytes)

class MolType(Generic[StrOrBytes]):
    ...
```

Having a return type of `MolType[Any]` essentially means it either returns a `MolType[str]` or a `MolType[bytes]`.

rmcar17:
> Here's what I've learned. Rigorous application of type hinting is laborious and adds complexity and lines of code (e.g. specifying the overload variants of methods).

Yep, but great for end users though (a lot of places which had overloads were formerly singledispatch functions  which already adds to the LOC.

> I have asked some questions for cases where I don't understand the change. Happy to chat about those in person to save you the typing!

I should have read your initial comment first!

> In a few places you're doing explicit imports from modules, for example, typing _extensions import ..., whereas I've just imported the module. I chose the latter strategy because when you need access to another attribute of that module, you only have to change the code where you need the new thing.

I generally prefer doing the explicit imports. Having to type `typing.Any` or `typing.AnythingElse` adds a lot of unneeded noise to the function signatures. This noise is carried through to users when they hover over functions to see hints.

I also don't go back up to the import statements when I need a new thing from the module. Say if I need `Generator` - just typing that out where I need it, moving the text cursor back to it, then pressing `ctrl + .` makes a thing pop up to automatically add it to the imports.

> I've never been quite sure on the choice of `Iterator` versus `Generator` when type-hinting a function with the yield statement. Is there a change in IntelliSense behaviour?

Essentially no difference, though `Generator` is more general because it lets you add send/return types optionally.

src/cogent3/core/alphabet.py
GavinHuttley: OK, gotcha

src/cogent3/core/alphabet.py
GavinHuttley: Just to be clear, and showing my ignorance, if you make the return type "CharAlphabet [StrOrBytes]", mypy will complain about `my_alphabet = CharAlphabet[str].from_rich_dict(my_dict)`?

GavinHuttley: Gotcha

GavinHuttley (approved):
Awesome, I learned new things!

PR #2433, opened 2025-08-20
clean up deprecated maths/stats/test.py code and tests

khiron:
2 tests of functions (`multiple_n`,` _permute_observations`) that have been deprecated - removing the tests

`G_2_by_2` was calling deprecated function `tail` when the arg direction is set (1-tailed test) - now inlined 

fix dist_names in test:theoretical_quantiles


GavinHuttley (approved):
Bonza!

PR #2461, opened 2025-09-15
API: SeqsDataABC.get_hash() abstractmethod

GavinHuttley:
[NEW] sequence storage classes should handle this, return
     value should be a string or None. Not enforcing any
     specific hashing algorithm. For AlignedSeqsData
     classes, cogent3's default does the hashing off the aligned
     sequence.

[CHANGED] array_hash64() noew returns the hexdigest string
     rather than an integer.

[CHANGED] SequenceCollection.duplicated_seqs() method now
     uses this for identifying duplicated sequences.

## Summary by Sourcery

Introduce a consistent sequence hashing API by defining an abstract get_hash method, updating array_hash64 to return a hex digest string, and implementing hash caching in both unaligned and aligned sequence storage. Update duplicate detection to use the new get_hash interface.

Enhancements:
- Define abstract get_hash(seqid) on SeqsDataABC to standardize sequence hashing
- Change array_hash64 to return a hex digest string instead of an integer
- Implement get_hash in SeqsData and AlignedSeqsData with cached hashes stored in a new _hashes attribute
- Update SequenceCollection.duplicated_seqs to use storage.get_hash for identifying duplicate sequences

### DIFF
--- a/src/cogent3/core/alignment.py
+++ b/src/cogent3/core/alignment.py
@@ -102,7 +102,7 @@ MolTypes = c3_moltype.MolTypeLiteral | c3_moltype.MolType
 EPS = 1e-6
 
 
-def array_hash64(data: numpy.ndarray) -> int:
+def array_hash64(data: numpy.ndarray) -> str:
     """returns 64-bit hash of numpy array.
 
     Notes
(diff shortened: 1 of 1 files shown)


GavinHuttley:
CI failures because coveralls service still down

PR #2472, opened 2025-09-25
BUG: Fix unclosed Connection warnings

rmcar17:
The problem with closing the Connection of deletion of an annotation db is that the same underlying Connection object could exist in a different annotation db. For example, the `__setstate__` method constructs a new annotation db object (including a new _db Connection) before updating `self`'s `__dict__` with that of the other object. If a connection is closed upon the deletion of an annotation db, then `__setstate__` would destroy the temporarily created annotation db, which would then close the connection of self's db.

The best workaround I considered was to create a wrapper class for storing Connections, and closing the Connection only on the deletion of the wrapper object. In the above scenario for `__setstate__`, the same instance of the wrapper class is propagated from `new` to `self` instead of the underlying Connection. When `new` is destroyed on function exit, the `__del__` method on `DBWrapper` is not called because the same instance still exists under `self`.

The change is backwards incompatible. Previously, it was possible to construct an annotation db with `db` set to a Connection object. However, this feature was only used in a single test and it is possible to construct it with the DBWrapper instead (which is also safer if only the wrapper is allowed to close the connection). It also involves the renaming of a semi-private attribute.

## Summary by Sourcery

Introduce a DBWrapper class to encapsulate sqlite3 connections and avoid unclosed Connection warnings by closing only when the wrapper is garbage-collected, and refactor AnnotationDb to use this wrapper throughout its internal state management.

Bug Fixes:
- Prevent premature closing of shared sqlite Connections during annotation db deletion and state restoration.

Enhancements:
- Replace the internal _db attribute with a DBWrapper and update methods (__getstate__, __setstate__, __deepcopy__, _setup_db, db property, close, and constructors) to use the wrapper for connection lifecycle management.

Tests:
- Update tests to construct and compare AnnotationDb instances using the new DBWrapper instead of raw sqlite3.Connection.

### DIFF
--- a/.github/workflows/testing_develop.yml
+++ b/.github/workflows/testing_develop.yml
@@ -45,7 +45,7 @@ jobs:
 
       - name: "Run nox for ${{ matrix.python-version }}"
         env:
-          PYTHONTRACEMALLOC: "20"
+          PYTHONTRACEMALLOC: "3"
         shell: bash
         run: |
           dn=$(pwd)
--- a/.github/workflows/testing_develop.yml
+++ b/.github/workflows/testing_develop.yml
@@ -44,6 +44,8 @@ jobs:
           uv tool install -p venv nox
 
       - name: "Run nox for ${{ matrix.python-version }}"
+        env:
+          PYTHONTRACEMALLOC: "20"
         shell: bash
         run: |
           dn=$(pwd)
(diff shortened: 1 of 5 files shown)


rmcar17:
Nevermind, the CI suggests the issue is somewhere else

GavinHuttley:
Unfortunately, this doesn't really address the resource warnings. But thanks heaps for the effort at diving further into the morass of SQLite connection shenanigans.

Issue #2216, opened 2025-02-20
ENH: refactor `.get_motif_probs()` to using array operations


GavinHuttley:
This is related to #2215 and should probably be addressed by the same developer.

6Chenwen:
I have completed the reconstruction function, the details are as follows:
https://github.com/cogent3/cogent3/pull/2232

PR #2735, opened 2026-09-16
ENH: accelerate per-sequence motif counts

### LINKED ISSUE #2215: ENH: improve performance of `.counts_per_seq()` method
Rewrite so the operations are done on array transformation of sequences (e.g. `numpy.array(self.seqs[name])`).

Look at the tests for this method in test_new_alignment. Note this needs to be implemented for **both** sequence collections and alignments.

This requires a medium coding skill level but does not require advanced knowledge of cogent3.

Qinzi27:
Per-sequence counting repeatedly validates and filters individual observed motifs inside `Sequence.counts()`. This change shares a monomer eligibility lookup across the collection, uses `numpy.bincount` for single characters, and counts longer non-overlapping motifs as fixed-width records. It applies to both `SequenceCollection` and `Alignment`.

Fixes #2215.

### Implementation and compatibility

- Split sequences into non-overlapping blocks before filtering, preserving motif phase and the moltype-specific gap/missing/ambiguity rules.
- Use exact fixed-width record keys for multi-character motifs, including embedded zero and non-UTF-8 bytes. No additional `alphabet_size ** motif_length` counting bins are allocated. The existing `exclude_unobserved=False` output still includes the complete canonical alphabet.
- Read one transformed sequence at a time, preserving reverse complements, slices and names without introducing a full-alignment array cache.
- Preserve public signatures, result labels/order/dtypes, warnings and the existing container-specific empty-result conventions.
- Fix the empty-axis `ValueError` for a nonempty sequence shorter than `motif_length` in a ragged collection: that row now has zero counts.

### Validation

On Windows 11, Python 3.12.14, NumPy 2.4.6:

- 192 new regression cases pass. A Python `Counter` oracle independently checks six moltypes, flag combinations, motif lengths, transformations, bytes, empty/ragged inputs and long observed motifs. Running these cases against the original methods gives 191 passes and the expected short-sequence failure.
- The alignment, sequence, profile and new regression modules pass: **11,316 passed, 8 skipped**. Skips require the optional piqtree/HDF5 plugins. This is the relevant module suite, not the complete project/platform matrix.
- Both new helpers have 100% statement and branch coverage. Ruff check/format and `git diff --check` pass.
- An additional local differential audit found no unexpected changes in 1,216 API cases: 1,132 identical results and 84 short-sequence error-to-zero corrections. Another 96 array-key/stride/read-only checks passed.

### Measured performance

The included `benchmarks/counts_per_seq.py` compares against the method definitions at `37cda7a777689a13c4e17e71bbca679142d6f15f`, using unchanged dependencies and remaining code. It checks output equivalence before timing, uses seed 2215, alternates old/new order and saves all samples. Results below use 7 repetitions and the ratio of median latencies.

For an 8-sequence DNA alignment with 30,000 characters per sequence and default flags:

| Motif length | Baseline | Candidate | Speedup |
| --- | ---: | ---: | ---: |
| 1 | 3.95 ms | 0.51 ms | 7.67x |
| 2 | 49.84 ms | 13.85 ms | 3.60x |
| 3 | 50.04 ms | 11.29 ms | 4.43x |

Across 48 synthetic cases, the observed ranges were 3.22–9.76x for DNA, 1.16–4.65x for protein and 1.94–5.86x for bytes. These results are workload/hardware dependent; the protein three-character cases have smaller benefits. Peak memory was not measured.

```sh
python benchmarks/counts_per_seq.py --repeats 7 --output counts-per-seq-run-01.json
python -m pytest tests/test_core/test_alignment.py tests/test_core/test_sequence.py tests/test_core/test_profile.py tests/test_core/test_counts_per_seq.py -m "not slow and not internet"
```

This follows up on my earlier experiment in #2215. AI assisted implementation, test development and review; the reported checks and benchmarks were executed locally. The scope is `counts_per_seq()`; #2216 is not included.

## Summary by Sourcery

Accelerate per-sequence motif counting for collections and alignments while preserving compatibility and correcting short-ragged-sequence handling.

Bug Fixes:
- Return zero-filled per-sequence motif counts for nonempty ragged rows shorter than the requested motif length instead of raising an empty-axis error.

Enhancements:
- Accelerate per-sequence motif counting for sequence collections and alignments while preserving existing outputs, filtering rules, transformations, warnings, and empty-result behavior.
- Add exact observed multi-character motif handling without allocating Cartesian k-mer counting bins.

Documentation:
- Document how to run and interpret the per-sequence motif-count benchmark.

Tests:
- Add broad regression coverage for moltypes, filtering options, motif lengths, transformations, bytes, empty and ragged inputs, warnings, and long observed motifs.

Chores:
- Add a reproducible benchmark comparing current per-sequence counting performance with a trusted baseline.


GavinHuttley:
Hi @Qinzi27 ,

Thanks for the the effort, but I cannot accept this contribution. There are a number of reasons. First, and as a general principle, you should be discussing with the developers of a project the issue or issues that you want to work on. The coding agent that you've used does not understand what our design perspective is and so has led to changes that are inconsistent with the organization of the project. It has also produced code which would not satisfy the different argument options provided to the method.

In addition to these issues, it has added additional files into the repository. We don't accept new files unless there's an exceptionally good reason for it. 

If you really want an issue on the project to work on, leave a response to this comment and I'll find one which is more appropriate.

Qinzi27:
Hi Gavin,
Thank you for explaining the concerns and for offering to suggest another issue. I'm sorry I submitted the changes without first discussing the approach with you.
I originally explored this issue during your course and recently revisited it as a way to continue learning.(Zhenyu He, 2025) I should have developed a better understanding of the project's design and the method's full behaviour before submitting the AI-assisted changes. Passing tests was not enough to establish that the contribution was appropriate.(I tried modifying my parents' code in an attempt to make progress.)
I'd very much like to take up your offer of a more suitable issue. For the next task, I'll first study the relevant implementation and tests, then share a short proposed approach for discussion before making changes. I'll also follow the existing file organisation and check with you before proposing new files.
Thank you again for the opportunity to learn and contribute.

